PR #41, opened 2020-10-19
Add ImageNet format support

yasakova-anastasia:
### Summary

Added support for ImageNet format

### Checklist

### License

```python
#
```


datumaro/plugins/imagenet_format.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+class ImagenetImporter(Importer):
+
+    @classmethod
+    def detect(cls, path):
+        with logging_disabled(log.WARN):
+            return (len(cls.find_sources(path)) != 0)
+
+    def __call__(self, path, **extra_params):
```
zhiltsov-max, line 74: I think, you don't need your own implementation for `detect` an `call` after the simplification patch.

datumaro/plugins/imagenet_format.py
```diff
@@ -0,0 +1,140 @@
...
+                    'url': ann_file,
+                    'format': 'imagenet',
+                    'options': dict(extra_params),
+                })
+
+        return project
+
+    @staticmethod
+    def find_sources(path):
+        subsets = defaultdict(dict)
```
zhiltsov-max, line 94: Please check if the base class can help with this method too.

datumaro/plugins/imagenet_format.py
```diff
@@ -0,0 +1,140 @@
...
+
+class ImagenetExtractor(SourceExtractor):
+    def __init__(self, path):
+        assert osp.isfile(path), path
+        self._path = path
+        self._dataset_dir = osp.dirname(path)
+
+        super().__init__(subset=osp.splitext(osp.basename(path))[0])
+
+        labels = osp.join(osp.dirname(path), ImagenetPath.LABELS_FILE)
```
zhiltsov-max, line 31: I think, the next if will always follow the `str` case.

datumaro/plugins/imagenet_format.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def _load_categories(self, labels):
+        label_cat = LabelCategories()
+        for label in labels:
+            label_cat.add(label)
+
+        return { AnnotationType.label: label_cat }
+
+    def _load_items(self, path):
+        items = defaultdict(dict)
```
zhiltsov-max, line 55: I don't think you need `defaultdict` here, as well as in other places: https://docs.python.org/3.8/library/collections.html#collections.defaultdict.

datumaro/plugins/imagenet_format.py
```diff
@@ -0,0 +1,140 @@
...
+
+        with open(path, encoding='utf-8') as f:
+            for line in f:
+                image_name, label_id = line.split()
+                items[image_name[:-4]] = DatasetItem(
+                    id=image_name[:-4], subset=self._subset,
+                    image=image_name,
```
zhiltsov-max, line 62: The path should be an absolute path to the image, where it should be searched for.

datumaro/plugins/imagenet_format.py
```diff
@@ -0,0 +1,140 @@
...
+
+class ImagenetPath:
+    LABELS_FILE = 'synsets.txt'
+    IMAGES_DIR = 'data'
+
+
+class ImagenetExtractor(SourceExtractor):
+    def __init__(self, path):
+        assert osp.isfile(path), path
+        self._path = path
```
zhiltsov-max, line 26: The variables are unused.

datumaro/plugins/imagenet_format.py
```diff
@@ -0,0 +1,140 @@
...
+        extractor = self._extractor
+        images_dir = osp.join(subset_dir, ImagenetPath.IMAGES_DIR)
+        os.makedirs(images_dir, exist_ok=True)
+        self._images_dir = images_dir
+        image_labels = defaultdict(dict)
+        for subset_name, subset in self._extractor.subsets().items():
+            annotation_file = osp.join(subset_dir, '%s.txt' % subset_name)
+            annotation = ''
+            for item in subset:
+                label = item.annotations[0].label
```
zhiltsov-max, line 118: This line can easily throw an error, if an item does not have annotations. I also think we should cover the case with multiple labels per image - both for reading and saving.

datumaro/plugins/imagenet_format.py
```diff
@@ -0,0 +1,140 @@
...
+                if label not in image_labels:
+                    image_dir = osp.join(images_dir, item.id[:-(len(item.id.split('_')[-1]) + 1)])
+                    os.makedirs(image_dir, exist_ok=True)
+                    image_labels[label] = image_dir
+                if not item.has_image:
```
zhiltsov-max, line 123: Since we can just save annotations as `.txt` files, I think, we can remove this error and control this with `save_images` option.

tests/test_imagenet_format.py
```diff
@@ -0,0 +1,83 @@
...
+)
+from datumaro.plugins.imagenet_format import ImagenetConverter
+from datumaro.plugins.imagenet_format import ImagenetImporter
+from datumaro.util.test_utils import TestDir, compare_datasets
+
+
+class ImagenetFormatTest(TestCase):
+    def test_can_save_and_load(self):
+        source_dataset = Dataset.from_iterable([
+            DatasetItem(id='label_0_1', subset='train',
```
zhiltsov-max, line 18: Please also check if no `subset` and `id='a/b/c'` works well.

zhiltsov-max (changes requested):

datumaro/plugins/imagenet_format.py
```diff
@@ -0,0 +1,124 @@
...
+        images_dir = osp.join(subset_dir, ImagenetPath.IMAGES_DIR)
+        os.makedirs(images_dir, exist_ok=True)
+        self._images_dir = images_dir
+        image_labels = {}
+        for subset_name, subset in self._extractor.subsets().items():
+            annotation_file = osp.join(subset_dir, '%s.txt' % subset_name)
+            annotation = ''
+            for item in subset:
+                image_name = self._make_image_filename(item)
+                if (len(item.annotations) == 1
```
zhiltsov-max, line 90: Why is it a special case? 

How about doing something like:
```
for each image:
    labels[image] = set(lbl for lbl in annotations)
    if save_images:
        for each label:
            save(label_dir)
        if not labels:
             save(extra_dir)
writelines(['filename %s\n' % ' '.join(labels) for fname, labels in labels])
```

zhiltsov-max (changes requested):

datumaro/plugins/imagenet_format.py
yasakova-anastasia: I understand correctly that if an image has multiple labels, then we save it several times?

zhiltsov-max: Apparently, that is how the format works.

tests/test_imagenet_format.py
```diff
@@ -0,0 +1,126 @@
...
+        with TestDir() as test_dir:
+            ImagenetConverter.convert(source_dataset, test_dir, save_images=True)
+
+            parsed_dataset = ImagenetImporter()(test_dir).make_dataset()
+
+            compare_datasets(self, source_dataset, parsed_dataset)
+
+    def test_can_save_dataset_with_no_subsets(self):
+        source_dataset = Dataset.from_iterable([
+            DatasetItem(id='a/b/c',
```
zhiltsov-max, line 83: In practice, this will be the main case. We should find a way to find and load the image.

zhiltsov-max (changes requested):
I think, you should try to import the original ImageNet data. Maybe, we should split implementations for AC and regular ImageNet, or find a way to support both in one.

datumaro/plugins/imagenet_format.py
```diff
@@ -59,14 +83,24 @@ def apply(self):
...
-                            os.makedirs(image_dirs[label], exist_ok=True)
-                        self._save_image(item, osp.join(image_dirs[label], image_name))
+                image_name = item.id
+                labels[image_name] =  [p.label for p in item.annotations]
+                for label in labels[image_name]:
+                    label_name = extractor.categories()[AnnotationType.label][label].name
+                    if label not in image_dirs:
+                        image_dirs[label] = osp.join(subset_dir, label_name)
+                        os.makedirs(image_dirs[label], exist_ok=True)
+                    self._save_image(item, osp.join(image_dirs[label],
```
zhiltsov-max, line 96: There is a parameter for directory creation in `save_image()`, so missing dirs will be created automatically with base `_save_image()`. It will also cover multiple dirs from image ids.

datumaro/plugins/imagenet_format.py
```diff
@@ -27,19 +28,42 @@ def __init__(self, path):
...
 
     def _load_items(self, path):
         items = {}
+        labels = {}
         for images_dir in os.listdir(path):
-            for image_path in glob(osp.join(path, images_dir, '*.jpg')):
-                image_name = osp.splitext(osp.basename(image_path))[0]
-                label_id = self._categories[AnnotationType.label].find(images_dir)[0]
+            if images_dir == ImagenetPath.IMAGES_DIR_NO_LABEL:
+                for image_path in glob(osp.join(path, images_dir, '*.jpg')):
```
zhiltsov-max, line 41: I think, we should avoid hardcoding `.jpg`. Instead, we can try to load a file, check https://github.com/openvinotoolkit/datumaro/blob/develop/datumaro/plugins/image_dir.py#L31

zhiltsov-max:
Let's refactor a bit and merge.

zhiltsov-max (dismissed):

PR #55, opened 2020-11-20
Add CamVid format support

yasakova-anastasia:
### Summary

Added support for CamVid format

### Checklist

### License

```python
#
```


datumaro/plugins/camvid_format.py
```diff
@@ -0,0 +1,248 @@
...
+            label_map = parse_label_map(label_map_path)
+        return make_camvid_categories(label_map)
+
+    def _load_items(self, path):
+        items = {}
+        with open(path, encoding='utf-8') as f:
+            for line in f:
+                image, gt = line.split()
+                item_id = ('/'.join(gt.split('/')[2:]))[:-len(CamvidPath.IMAGE_EXT)]
+                image_path = osp.join(self._dataset_dir, image[1:])
```
zhiltsov-max, line 124: Let's drop a leading `/`only if it is there.

datumaro/plugins/camvid_format.py
```diff
@@ -0,0 +1,248 @@
...
+
+    @staticmethod
+    def _lazy_extract_mask(mask, c):
+        return lambda: mask == c
+
+
+class CamvidImporter(Importer):
+    @classmethod
+    def find_sources(cls, path):
+        subset_paths = [p for p in glob(osp.join(path, '*.txt'))
```
zhiltsov-max, line 147: Let's also cover the case with subsets placed somewhere deeper with `**` and `recursive=True`

datumaro/plugins/camvid_format.py
```diff
@@ -0,0 +1,248 @@
...
+                    self._categories[AnnotationType.mask].inverse_colormap
+                mask = lazy_mask(gt_path, inverse_cls_colormap)
+                mask = mask()
+                classes = np.unique(mask)
+                item_annotations = []
+                for label_id in classes:
+                    image = self._lazy_extract_mask(mask, label_id)
+                    item_annotations.append(Mask(image=image, label=label_id))
+                items[item_id] = DatasetItem(id=item_id, subset=self._subset,
+                    image=osp.join(image_path), annotations=item_annotations)
```
zhiltsov-max, line 136: There is an extra `osp.join`

datumaro/plugins/camvid_format.py
```diff
@@ -0,0 +1,248 @@
...
+        with open(path, encoding='utf-8') as f:
+            for line in f:
+                image, gt = line.split()
+                item_id = ('/'.join(gt.split('/')[2:]))[:-len(CamvidPath.IMAGE_EXT)]
+                image_path = osp.join(self._dataset_dir, image[1:])
+                gt_path = osp.join(self._dataset_dir, gt[1:])
+                inverse_cls_colormap = \
+                    self._categories[AnnotationType.mask].inverse_colormap
+                mask = lazy_mask(gt_path, inverse_cls_colormap)
+                mask = mask()
```
zhiltsov-max, line 129: Please, add a comment about using `lazy_mask` for loading through cache.

datumaro/plugins/camvid_format.py
```diff
@@ -0,0 +1,248 @@
...
+
+    def apply(self):
+        subset_dir = self._save_dir
+        os.makedirs(subset_dir, exist_ok=True)
+
+        for subset_name, subset in self._extractor.subsets().items():
+            segm_list = []
+            for item in subset:
+                masks = []
+                for a in item.annotations:
```
zhiltsov-max, line 179: Let's use `[x for x in ...]`

datumaro/plugins/camvid_format.py
```diff
@@ -0,0 +1,248 @@
...
+                    if a.type == AnnotationType.mask:
+                        masks.append(a)
+
+                if masks:
+                    compiled_mask = CompiledMask.from_instance_masks(masks)
+
+                    self.save_segm(
+                        osp.join(subset_dir, subset_name + CamvidPath.SEGM_DIR, item.id + CamvidPath.IMAGE_EXT),
```
zhiltsov-max, line 187: Please, split the line to fit in 80 characters.

datumaro/plugins/camvid_format.py
```diff
@@ -0,0 +1,248 @@
...
+    def save_label_map(self):
+        path = osp.join(self._save_dir, CamvidPath.LABELMAP_FILE)
+        write_label_map(path, self._label_map)
+
+    def _load_categories(self, label_map_source):
+        if label_map_source == LabelmapType.camvid.name:
+            # use the default Camvid colormap
+            label_map = CamvidLabelMap
+
+        elif label_map_source == LabelmapType.source.name and \
```
zhiltsov-max, line 224: We should also cover the case when `source`, but it does not have `MaskCategories`. It is the most used case - for example, it would be used after conversion of polygons to masks in a dataset. We also need to make sure `0,0,0` exists, and has index 0, otherwise masks will be wrong. Check VOC code for example. Maybe, it is worth trying to extract the common code fragment.

datumaro/plugins/camvid_format.py
```diff
@@ -0,0 +1,248 @@
...
+                subset_path, '.txt', 'camvid')
+        return sources
+
+
+LabelmapType = Enum('LabelmapType', ['camvid', 'source'])
+
+class CamvidConverter(Converter):
+    DEFAULT_IMAGE_EXT = '.png'
+
+    def __init__(self, extractor, save_dir,
```
zhiltsov-max, line 161: I feel like we need a bit of documentation an help here. It is easy to add with `build_cmdline_parser`.

tests/test_camvid_format.py
```diff
@@ -0,0 +1,196 @@
...
+
+class CamvidImportTest(TestCase):
+    def test_can_import(self):
+        class DstExtractor(TestExtractorBase):
+            def __iter__(self):
+                return iter([
+                    DatasetItem(id='0001TP_008550', subset='test',
+                        image=np.ones((2, 5, 3)),
+                        annotations=[
+                            Mask(image=np.array([[1, 1, 1, 0, 0], [0, 0, 0, 0, 0]]), label=0),
```
zhiltsov-max, line 45: Let's split the array definition into few lines to simplify reading. You can also simplify the test by leaving only one line in the image.

tests/test_camvid_format.py
```diff
@@ -0,0 +1,196 @@
...
+class TestExtractorBase(Extractor):
+    def _label(self, camvid_label):
+        return self.categories()[AnnotationType.label].find(camvid_label)[0]
+
+    def categories(self):
+        return Camvid.make_camvid_categories()
+
+class CamvidImportTest(TestCase):
+    def test_can_import(self):
+        class DstExtractor(TestExtractorBase):
```
zhiltsov-max, line 39: Let's use `Dataset.from_iterable` for definition of the dataset.

tests/test_camvid_format.py
```diff
@@ -0,0 +1,196 @@
...
+                        Mask(image=np.array([[0, 0, 0, 1, 0]]), label=0),
+                        Mask(image=np.array([[0, 1, 1, 0, 0]]), label=3),
+                        Mask(image=np.array([[1, 0, 0, 0, 1]]), label=4),
+                    ]),
+                ])
+
+        with TestDir() as test_dir:
+            self._test_save_and_load(TestExtractor(),
+                partial(CamvidConverter.convert, label_map='camvid'),
+                test_dir, target_dataset=DstExtractor())
```
zhiltsov-max, line 114: I can't see differences in the extractors above, I feel like a single definition is enough.

tests/test_camvid_format.py
```diff
@@ -0,0 +1,196 @@
...
+
+        with TestDir() as test_dir:
+            self._test_save_and_load(TestExtractor(),
+                partial(CamvidConverter.convert, label_map='camvid'), test_dir)
+
+    def test_dataset_with_source_labelmap(self):
+        class SrcExtractor(TestExtractorBase):
+            def __iter__(self):
+                yield DatasetItem(id=1, image=np.ones((8, 8, 3)),
+                annotations=[
```
zhiltsov-max, line 166: Please, fix indentation.

zhiltsov-max (changes requested):
Looks good, but needs some additions.

datumaro/plugins/camvid_format.py
yasakova-anastasia: How are we supposed to generate colormap if we don't have MaskCategories? Can we do the same as in VOC?

zhiltsov-max: Yes, let's use the same colormap generation algorithm.

zhiltsov-max (approved):
The only problem I see is that the new format does not support changing colormap, like VOC does. But this can be added later, when needed.

PR #68, opened 2020-12-23
Kate/splitter

jihyeonyi:
### Summary
Added task-specific annotation splitters.
- SplitforClassification
- SplitforMatchingReID
- SplitforDetection

### How to test
```bash
python -m unittest -v tests/test_splitter.py
```

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/plugins/splitter.py
+++ b/datumaro/plugins/splitter.py
@@ -495,7 +495,7 @@ class DetectionSplit(_TaskSpecificSplit):
             )
 
         by_splits = dict()
-        for sname in subsets:
+        for sname in self._subsets:
             by_splits[sname] = []
 
         total = len(self._extractor)
--- a/datumaro/plugins/splitter.py
+++ b/datumaro/plugins/splitter.py
@@ -10,20 +10,20 @@ from datumaro.components.extractor import (Transform, AnnotationType, \
     DEFAULT_SUBSET_NAME)
 from datumaro.components.cli_plugin import CliPlugin
 
-class TaskSpecificSplitter(Transform):            
+class TaskSpecificSplitter(Transform):
     _parts = []
 
     def __init__(self, dataset:Dataset, seed):
         super().__init__(dataset)
-        
-        np.random.seed(seed)  
-    
+
+        np.random.seed(seed)
+
     def _get_required(self, ratio):
         min_value = np.max(ratio)
         for i in ratio:
             if i < min_value and i > 1e-7:
-                min_value = i        
-        required = int (np.around(1.0) / min_value) 
+                min_value = i
+        required = int (np.around(1.0) / min_value)
         return required
 
     def _normalize_ratio(ratio):
@@ -34,23 +34,23 @@ class TaskSpecificSplitter(Transform):
 
         ratio = np.array(ratio)
         ratio /= np.sum(ratio)
-        return ratio       
+        return ratio
 
     def _get_sections(self, dataset_size, ratio):
         n_splits = [int(np.around(dataset_size * r)) for r in ratio[:-1]]
         n_splits.append(dataset_size - np.sum(n_splits))
 
-        # if there are splits with zero samples even if ratio is not 0, 
+        # if there are splits with zero samples even if ratio is not 0,
         # borrow one from the split who has one or more.
         for ii in range(len(n_splits)):
             if n_splits[ii] == 0 and ratio[ii] > 1e-7:
                 midx = np.argmax(n_splits)
                 if n_splits[midx] > 0:
                     n_splits[ii] += 1
-                    n_splits[midx] -= 1   
-        sections = np.add.accumulate(n_splits[:-1])        
+                    n_splits[midx] -= 1
+        sections = np.add.accumulate(n_splits[:-1])
         return sections
-    
+
     def _group_by_attributes(self, items):
         '''
         Args:
@@ -68,16 +68,16 @@ class TaskSpecificSplitter(Transform):
         return by_attributes
 
     def _split_indice(self, indice, group_name, ratio, required):
-        filtered_size = len(indice)  
+        filtered_size = len(indice)
         if required > filtered_size:
             log.warning("There's not enough samples for filtered group, \
                 '{}'}'".format(group_name))
-        sections = self._get_sections(filtered_size, ratio) 
-        splits = np.array_split(indice, sections)       
-        assert len(ratio)==len(splits) 
-        return splits    
+        sections = self._get_sections(filtered_size, ratio)
+        splits = np.array_split(indice, sections)
+        assert len(ratio)==len(splits)
+        return splits
 
-    def _find_split(self, index):        
+    def _find_split(self, index):
         for subset_indices, subset in self._parts:
             if index in subset_indices:
                 return subset
@@ -90,29 +90,29 @@ class TaskSpecificSplitter(Transform):
 class SplitforClassification(TaskSpecificSplitter, CliPlugin):
     """
     Splits dataset into train/val/test set in class-wise manner. |n
-    |n   
+    |n
     Notes:|n
-    - Single label is expected for each DatasetItem.|n 
-    - If there are not enough images in some class or attributes group, 
-      the split ratio can't be guaranteed.|n 
-    - The split ratio will be normalized so that the sum is 1.|n  
+    - Single label is expected for each DatasetItem.|n
+    - If there are not enough images in some class or attributes group,
+      the split ratio can't be guaranteed.|n
     |n
     Example:|n
     |s|s%(prog)s --train .5 --val .2 --test .3
-    """        
+    """
     @classmethod
     def build_cmdline_parser(cls, **kwargs):
-        parser = super().build_cmdline_parser(**kwargs)            
-        parser.add_argument('-t', '--train', type=float, 
-            help="Ratio for train set")        
+        parser = super().build_cmdline_parser(**kwargs)
+        parser.add_argument('-t', '--train', type=float,
+            help="Ratio for train set")
         parser.add_argument('-v', '--val', type=float,
-            help="Ratio for validation set")        
+            help="Ratio for validation set")
         parser.add_argument('-e', '--test',type=float,
             help="Ratio for test set")
         parser.add_argument('--seed', type=int, help="Random seed")
         return parser
 
-    def __init__(self, dataset:Dataset, train=0.0, val=0.0, test=0.0, seed=None):        
+    def __init__(self, dataset:Dataset,
+        train=0.0, val=0.0, test=0.0, seed=None):
         super().__init__(dataset, seed)
 
         subsets = ['train', 'val', 'test']
@@ -123,19 +123,17 @@ class SplitforClassification(TaskSpecificSplitter, CliPlugin):
             "Ratios are expected to be in the range [0, 1], but got %s" % splits
 
         total_ratio = np.sum(sratio)
-        assert total_ratio > 1e-7, \
-            "Sum of ratios is expected to be positive, got %s, which is %s" % \
-            (splits, total_ratio)
+        if not abs(total_ratio - 1.0) <= 1e-7:
+            raise Exception(
+                "Sum of ratios is expected to be 1, got %s, which is %s" %
+                (splits, total_ratio))
 
-        sratio /= total_ratio  # normalize
-        required = self._get_required(sratio)
-
-        ## support only single label for a DatasetItem    
+        ## support only single label for a DatasetItem
         ## 1. group by label
         by_labels = dict()
         for idx, item in enumerate(self._extractor):
             labels = []
-            for ann in item.annotations:                
+            for ann in item.annotations:
                 if ann.type == AnnotationType.label:
                     labels.append(ann)
             assert len(labels) == 1, \
@@ -148,13 +146,14 @@ class SplitforClassification(TaskSpecificSplitter, CliPlugin):
             if label not in by_labels:
                 by_labels[label] = []
             by_labels[label].append((idx, ann))
-        
+
         self._subsets = set(subsets) # output subset names
         by_splits = dict()
         for subset in subsets:
             by_splits[subset] = []
-     
-        ## 2. group by attributes    
+
+        ## 2. group by attributes
+        required = self._get_required(sratio)
         for label, items in by_labels.items():
             np.random.shuffle(items)
             by_attributes = self._group_by_attributes(items)
@@ -164,7 +163,7 @@ class SplitforClassification(TaskSpecificSplitter, CliPlugin):
                 for subset, split in zip(subsets, splits):
                     if len(split) > 0:
                         by_splits[subset].extend(split)
-        
+
         parts = []
         for subset in self._subsets:
             parts.append((set(by_splits[subset]), subset))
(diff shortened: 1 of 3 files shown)


datumaro/plugins/splitter.py
```diff
@@ -0,0 +1,480 @@
...
+            attributes = tuple(sorted(ann.attributes.items()))
+            if attributes not in by_attributes:
+                by_attributes[attributes] = []
+            by_attributes[attributes].append(idx)
+        return by_attributes
+
+    def _split_indice(self, indice, group_name, ratio, required):
+        filtered_size = len(indice)
+        if required > filtered_size:
+            log.warning("Not enough samples for filtered group, '{}'".format(
```
zhiltsov-max, line 85: Please follow the `%` formatting style, used everywhere else in the project. You can also omit formatting at all, because `logging` can substitute values itself.

datumaro/plugins/splitter.py
```diff
@@ -0,0 +1,480 @@
...
+
+from datumaro.components.dataset import Dataset
+from datumaro.components.extractor import (Transform, AnnotationType, \
+    DEFAULT_SUBSET_NAME)
+
+class TaskSpecificSplitter(Transform):
+    def __init__(self, dataset, seed):
+        super().__init__(dataset)
+
+        np.random.seed(seed)
```
zhiltsov-max, line 16: I think, this should be done closer to the actual use, because a lot can happen after constructor is called.

datumaro/plugins/splitter.py
```diff
@@ -0,0 +1,480 @@
...
+            The sum of ratios is expected to be 1.
+        attr_pid: str
+            attribute name representing the person id. (default: PID)
+        seed : int, optional
+        """
+        super().__init__(dataset, seed)
+
+        id_subsets, id_ratio = self._validate_splits(id_splits)
+
+        groups = set()
```
zhiltsov-max, line 203: Let's defer these calculations to the first use of `__iter__`.

datumaro/plugins/splitter.py
```diff
@@ -0,0 +1,480 @@
...
+        trainval_ratio = []
+        for subset in trainval_subsets:
+            if subset in id_subsets:
+                val = id_ratio[id_subsets.index(subset)]
+            else:
+                val = 0.0
+            trainval_ratio.append(val)
+        trainval_ratio = np.array(trainval_ratio)
+        total_ratio = np.sum(trainval_ratio)
+        if total_ratio < 1e-7:
```
zhiltsov-max, line 334: This number is used multiple times, maybe it's worth to be extracted as a constant.

datumaro/plugins/splitter.py
```diff
@@ -0,0 +1,480 @@
...
+    Then, tags 'test' into 'gallery'/'query' in class-wise random manner.|n
+    Then, splits 'train+val' into 'train'/'val' sets in the same way.|n
+    Therefore, the final subsets would be 'train', 'val', 'test'. |n
+    And 'gallery', 'query' are tagged using anntoation group.|n
+    You can get the 'gallery' and 'query' subsets using 'get_subset_by_group'.|n
+    Notes:|n
+    - Single label is expected for each DatasetItem.|n
+    - Each label is expected to have attribute representing the person id. |n
+    """
+    _group_map = dict()
```
zhiltsov-max, line 179: This will be a global variable at class level.

datumaro/plugins/splitter.py
```diff
@@ -0,0 +1,480 @@
...
+            )
+        group_id = self._group_map[group]
+        subset = self.select(lambda item: item.annotations[0].group == group_id)
+        return subset
+
+class SplitforDetection(TaskSpecificSplitter):
+    """
+    Splits dataset into train/val/test set for detection task.|n
+    For detection dataset, each image can have multiple bbox annotations.|n
+    Since one DataItem can't be included in multiple subsets at the same time,
```
zhiltsov-max, line 370: Technically, you can copy an item with different subset values, if it is needed. An item is addressed by `(id, subset)`

datumaro/plugins/splitter.py
```diff
@@ -0,0 +1,480 @@
...
+            The sum of ratios is expected to be 1.
+        seed : int, optional
+        """
+        super().__init__(dataset, seed)
+
+        subsets, sratio = self._validate_splits(splits)
+
+        ## 1. group by bbox label
+        by_labels = dict()
+        for idx, item in enumerate(self._extractor):
```
zhiltsov-max, line 397: Again, lets avoid iterating over input dataset, unless required.

datumaro/plugins/splitter.py
```diff
@@ -0,0 +1,480 @@
...
+            for ann in item.annotations:
+                if ann.type == AnnotationType.bbox:
+                    if not hasattr(ann, 'label') or ann.label is None:
+                        label = None
+                    else:
+                        label = str(ann.label)
```
zhiltsov-max, line 403: `annotation.label` is an index in `extractor.categories()[AnnotationType.label]` (of type [LabelCategories](https://github.com/openvinotoolkit/datumaro/blob/develop/datumaro/components/extractor.py#L53)). To get the name of a label, do `extractor.categories()[AnnotationType.label][label].name`.

tests/test_splitter.py
```diff
@@ -0,0 +1,388 @@
...
+
+class SplitterTest(TestCase):
+    def _generate_dataset(self, config):
+        # counts = {(0,0):20, (0,1):20, (0,2):30, (1,0):20, (1,1):10, (1,2):20}
+        # attr1 = ['attr1', 'attr2']
+        # attr2 = ['attr1', 'attr3']
+        # config = { "label1": { "attrs": attr1, "counts": counts },
+        #            "label2": { "attrs": attr2, "counts": counts }}
+        def _get_subset():
+            return np.random.choice(['', 'a', 'b'], p = [0.5, 0.3, 0.2])
```
zhiltsov-max, line 20: It is better to avoid randomness in tests.

tests/test_splitter.py
```diff
@@ -0,0 +1,388 @@
...
+                    attributes = dict()
+                    if isinstance(attrs, tuple):
+                        for aname, value in zip(anames, attrs):
+                            attributes[aname] = value
+                    else:
+                        attributes[anames[0]] = attrs
+                    for _ in range(count):
+                        idx += 1
+                        iterable.append(DatasetItem(
+                            id=str(idx),
```
zhiltsov-max, line 40: `str` is applied automatically.

tests/test_splitter.py
```diff
@@ -0,0 +1,388 @@
...
+                annotations=[Label(0), Label(2)],
+                subset=""
+            )
+        ]
+        categories = {
+            AnnotationType.label: LabelCategories.from_iterable(
+                'label_' + str(label) for label in range(3)
+            )}
+        source = Dataset.from_iterable(iterable, categories)
+        with self.assertRaises(Exception):
```
zhiltsov-max, line 199: Consider using at least `assertRaisesRegex` to avoid successful test passing in unexpected failures, like `KeyError` etc. .

tests/test_splitter.py
```diff
@@ -0,0 +1,388 @@
...
+
+        dataset = Dataset.from_iterable(iterable, categories)
+        return dataset, totals
+
+    def test_split_for_detection(self):
+        def append_bbox_coco(annotations, **kwargs):
+            annotations.append(Bbox(1,1,2,2, label=kwargs['label_id'],
+                                id=kwargs['ann_id'],
+                                attributes=kwargs['attributes'],
+                                group=kwargs['label_id']))
```
zhiltsov-max, line 314: In `COCO` (and in general) `group` is expected to match `id`. A group is supposed to represent a single object, having multiple annotations (polygons, boxes, (key-)points, masks etc.).

tests/test_splitter.py
```diff
@@ -0,0 +1,388 @@
...
+
+                splits = [('train', train/total),
+                          ('val',   val/total),
+                          ('test',  test/total)]
+                actual = splitter.SplitforDetection(source, splits)
+                subsets = dict()
+                for sname in ['train', 'val', 'test']:
+                    subset = actual.get_subset(sname)
+                    subsets[sname] = subset
+                self.assertEqual(train, len(subsets['train']))
```
zhiltsov-max, line 386: Consider using `len(get_subset())` directly, as nothing else is needed for checking.

tests/test_splitter.py
```diff
@@ -0,0 +1,388 @@
...
+                id='2',
+                annotations=[Label(0), Label(2)],
+                subset=""
+            )
+        ]
+        categories = {
+            AnnotationType.label: LabelCategories.from_iterable(
+                'label_' + str(label) for label in range(3)
+            )}
+        source = Dataset.from_iterable(iterable, categories)
```
zhiltsov-max, line 198: You can do the same simpler with:

```python
Dataset.from_iterable([
    DatasetItem(1, annotations=[Label(0), Label(1)]),
    DatasetItem(2, annotations=[Label(0), Label(2)])
], categories=['a', 'b', 'c'])
```

zhiltsov-max (changes requested):
Thank you for a great PR!

I'd suggest fixing code formatting (spaces in arguments) and string formatting to keep the project code in a single style.

tests/test_splitter.py
```diff
@@ -0,0 +1,388 @@
...
+        def append_bbox_mot(annotations, **kwargs):
+            annotations.append(Bbox(1,1,2,2, label=kwargs['label_id'],
+                                attributes=kwargs['attributes']))
+            annotations.append(Label(kwargs['label_id'],
+                                attributes=kwargs['attributes']))
+        def append_bbox_widerface(annotations, **kwargs):
+            annotations.append(Bbox(1,1,2,2, attributes=kwargs['attributes']))
+            annotations.append(Label(0, attributes=kwargs['attributes']))
+
+        functions = {
```
zhiltsov-max, line 355: I think, this kind of testing is hard to maintain, but let's leave it as is. However, a generic test should also be added.

datumaro/plugins/splitter.py
```diff
@@ -0,0 +1,480 @@
...
+
+import logging as log
+import numpy as np
+
+from datumaro.components.dataset import Dataset
+from datumaro.components.extractor import (Transform, AnnotationType, \
+    DEFAULT_SUBSET_NAME)
+
+class TaskSpecificSplitter(Transform):
+    def __init__(self, dataset, seed):
```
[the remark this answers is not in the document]
jihyeonyi: I see. 
I'll add the command line feature later for the other classes (SplitterforXXX) because I have not tested it yet.
Can I add the feature later, not in this pull request?

datumaro/plugins/splitter.py
jihyeonyi: Thanks for the comment. I'll replace them all.

datumaro/plugins/splitter.py
jihyeonyi: I'll change this too.

jihyeonyi: Random shuffling occurs several times. 
Do you think I really need to set seed whenever I use the shuffling function?

datumaro/plugins/splitter.py
jihyeonyi: Thank you. I'll make it to a constant.

datumaro/plugins/splitter.py
jihyeonyi: Yes. it is used by 'get_subset_by_group' function, so I make it a global variable.

datumaro/plugins/splitter.py
jihyeonyi: Oh, for the detection task, it should not occur. Because testing should be done with UNSEEN images.

datumaro/plugins/splitter.py
jihyeonyi: I don't need the name itself. ID is enough.
I just want to treat it as a string type.

tests/test_splitter.py
jihyeonyi: I'll consider a different way to do this.

tests/test_splitter.py
jihyeonyi: ok. I'll remove it.

tests/test_splitter.py
jihyeonyi: Thank you for letting me know this. I'll change all the occurence.

tests/test_splitter.py
jihyeonyi: Oh.. I missed it. You're right. group should be same to ann_id.  Thanks.

tests/test_splitter.py
jihyeonyi: Actually.. I need to check other values, but I can't, with the current statistics function.
Anyway, I'll change this, too. Thank you!

tests/test_splitter.py
jihyeonyi: Thank you

tests/test_splitter.py
jihyeonyi: Can you share your thought more on the generic test?

datumaro/plugins/splitter.py
jihyeonyi: It is required. Can you suggest a better way to iterating over the input dataset?

datumaro/plugins/splitter.py
jihyeonyi: I don't understand what you mean.. Can you explain more about this?

jihyeonyi:
I need help regarding complexity. 
I refactor the code and get complexities like below.
```bash
$ python -m mccabe datumaro/plugins/splitter.py --min 10
188:4: 'SplitforMatchingReID.__init__' 16
307:4: 'SplitforMatchingReID._rebalancing' 12
370:4: 'SplitforDetection.__init__' 19
```
But Codacy gives 35 for SplitforMatchingReID.__init__ and 20 for SplitforDetection.__init__
Why they're different? And how can resolve this issue?

zhiltsov-max:
Complexity from these tools is not something we look at. Don't bother with this.

tests/test_splitter.py
zhiltsov-max: Basically, the same, as these, but with the most complete inputs. For example - full instances from COCO (bbox or polygons, or mask + points, if it matters, in a group) with a couple of attributes and dataset items - depending on the algorithm. You can also look at test coverage, if you need some hints.

datumaro/plugins/splitter.py
zhiltsov-max: Currently, there is a lot of computations going on in the construction time, including input dataset traversal etc. I suggest moving these computations to the first use - in `__iter__`. This can be done by saving parameters in constructor and introducing a flag variable like `initialized` or something like this. Then, when splits are actually required, compute them.

Input dataset traversal should be considered heavy operation, so let's try to avoid it, when not necessary.

datumaro/plugins/splitter.py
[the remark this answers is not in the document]
zhiltsov-max: Ok, no problems with this. Actually, they are added automatically, it's just some polishing and customization.

tests/test_splitter.py
jihyeonyi: I see. Do you think I should enter more complicated test sets even though the DetectionSplit (I changed the class name) only cares Bbox annotations?

datumaro/plugins/splitter.py
```diff
@@ -0,0 +1,478 @@
...
+            self._split_by_attr(testset, test_subsets, test_ratio, by_groups,
+                                dataset_key=attr_pid)
+
+            # tag using group
+            for idx, item in enumerate(self._extractor):
+                for subset, split in by_groups.items():
+                    if idx in split:
+                        group_id = self._group_map[subset]
+                        item.annotations[0].group = group_id
+                        break
```
jihyeonyi, line 287: Here, I use 'group' property to split 'test' subset into 'query' and 'gallery' set. 
It seems that the default subset doesn't support a hierarchy. 
In the original implementation, we add "Auxiliary" attribute to the annotation. 
Do you think I change the current implementation w.r.t the original one? 
Or do I have to implement the subset hierarchy? 
Please share your opinion.

datumaro/plugins/splitter.py
jihyeonyi: I understand now. Thank you, I'll apply your suggestion.

jihyeonyi: If we apply this lazy initialization, some error checking would be delayed, too. 
It's because some errors can be noticed while iterating over the dataset.
So even if XXXSplit is constructed successfully, it can give an error later. 
I don't know this is good for users. How do you think?

datumaro/plugins/splitter.py
zhiltsov-max: Just make sure the results are reproducible. I suggest moving seed setting as much close to the point of use, as possible.

datumaro/plugins/splitter.py
zhiltsov-max: I think we only need to validate input parameters in constructor. If their application to this specific input dataset produces an error - it looks ok for me to know about this when I run the operation.

datumaro/plugins/splitter.py
zhiltsov-max: Yes, there in no exact support for split hierarchy yet. To do what you want, I'd create 2 extra subsets - `test_gallery` and `test_query`, this would look quite natural (at least, in Datumaro). Technically, subsets are extractors themselves, so it is possible to extend the implementation with some nesting. But it would require lots of changes in the importing/exporting code.

tests/test_splitter.py
zhiltsov-max: No, there is no need in that.

tests/test_splitter.py
```diff
@@ -0,0 +1,631 @@
...
+            iterable.append(item)
+
+        dataset = Dataset.from_iterable(iterable, categories)
+        return dataset, totals
+
+    @staticmethod
+    def _get_append_bbox(dataset_type):
+        def append_bbox_coco(annotations, **kwargs):
+            annotations.append(
+                Bbox(
```
zhiltsov-max, line 447: Please fix the formatting.

jihyeonyi: I've run black and check format with flake8 to follow pep8 format. 
If it is insufficient, can you give me more hints to resolve this?

datumaro/plugins/splitter.py
jihyeonyi: Then, I'll add test codes to check the results are reproducible.

datumaro/plugins/splitter.py
jihyeonyi: Thank you for your comment. I'll change the code according to your comment.

tests/test_splitter.py
zhiltsov-max: The code follows the rules in these specs, but it does not mean it is comfortable to read. Splitting a line on each parameter does not help with readability, in my opinion.

jihyeonyi: I see. I've written bbox info in one line like below, but it seems that 'black' modifies these automatically.
```python
Bbox(
    1, 1, 2, 2, 
    label=...
)
```
I should remove 'black' from pre-commit hook or find another way to resolve this.
Anyway, I'll apply your comment. Thanks.

datumaro/plugins/splitter.py
jihyeonyi: Actually, the gallery and query sets are not subsets. They're a kind of tag. 
I need more time to resolve this, so let's leave this as it is at the moment.
I'll resolve this issue while applying the command line feature.

jihyeonyi:
I've rebased and squashed some commits. 
I'd like to resolve the gallery/query tagging issue and support command line execution in the future pull request.

zhiltsov-max (approved):

zhiltsov-max (approved):

Issue #77, opened 2021-01-07
VGGFace2 as FR dataset


jihyeonyi:
VGGFace2 is basically FR (Face recognition) dataset, 
but our implementation only supports landmark and bounding box annotations.
I think we should add Label(...) so it can be used for FR task.

And many fr data has the same structure except the landmark information.
That is, a directory represents a person and it contains images for the person.
So if we make landmark annotations to be optional, we can support lots of FR data.

In summary, 
1. we need Label(...) annotations
2. (optional) I suggest making the landmark/bbox annotations to be optional.

zhiltsov-max:
Hi, if I understand right, you're suggesting: 
- to consider directory names as labels
- to produce labels, when no bbox information available

How about reading in `ImageNet` format (at least, as a temporary solution)? You can also apply `anns_to_labels` transform to create a classification dataset from a detection one.

jihyeonyi:
Yes, you're right.
I've tried `ImageNet` format, but it requires an additional filename format. 
So, I can't read VGGFace2 data with it.
Thank you for suggesting `anns_to_labels`. I'll check it later.

PR #82, opened 2021-01-13
Fixed VGGFace2

yasakova-anastasia:
### Summary

- Fixed the problem #77 (add labels; landmarks/bboxes are optional, but the .csv file with image names must be)

### Checklist

### License

```python
#
```


datumaro/plugins/vgg_face2_format.py
```diff
@@ -43,16 +57,29 @@ def _load_items(self, path):
...
+            if 0 < len(label):
+                item_id = osp.basename(item_id)
+
+            points = None
             if len([p for p in row if row[p] == '']) == 0 and len(row) == 11:
-                annotations.append(Points(
-                    [float(row[p]) for p in row if p != 'NAME_ID']))
-            if item_id in items and 0 < len(annotations):
+                points = Points(
+                    [eval(row[p]) for p in row if p != 'NAME_ID'])
```
zhiltsov-max, line 67: `eval()` here is a security breach. Why is it used here?

`Points` can have a label too.

`Points` also need to be in a group with `Bbox`, and the `Bbox`, being a leader of the group, should have the same `id` as the `group`. `Points` and other additional descriptors of the object may, or may not have the same `id`.
```
Bbox(id=n, group=n)
Points(group=n)
```

datumaro/plugins/vgg_face2_format.py
```diff
@@ -43,16 +57,29 @@ def _load_items(self, path):
...
                 annotation = items[item_id].annotations
-                annotation.append(annotations[0])
+                if points is not None:
+                    annotation.append(points)
             else:
+                image_path = osp.join(self._dataset_dir, self._subset, label,
+                    item_id + VggFace2Path.IMAGE_EXT)
+                annotations = []
+                if 0 < len(label):
+                    label = self._categories[AnnotationType.label].find(label)[0]
```
zhiltsov-max, line 78: I think, the file with labels would be useful, if it denoted the mapping between folder names and labels. If there is no file, we just use folder names. The same thing as in `ImageNet` - folders are called `n12345`, but there is a file with mapping to normal descriptive names.

datumaro/plugins/vgg_face2_format.py
```diff
@@ -43,16 +57,29 @@ def _load_items(self, path):
...
+            points = None
             if len([p for p in row if row[p] == '']) == 0 and len(row) == 11:
-                annotations.append(Points(
-                    [float(row[p]) for p in row if p != 'NAME_ID']))
-            if item_id in items and 0 < len(annotations):
+                points = Points(
+                    [eval(row[p]) for p in row if p != 'NAME_ID'])
+
+            if item_id in items:
                 annotation = items[item_id].annotations
```
zhiltsov-max, line 70: Lines 81-82 expect `annotations`.

datumaro/plugins/vgg_face2_format.py
```diff
@@ -43,16 +57,29 @@ def _load_items(self, path):
...
-                annotation.append(annotations[0])
+                if points is not None:
```
zhiltsov-max, line 71: Lines 81-82 should cover this.

datumaro/plugins/vgg_face2_format.py
```diff
@@ -43,16 +57,29 @@ def _load_items(self, path):
...
+                    annotation.append(points)
             else:
+                image_path = osp.join(self._dataset_dir, self._subset, label,
+                    item_id + VggFace2Path.IMAGE_EXT)
+                annotations = []
+                if 0 < len(label):
```
zhiltsov-max, line 77: Let's remove 1 step of indentation, so that it was:
```
if not exists:
     create item
annotations = item.annotations
```

datumaro/plugins/vgg_face2_format.py
```diff
@@ -43,16 +57,29 @@ def _load_items(self, path):
 
         for row in landmarks_table:
             item_id = row['NAME_ID']
-            image_path = osp.join(self._dataset_dir, self._subset,
-                item_id + VggFace2Path.IMAGE_EXT)
-            annotations = []
+            label = osp.dirname(item_id)
```
zhiltsov-max, line 60: Looking at relative item paths, I think, we could use only the first directory name as a label, and only if there is a corresponding label name mapping entry. `item_id` is just a full path. This is a topic for future discussion.

datumaro/plugins/vgg_face2_format.py
```diff
@@ -64,9 +91,18 @@ def _load_items(self, path):
...
                     item_id = row['NAME_ID']
+                    label = osp.dirname(item_id)
+                    if 0 < len(label):
+                        item_id = osp.basename(item_id)
+
                     annotations = items[item_id].annotations
-                    annotations.append(Bbox(int(row['X']), int(row['Y']),
-                        int(row['W']), int(row['H'])))
+                    label = self._categories[AnnotationType.label].find(label)[0]
+                    if label != None:
```
zhiltsov-max, line 100: You can just pass `None`, when there is no label. I think, we need to replace existing `Label` with `Bbox`. Or we need to only generate `Label`s when there is no other annotations.

zhiltsov-max (changes requested):

jihyeonyi (changes requested):
Most FR researchers usually align and crop faces before the training as a pre-process.
So in the training time, they only use images without the '.csv' file.
So how about generating Lable annotation even if '.csv' file isn't provided.
Or another way to handle this is to create a new extractor for the general FR dataset.

zhiltsov-max:
> Most FR researchers usually align and crop faces before the training as a pre-process.
> So in the training time, they only use images without the '.csv' file.
> So how about generating Lable annotation even if '.csv' file isn't provided.
> Or another way to handle this is to create a new extractor for the general FR dataset.

Isn't it exactly `ImageNet` classification format?

datumaro/plugins/vgg_face2_format.py
yasakova-anastasia: `eval()` is needed for `Bbox`. It is necessary that from `str` `int` be cast to `int`, `float` to `float`. From `Points` `eval()` I need to be replaced with `float`.

We can go through folders and create `Label` (if `save_images is true`). Then go through the .csv files and add `Bbox` and `Points`, combining into groups (without labels inside them).

zhiltsov-max: > It is necessary that from str int be cast to int, float to float.

Why do we need such a cast for coordinates? All the coordinates are supposed to be `float`.


> We can go through folders ...

I think, it should be done vise-versa - we parse boxes and points, and finish, or parse labels, if there is no boxes. I don't think we need to have both labels and boxes simultaneously in a group, it looks like overkill here. Preferring boxes over labels will simplify exporting to other formats.

datumaro/plugins/vgg_face2_format.py
yasakova-anastasia: I agree, this file is not needed. If `save_images is false`, we create empty folders. But this does not work in CVAT, because empty folders are not saved in the zip archive.

zhiltsov-max: It is important notion about CVAT, but it does not contradict anything I said. Do we need to maintain empty label folders from CVAT?

What I suggested was allowing to specify label-directory mapping in these files (`ImageNet` `synsets.txt`-style), but it looks like a good candidate for `ImageNet` format extension, though.

tests/test_vgg_face2_format.py
```diff
@@ -51,21 +53,37 @@ def test_can_save_and_load(self):
...
+                    Bbox(0, 2, 4, 2, label=0),
+                    Points([4.23, 4.32, 5.34, 4.45, 3.54,
+                        3.56, 4.52, 3.51, 4.78, 3.34], label=0),
+                ]
+            ),
+        ], categories=['label_0'])
+
+        with TestDir() as test_dir:
+            VggFace2Converter.convert(source_dataset, test_dir, save_images=False)
+            parsed_dataset = VggFace2Importer()(test_dir).make_dataset()
```
zhiltsov-max, line 84: Please, use `Dataset.import_from` in new tests.

jihyeonyi:
@zhiltsov-max 

> > @jihyeonyi,
> > 
> > Most FR researchers usually align and crop faces before the training as a pre-process.
> > So in the training time, they only use images without the '.csv' file.
> > So how about generating Lable annotation even if '.csv' file isn't provided.
> > Or another way to handle this is to create a new extractor for the general FR dataset.
> 
> Isn't it exactly ImageNet classification format?

Technically you're right if ImageNet supports more image formats.
But is it a little bit strange if the user should use `ImageNet` format for `VggFace2`?
I'd like to know your opinion.

datumaro/plugins/vgg_face2_format.py
yasakova-anastasia: I don't quite understand what you are suggesting. 
`synsets.txt` in `ImageNet_txt` format is the same as here. It is needed to create `categories` correctly when we do not save images or when not all labels  are affected.
Do you propose in this file to describe the mapping of the folder name and the object name (n01234 cat)?

zhiltsov-max: > Do you propose in this file to describe the mapping of the folder name and the object name (n01234 cat)?

Yes, like this: 
https://gist.github.com/yrevar/942d3a0ac09ec9e5eb3a#gistcomment-2575723

zhiltsov-max:
@jihyeonyi, yes, it is strange. Actually, I think, we need to slightly change the structure of `Extractor` to simplify this kind of integration between sub-formats (ImageNet dataset uses detection from from PASCAL VOC, CamVid uses segmentation from VOC etc.).

jihyeonyi:
> @jihyeonyi, yes, it is strange. Actually, I think, we need to slightly change the structure of `Extractor` to simplify this kind of integration between sub-formats (ImageNet dataset uses detection from from PASCAL VOC, CamVid uses segmentation from VOC etc.).

Good idea. I'm of the same opinion.
So, we can expect this issue to be solved next time, right?

zhiltsov-max:
@jihyeonyi, it definitely won't be done in this PR, more likely in Q1.

datumaro/plugins/vgg_face2_format.py
yasakova-anastasia: If `item.id` is full path, there is a problem in one test.
Let's say `item.id` is not a full path. Convert with label and get `row['NAME_ID'] = label/item.id`. When exporting, we save the full path and get a mismatch.

zhiltsov-max: How about using only the first directory as class, and all the rest path as the `id`? It looks like a working solution, if there is a guarantee, that the first directory actually is the label (i.e. we prohibit using an arbitrary directory with images).

datumaro/plugins/vgg_face2_format.py
```diff
@@ -43,30 +62,46 @@ def _load_items(self, path):
...
             item_id = row['NAME_ID']
-            image_path = osp.join(self._dataset_dir, self._subset,
-                item_id + VggFace2Path.IMAGE_EXT)
-            annotations = []
+            label = None
+            if '/' in item_id:
+                label_name = item_id.split('/')[0]
+                label = self._categories[AnnotationType.label].find(label_name)[0]
+                if label is not None:
+                    item_id = item_id[len(label_name) + 1:]
```
zhiltsov-max, line 70: Why this can happen?

zhiltsov-max (dismissed):
The only problem I see is that the former issue is not resolved.

datumaro/plugins/vgg_face2_format.py
yasakova-anastasia: If for example item.id = a/1 and there is no label

jihyeonyi (approved):

datumaro/plugins/vgg_face2_format.py
yasakova-anastasia: Maybe add a folder for unlabeled items (like in ImageNet)?

zhiltsov-max: I think it is a good idea, because otherwise we require `labels.txt`.

PR #96, opened 2021-01-20
Support for ICDAR13/15 dataset format

yasakova-anastasia:
### Summary

- Add support for ICDAR13/15 dataset format (tasks: word_recognition 2013 and text_localization 2013/2015)

### How to test

Unit tests
### Checklist

### License

```python
#
```


datumaro/plugins/icdar_format/extractor.py
```diff
@@ -0,0 +1,173 @@
...
+        labels = []
+        if osp.isfile(path):
+            with open(path, encoding='utf-8') as labels_file:
+                labels = [s.strip() for s in labels_file]
+        else:
+            if self._task is IcdarTask.word_recognition:
+                paths = [self._path]
+            else:
+                paths = [p for p in glob(osp.join(self._path, '*.txt'),
+                    recursive=True)]
```
zhiltsov-max, line 45: Looks like `recursive` is not needed here.

datumaro/plugins/icdar_format/extractor.py
```diff
@@ -0,0 +1,173 @@
...
+            for path in paths:
+                with open(path, encoding='utf-8') as f:
+                    for line in f:
+                        objects = line.split()
+                        if len(objects) == 1:
+                            objects = line.split(',')
+
+                        if len(objects) == 9:
```
zhiltsov-max, line 53: All these variants are valid and used in the dataset? 
9-values format looks like the segmentation task format. 
2-value format should probably produce `Caption` annotations (word recognition task) and have no labels.

They also say that 

> Please note that the escape character (\) is used for double quotes and backslashes.

so we probably need to support this.

datumaro/plugins/icdar_format/extractor.py
```diff
@@ -0,0 +1,173 @@
...
+                        if label[0] == '"' and label[-1] == '"':
+                            label = label[1:-1]
+                        labels.append(label)
+
+        categories[AnnotationType.label] = \
+            LabelCategories().from_iterable(sorted(labels))
+
+        return categories
+
+    def _load_items(self):
```
zhiltsov-max, line 70: It seems like there is nothing in common in these subformats. Wouldn't it be better to split their implementations?

datumaro/plugins/icdar_format/extractor.py
```diff
@@ -0,0 +1,173 @@
...
+        for task, extractor_type, task_dir in cls._TASKS:
+            task_path = osp.join(path, task_dir)
+            if not osp.isdir(task_path):
+                continue
+            if task is IcdarTask.word_recognition:
+                sources += cls._find_sources_recursive(
+                    task_path, '.txt', extractor_type, file_filter=lambda p: \
+                    osp.basename(p) != IcdarPath.VOCABULARY_FILE)
+            elif task is IcdarTask.text_localization:
+                subset_paths = [p for p in glob(osp.join(task_path, '**'))
```
zhiltsov-max, line 169: Why can't we use `find_sources_recursive` here?

tests/test_icdar_format.py
```diff
@@ -0,0 +1,154 @@
...
+                    Bbox(1, 3, 6, 10, label=1),
+                    Label(1),
+                ]),
+            DatasetItem(id=3, subset='train',
+                annotations=[
+                    Bbox(2, 4, 6, 10),
+                    Bbox(0, 2, 5, 9, label=0),
+                ]),
+            DatasetItem(id=4, subset='train'),
+        ], categories={
```
zhiltsov-max, line 146: You could write `['label_0', 'label_1']` for the same result.

tests/test_icdar_format.py
```diff
@@ -0,0 +1,154 @@
...
+                    Bbox(0, 0, 2, 3, label=2),
+                    Bbox(3, 3, 2, 3, label=1),
+                ]
+            ),
+        ], categories={
+            AnnotationType.label: LabelCategories.from_iterable(
+                ['FOOD', 'LION', 'RED']),
+        })
+
+        dataset = Project.import_from(DUMMY_DATASET_DIR, 'icdar') \
```
zhiltsov-max, line 59: Please, use `Dataset.import_from`

zhiltsov-max (changes requested):
Probably, the dataset should include 3 sub-formats. 1 produces Captions, 2 produces Bboxes, 3 produces grouped Masks.

datumaro/plugins/icdar_format/converter.py
```diff
@@ -0,0 +1,138 @@
+# Copyright (C) 2020 Intel Corporation
+#
+# SPDX-License-Identifier: MIT
+
+import os
+import os.path as osp
+
+from datumaro.components.converter import Converter
+from datumaro.components.extractor import AnnotationType
+from datumaro.plugins.icdar_format.format import IcdarPath, IcdarTask
```
zhiltsov-max, line 10: Plugins should use relative import like `.format`

datumaro/plugins/icdar_format/extractor.py
yasakova-anastasia: 9-values and 5-values ​​formats are needed for text localization tasks in [Incidental Scene Text](https://rrc.cvc.uab.es/?ch=4&com=tasks) and [Focused Scene Text](https://rrc.cvc.uab.es/?ch=2&com=tasks) (supported in AC).

yasakova-anastasia:
> Probably, the dataset should include 3 sub-formats. 1 produces Captions, 2 produces Bboxes, 3 produces grouped Masks.

Yes, there is still a task of text segmentation, but it is not in AC and it is a little strange: different colors correspond to the same symbols, that is, each symbol is a separate color.
For example:
```
212 104 103 974 392 891 302 1056 493 "w"
47 22 217 1146 398 1065 306 1228 497 "w" 
```
(first three numbers - color)

zhiltsov-max:
Please take a look at [format description](https://rrc.cvc.uab.es/?ch=2&com=tasks). As there is no classes, I consider these colors just as a color filter/threshold in the mask bbox for a character (instance). Maybe we could consider the dictionary as a 'label set' or just all the characters, but it doesn't look like a good idea, an attribute value looks like a better option.

yasakova-anastasia:
Then there will be no categories. The color and symbol will be stored as mask attributes.
There is also a bounding box and a center point. They need to be grouped with a mask. Also, masks must be grouped by words. It is possible to store bbox and center point as attributes, but I think this is a bad idea. Or we can use label in Mask, Bbox and Points as a group (just a number).

zhiltsov-max:
> There is also a bounding box and a center point ...

I'd consider these values only as technical information for finding mask section on the image. I don't think there is any need in producing BBox annotation along with Mask, because it can easily be computed from masks.

> Also, masks must be grouped by words.

Maybe, if we knew the words. If word annotations are split by a line - ok, there should be no problem with group saving and parsing.

datumaro/plugins/icdar_format/converter.py
```diff
@@ -0,0 +1,211 @@
...
+
+    def is_empty(self):
+        return len(self.annotations) == 0
+
+class _TextLocalizationConverter():
+    def __init__(self):
+        self.annotations = {}
+
+    def save_categories(self, save_dir, categories):
+        vocabulary_file = osp.join(save_dir,
```
zhiltsov-max, line 45: I still don't known if is a good idea or not to store text in the `label` field. It might be better to introduce `CaptionCategories` with vocabulary, if it is provided, or just use an attribute for text.

datumaro/plugins/icdar_format/converter.py
```diff
@@ -0,0 +1,211 @@
...
+        annotation = ''
+        for ann in item.annotations:
+            if ann.type == AnnotationType.bbox:
+                annotation += '%s %s %s %s' % (ann.x, ann.y,
+                    ann.x + ann.w, ann.y + ann.h)
+                if ann.label is not None:
+                    annotation += ' %s' % \
+                        categories[AnnotationType.label][ann.label].name
+            elif ann.type == AnnotationType.polygon:
+                annotation += ','.join(str(p) for p in ann.points)
```
zhiltsov-max, line 64: This approach should work for boxes too.

datumaro/plugins/icdar_format/converter.py
```diff
@@ -0,0 +1,211 @@
...
+    def save_annotations(self, item, categories):
+        annotation = ''
+        colormap = [(255, 255, 255)]
+        group = 0
+        for ann in item.annotations:
+            if ann.type == AnnotationType.mask:
+                char = ''
+                if ann.group != group:
+                    annotation += '\n'
+                    group += 1
```
zhiltsov-max, line 98: I think, there are no guarantees the annotations will be sorted by group number. Take a look at `utils.annotation_utils.find_instances` for reference.

datumaro/plugins/icdar_format/converter.py
```diff
@@ -0,0 +1,211 @@
...
+                            for p in ann.attributes['color'])
+                    else:
+                        annotation += '- - -'
+                    if 'center' in ann.attributes:
+                        annotation += ' '
+                        annotation += ' '.join(str(p)
+                            for p in ann.attributes['center'])
+                    else:
+                        annotation += ' - -'
+                image = ann.image
```
zhiltsov-max, line 116: Masks have a method for this.

datumaro/plugins/icdar_format/converter.py
```diff
@@ -0,0 +1,211 @@
...
+                bbox = find_mask_bbox(image)
+                annotation += ' %s %s %s %s' % (bbox[0], bbox[1],
+                    bbox[0] + bbox[2], bbox[1] + bbox[3])
+                annotation += ' \"%s\"' % char
+                annotation += '\n'
+
+        masks = [a for a in item.annotations
```
zhiltsov-max, line 123: Probably, we can move this higher, and replace the first loop iterable with this.

datumaro/plugins/icdar_format/extractor.py
```diff
@@ -0,0 +1,261 @@
...
+            self._path)
+        self._items = list(task_extractor.load_items(self._path, self._subset,
+            self._categories).values())
+
+    def _make_task_extractor(self, task):
+        if task not in self._TASK_EXTRACTORS:
+            raise NotImplementedError()
+        return self._TASK_EXTRACTORS[task]()
+
+class IcdarWordRecognitionExtractor(_IcdarExtractor):
```
zhiltsov-max, line 220: What is the point in classes above? We could put their methods in these classes and just call them in the base class.

datumaro/plugins/icdar_format/extractor.py
```diff
@@ -0,0 +1,261 @@
...
+                        objects = line.split(',')
+                    if 8 <= len(objects):
+                        points = [float(objects[p]) for p in range(8)]
+                        label = None
+                        if len(objects) == 9:
+                            label_name = objects[8]
+                            if label_name[0] == '\"' and label_name[-1] == '\"':
+                                label_name = label_name[1:-1]
+                            label = \
+                                _categories[AnnotationType.label]._indices[label_name]
```
zhiltsov-max, line 103: Use `.find()` to get label id.

datumaro/plugins/icdar_format/converter.py
yasakova-anastasia: Please, explain what you mean.

datumaro/plugins/icdar_format/converter.py
```diff
@@ -0,0 +1,193 @@
...
+        self.masks = {}
+
+    def save_annotations(self, item):
+        annotation = ''
+        colormap = [(255, 255, 255)]
+        masks = [a for a in item.annotations
+            if a.type == AnnotationType.mask]
+        annotations = {}
+        for ann in find_instances(masks):
+            if ann[0].group in annotations:
```
zhiltsov-max, line 78: Group=0 means 'no group', so such annotations should not be joined.

yasakova-anastasia: Probably, I need to correct the group numbering in the tests and in the extractor. But here items with `group = 0` will just be a separate group.

zhiltsov-max: It's just a group of one item.

yasakova-anastasia: In our case, the group also shows the order in the text annotation. If we make it so that each one item group = 0, when converting, we get a different order in the annotation. I think this is not very good.

tests/test_icdar_format.py
```diff
@@ -0,0 +1,176 @@
...
+        dataset = Dataset.import_from(osp.join(DUMMY_DATASET_DIR, 'text_localization'), 'icdar')
+
+        compare_datasets(self, expected_dataset, dataset)
+
+    def test_can_import_masks(self):
+        expected_dataset = Dataset.from_iterable([
+            DatasetItem(id='1', subset='train',
+                image=np.ones((2, 5, 3)),
+                annotations=[
+                    Mask(id=0, image=np.array([[0, 1, 1, 0, 0], [0, 0, 0, 0, 0]]), label=1,
```
zhiltsov-max, line 69: What is the point in `center` attribute? Does it help with anything? It can be easily computed from the mask as `mask.get_bbox()` etc.

If any annotation has `label` property set, it should be an index from `LabelCategories`.

datumaro/plugins/icdar_format/converter.py
```diff
@@ -0,0 +1,193 @@
...
+                        for p in ann.attributes['color'])
+                else:
+                    annotation += '- - -'
+                if 'center' in ann.attributes:
+                    annotation += ' '
+                    annotation += ' '.join(str(p)
+                        for p in ann.attributes['center'])
+                else:
+                    annotation += ' - -'
+            bbox = find_mask_bbox(ann.image)
```
zhiltsov-max, line 104: Mask class has a method for getting a bbox. Also, for RLE-encoded masks it performs better.

tests/test_icdar_format.py
yasakova-anastasia: > What is the point in `center` attribute? Does it help with anything? It can be easily computed from the mask as `mask.get_bbox()` etc.
I tried to find the center by `bbox`, but this does not always work on data of this format. Maybe the center of the symbol is not always the center of the `bbox`.

PR #102, opened 2021-02-08
Inplace save & update dataset

zhiltsov-max:
### Summary

- `Dataset` operations are finally made lazy
- Transforms can be performed lazily for a `Dataset`
- Dataset implements caching for input source. Multiple sources are immediately merged.
- The order of elements in a `Dataset` is maintained, but is not guaranteed to be the same after saving and loading
- Added partial saving interface for datasets (for in-place dataset updates in the same format)
- Implemented partial saving for `Datumaro` format
- Extended `Dataset` interface with cache control, changed data info, source path and format info
- `Dataset.get()` returns `None` instead of raising an exception when the item doesn't exist
- Supported `in` operator for `Dataset`
- Added `get` operation for `Extractor`
- Added type annotations for `Dataset` class
- Extended API model with new interfaces
- `Converter` interface is extended by optional operation to support partial data update (`patch()`). The default implementation uses the regular full-dataset saving.
- Added specific error types to be used instead of generic `Exception`
- `Dataset` can track updates and generate patches. Transform is considered updating the whole dataset
- `Dataset.get_subset` provides modifiable slices

TBD: 
- [ ] update docs
- [ ] update CVAT
- [ ] implement partial save in formats

### Checklist

### License

```python
#
```


nmanovic:
@zhiltsov-max , do we have any difficulties to solve that: "The order of elements in a Dataset is maintained, but is not guaranteed to be the same after saving and loading"?

I'm not sure that it is critical, but I prefer deterministic behavious if it is easy to achieve.

zhiltsov-max:
@nmanovic, if a format represents a dataset with several subset files, it is impossible to reproduce initial item ordering.

Example:

```
Dataset:
item(1, 'train')
item(2, 'val')
item(3, 'train')

.save():

train_list.txt
val_list.txt

.load()

item(1, 'train')
item(3, 'train')
item(2, 'val')

```

tests/test_dataset.py
```diff
@@ -189,16 +195,355 @@ def test_cant_join_different_categories(self):
...
+            ])
+            dataset.save(path)
+            ts1_a = os.stat(osp.join(path, 'annotations', 'a.json')).st_mtime_ns
+            ts1_b = os.stat(osp.join(path, 'annotations', 'b.json')).st_mtime_ns
+            ts1_c = osp.isfile(osp.join(path, 'annotations', 'c.json'))
+
+            dataset = Dataset.load(path)
+            dataset.put(DatasetItem(2, subset='a'))
+            dataset.remove(3, 'c')
+            time.sleep(1)
```
nmanovic, line 228: Why do we need sleep here?

zhiltsov-max: In the test there are 3 subset files generated - a, b and c.
a should be updated, b should be kept the same, c should be removed.

There is a comparison between file modification times. Without the sleep the test was passing too fast, so that the comparison failed. The other way to implement the check can be through comparison of file hashes or something more clever.

nmanovic:
@zhiltsov-max , should we update documentation? Are you planning to add some short tutorials for new use cases?

zhiltsov-max:
@nmanovic, I'd prefer to update documentation after new API for operations are introduced, otherwise the changes are hard to perceive. Small catchy examples were added earlier, they still work - but now they also have good performance because of added transparent caching. Thorough documentation will be added with r0.2 (VCS) / r0.3 (stable API) and stable API introduction.

PR #103, opened 2021-02-10
Fixes in WiderFace and VGGFace2 formats

yasakova-anastasia:
### Summary

WiderFace:
- Fix type casting of points and attributes
- Add the ability to save the label to bbox (for CVAT)

VGGFace2:
- Add check for wrong number of points
- Add the ability to save multiple groups

### Checklist

### License

```python
#
```


datumaro/plugins/vgg_face2_format.py
```diff
@@ -75,13 +75,24 @@ def _load_items(self, path):
                     row['NAME_ID'] + VggFace2Path.IMAGE_EXT)
                 items[item_id] = DatasetItem(id=item_id, subset=self._subset,
                     image=image_path)
+            group = 0
```
zhiltsov-max, line 78: Is this situation valid for this format? Shouldn't a face have only one set of landmarks? Maybe, throwing an error could be better?

zhiltsov-max (approved):

datumaro/plugins/vgg_face2_format.py
yasakova-anastasia: I agree, this is strange.

I have another question. I'm not sure that we should automatically group annotations. If we save a dataset in CVAT in this format, then we get grouped annotations. Probably not bad. But the tests fail when comparing datasets.

PR #148, opened 2021-03-05
Fix dataset formats (cyrilic and spaces in filename)

yasakova-anastasia:
### Summary

Resolves [openvinotoolkit/cvat#2840](https://github.com/openvinotoolkit/cvat/issues/2840)
Resolves [openvinotoolkit/cvat#2922](https://github.com/openvinotoolkit/cvat/issues/2922)
- Added the ability to work with file names containing Cyrillic and spaces

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/plugins/camvid_format.py
+++ b/datumaro/plugins/camvid_format.py
@@ -61,7 +61,7 @@ class CamvidPath:
     SEGM_DIR = "annot"
     IMAGE_EXT = '.jpg'
     MASK_EXT = '.png'
-    PATTERN = re.compile(r'(.+[.]\S+) (.+[.]\S+)?')
+    PATTERN = re.compile(r'(.+\.\S+)(?:\s+(.+\.\S+)?)?\s*')
 
 
 def parse_label_map(path):
@@ -160,9 +160,9 @@ class CamvidExtractor(SourceExtractor):
         items = {}
         with open(path, encoding='utf-8') as f:
             for line in f:
-                search = CamvidPath.PATTERN.search(line)
+                search = CamvidPath.PATTERN.search(line.strip())
                 if search:
-                    objects = CamvidPath.PATTERN.search(line).groups()
+                    objects = search.groups()
                 else:
                     raise Exception("Line %s: invalid path format" % line)
                 image = objects[0]
--- a/datumaro/plugins/camvid_format.py
+++ b/datumaro/plugins/camvid_format.py
@@ -61,7 +61,7 @@ class CamvidPath:
     SEGM_DIR = "annot"
     IMAGE_EXT = '.jpg'
     MASK_EXT = '.png'
-    PATTERN = re.compile(r'(.+[.]\S+) (.+)?')
+    PATTERN = re.compile(r'(.+[.]\S+) (.+[.]\S+)?')
 
 
 def parse_label_map(path):
--- a/datumaro/plugins/camvid_format.py
+++ b/datumaro/plugins/camvid_format.py
@@ -7,6 +7,7 @@ import os
 import os.path as osp
 from collections import OrderedDict
 from enum import Enum
+import re
 
 import numpy as np
 from datumaro.components.converter import Converter
@@ -57,6 +58,7 @@ class CamvidPath:
     LABELMAP_FILE = 'label_colors.txt'
     SEGM_DIR = "annot"
     IMAGE_EXT = '.png'
+    PATTERN = re.compile(r'(.+[.]\S+) (.+)?')
 
 
 def parse_label_map(path):
@@ -155,16 +157,8 @@ class CamvidExtractor(SourceExtractor):
         items = {}
         with open(path, encoding='utf-8') as f:
             for line in f:
+                objects = CamvidPath.PATTERN.search(line).groups()
                 objects = line.split()
-                if 2 < len(objects):
-                    if len(objects) % 2:
-                        raise Exception("Line %s: image and gt file must have  "
-                            "the same name" % line)
-                    else:
-                        mid = int(len(objects) / 2)
-                        objects[0] = ' '.join(objects[i] for i in range(mid))
-                        objects[1] = ' '.join(objects[i] for i in range(mid, 2 * mid))
-                        objects = objects[:2]
                 image = objects[0]
                 item_id = ('/'.join(image.split('/')[2:]))[:-len(CamvidPath.IMAGE_EXT)]
                 image_path = osp.join(self._dataset_dir,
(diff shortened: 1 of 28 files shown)


zhiltsov-max:
Are any other formats affected? Should we add such tests in all formats?

datumaro/plugins/voc_format/converter.py
```diff
@@ -367,8 +367,10 @@ def save_action_lists(self, subset_name, action_list):
         items = {k: True for k in action_list}
         if self._patch and osp.isfile(ann_file):
             self._get_filtered_lines(ann_file, self._patch, subset_name, items)
-        with open(ann_file, 'w') as f:
+        with open(ann_file, 'w', encoding='utf-8') as f:
             for item in items:
+                if 1 < len(item.split()):
```
zhiltsov-max, line 372: I don't think this solution is right. All the VOC subformats should handle such names without changes. The only problematic subformat should be `Layout`, where only parsing should be changed.

zhiltsov-max (changes requested):

datumaro/plugins/imagenet_txt_format.py
```diff
@@ -49,9 +49,14 @@ def _load_items(self, path):
         items = {}
         with open(path, encoding='utf-8') as f:
             for line in f:
-                item = line.split()
-                item_id = item[0]
-                label_ids = [int(id) for id in item[1:]]
+                item = line.split('\"')
```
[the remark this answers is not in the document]
yasakova-anastasia: Could you explain how best to do it? Now we go through `if` when we have two characters `"`. In all other cases we go through `else`.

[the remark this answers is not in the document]
zhiltsov-max: For instance, we could raise an error if there is any unexpected number of quotes - 1 or 3+.

datumaro/plugins/voc_format/extractor.py
```diff
@@ -78,11 +84,13 @@ def _load_annotations(self):
         anno_files = [s for s in dir_items(task_dir, '.txt')
             if s.endswith('_' + osp.basename(self._path))]
         for ann_filename in anno_files:
-            with open(osp.join(task_dir, ann_filename)) as f:
+            with open(osp.join(task_dir, ann_filename), encoding='utf-8') as f:
                 label = ann_filename[:ann_filename.rfind('_')]
                 label_id = self._get_label_id(label)
                 for line in f:
-                    item, present = line.split()
+                    objects = line.split()
```
zhiltsov-max, line 91: Consider using `rsplit(maxsplit=1)`

datumaro/plugins/camvid_format.py
```diff
@@ -61,6 +61,7 @@ class CamvidPath:
     SEGM_DIR = "annot"
     IMAGE_EXT = '.jpg'
     MASK_EXT = '.png'
+    PATTERN = re.compile(r'(.+[.]\S+) (.+[.]\S+)?')
```
zhiltsov-max, line 64: Please don't expect input datasets to be generated by Datumaro. Instead, consider meaningful relaxations to the format. Expect and fix little and insignificant errors you can think of.

I think, the pattern can be improved with `r(.+\.\S+)(?:\s+(.+\.\S+)?)\s*`
- to allow any number of whitespaces between entries
- to allow any number of trailing whitespaces (.strip() would also work)
- to allow lines like `a/b/c .d  \n`

datumaro/plugins/voc_format/extractor.py
```diff
@@ -58,8 +58,24 @@ def _load_categories(dataset_path):
 
     @staticmethod
     def _load_subset_list(subset_path):
-        with open(subset_path) as f:
-            return [line.split()[0] for line in f]
+        subset_list = []
+        with open(subset_path, encoding='utf-8') as f:
+            for line in f:
+                dirname = osp.basename(osp.dirname(subset_path))
```
zhiltsov-max, line 64: I think this is better to be done with check of `self._task`. These files are not required to be inside the default task lists directory - it is just a default location.

datumaro/plugins/voc_format/extractor.py
```diff
@@ -58,8 +58,24 @@ def _load_categories(dataset_path):
...
+                if dirname == VocPath.TASK_DIR[VocTask.person_layout]:
+                    objects = line.split('\"')
+                    if 1 < len(objects):
```
zhiltsov-max, line 67: `0 <` ? 1 will break this code.

yasakova-anastasia (commented):
.

datumaro/plugins/camvid_format.py
```diff
@@ -159,13 +160,17 @@ def _load_items(self, path):
         items = {}
         with open(path, encoding='utf-8') as f:
             for line in f:
-                objects = line.split()
+                search = CamvidPath.PATTERN.search(line.strip('\n'))
```
zhiltsov-max, line 163: Why we strip only `\n`?

zhiltsov-max (dismissed):

datumaro/plugins/camvid_format.py
yasakova-anastasia: We must always have a space after the first path. In [test without masks](https://github.com/openvinotoolkit/datumaro/blob/944e9f9f8ae2b10aea5eff9573e2de65844538c0/tests/test_camvid_format.py#L148), after `strip()`, we get the `test/a/b/1.png` without a space. As a result, search returns None.

yasakova-anastasia: There is no such requirement. But I'm having difficulty with a regex that supports a string without a space. If we replace `+` with `*` in the regex (`r'(.+\.\S+)(?:\s*(.+\.\S+)?)\s*'`), we cannot find two paths.

zhiltsov-max: Remove the last `\s*`

yasakova-anastasia: This doesn't work as everything is recorded in the first group.

zhiltsov-max: Did you try `(.+\.\S+)(?:\s+(.+\.\S+)?)?` ?

yasakova-anastasia: Yes, also everything is recorded in one group

zhiltsov-max: If we don't have a mask, it is what we need to obtain, isn't it?

yasakova-anastasia: Yes. But everything is recorded in one group even we have a mask.

zhiltsov-max: How about eager capture in the first group then? `(.+?\.\S+)(?:\s+(.+\.\S+)?)?`

yasakova-anastasia: It works, thanks!

PR #160, opened 2021-03-15
Add validator for classification & detection tasks

seongjun-park-dl:
### Summary
This PR includes:
- Validator for classification and detection tasks.
- CLI support for validator.

### How to test
`python -m unittest -v tests/test_validator.py`

### Checklist

### License

```python
#
```


tests/assets/validator_test_data/stats.py
```diff
@@ -0,0 +1,6231 @@
+# Copyright (C) 2021 Intel Corporation
```
zhiltsov-max, line 1: Can we reduce the size of testing code somehow? I wouldn't like to have a piece of write-only code.

datumaro/components/validator.py
```diff
@@ -0,0 +1,951 @@
...
+
+    def __init__(self,
+                task_type='classification',
+                ann_type=AnnotationType.label,
+                far_from_mean_thr=None):
+        self.task_type = task_type
+        self.ann_type = ann_type
+        self.far_from_mean_thr = far_from_mean_thr
+
+    def compute_statistics(self, dataset):
```
zhiltsov-max, line 52: Please try to split this function into smaller parts. Maybe, using local functions.

datumaro/components/validator.py
```diff
@@ -0,0 +1,951 @@
...
+                                            anns_far_from_mean = \
+                                                items_far_from_mean.setdefault(
+                                                    item.id, {})
+                                            anns_far_from_mean[ann.id] = val
+
+        return stats
+
+    def _check_missing_label_categories(self, stats):
+        validation_reports = []
+        description = "'LabelCategories(...)' should be defined" \
```
zhiltsov-max, line 382: Consider using and extending typed errors from `datumaro.components.errors`.

Avoid using implementation information in error messages, these class names are mostly useless for the potential customer.

zhiltsov-max (changes requested):

tests/assets/validator_test_data/stats.py
seongjun-park-dl: This is where the expected statistics/results of 'compute_statistics' method of the validator is stored (which is a rather long dictionary holding information). Are you suggesting that I should reduce the size of the test dataset (in 'tests/assets/validator_test_data/datasets.py') that corresponds with the expected stats (which would reduce the size of the expected stats)? Or are you suggesting that I should cut down the main test script in 'tests/test_validator.py'? Perhaps both?

zhiltsov-max: I suggest to think about maintaining these tests. If the answer to the questions "is it easy to maintain them? Would you like to do it on a constant basis?" is "no", then, we're probably moving in a wrong direction. I can recommend few ways to improve:
- test only what is necessary in "common", function-specific tests, and have 1-2 more complex, general tests
- do not test full text of error messages (unless it is a direct business requirement)
- minimize test scenarios, deduplicate (eg. leave only single subset instead of multiple, leave only single `DatasetItem`, if it is enough, leave only single `Bbox`, if it is enough etc.)

seongjun-park-dl: Got it! I got rid of tests/assets/validator_test_data entirely and reworked the unit test along with other changes according to your comments on the most recent commit.

zhiltsov-max:
1. Generally, looks ok. I've tried to validate VOC and this is what I got:

```
"summary": {
        "errors": 319372,
        "warnings": 81169
    },
```

2. The other thing I found is:
```
{
            "anomaly_type": "MissingBboxAnnotation",
            "description": "Item needs one or more bounding box annotations, but not found.",
            "item_id": "2007_000032",
            "severity": "warning"
        },
```
I think, it would be useful to add item subset as well (or group results by subsets)

3. This check
```
{
            "anomaly_type": "FarFromLabelMean",
            "description": "Bounding box annotation '11' in the item has a value of bounding box 'y' that is too far from the label average. (mean of 'person' label: 108.82, got '286.0').",
            "item_id": "2008_004948",
            "severity": "warning"
        },
```
looks quite strange for 'x' and 'y' bbox attributes - why images in the dataset must have anything common for them? 'width' and 'height' are more or less understandable, but it might be good to check against clusters, instead of a single value of average. Another thought is to allow setting acceptable boundaries.
'long' and 'short' are hard to understand in bbox context.

4. This check
```
{
            "anomaly_type": "UndefinedAttribute",
            "description": "Item has the attribute 'difficult' for the label 'person' which is not defined in metadata.",
            "item_id": "2011_002650",
            "severity": "error"
        },
```
should consider attributes in LabelCategories.attributes as general, i.e. applicable to all labels. Label-specific ones are written in each label.

5. As a further step, I'd recommend think of providing an option to control enabled checks.

jihyeonyi:
For the 2nd comment (grouping missing bbox cases), your idea seems really nice but I don't think we can apply the groupping function at this PR.
For now, the concept of validator is just listing all the warnings verbosely, so the user should group or filter items by parsing the validation results file.
And I can't think of a nice interface? process? to do that. 
We can implement this as a filtering function, transformer, or utility function. I don't know which is the best.

jihyeonyi:
For 3rd comment, I agree with you. 
And for the short(=min(w,h)) and long(=max(w,h)), some researchers think they are important for the detection task. 
So we include those terms.

For 4rd comment, we didn't know that. Seongjun would fix that.

And for the last comment, we'll consider that later, because it is a design issue.

seongjun-park-dl:
I've addressed 3rd and 4th comments in my most recent commit.

zhiltsov-max:
> For now, the concept of validator is just listing all the warnings verbosely, so the user should group or filter items by parsing the validation results file.

I mean, it would simplify/clarify things, if error messages contained a "subset" field, because a single id can (theoretically) be found in several subsets.


{
            "anomaly_type": "...",
            "description": "...",
            **"item_id": "2007_000032",
            "subset": "train",**
            "severity": "..."
        },

jihyeonyi:
> > For now, the concept of validator is just listing all the warnings verbosely, so the user should group or filter items by parsing the validation results file.
> 
> I mean, it would simplify/clarify things, if error messages contained a "subset" field, because a single id can (theoretically) be found in several subsets. And even if subsets do not have same item ids, it still makes the process of error localization simpler.
> 
> {
> "anomaly_type": "...",
> "description": "...",
> **"item_id": "2007_000032", "subset": "train",**
> "severity": "..."
> },

Oh, I can understand now. It's not difficult to add "subset" information and it seems useful.

seongjun-park-dl:
> > For now, the concept of validator is just listing all the warnings verbosely, so the user should group or filter items by parsing the validation results file.
> 
> I mean, it would simplify/clarify things, if error messages contained a "subset" field, because a single id can (theoretically) be found in several subsets. And even if subsets do not have same item ids, it still makes the process of error localization simpler.
> 
> {
> "anomaly_type": "...",
> "description": "...",
> **"item_id": "2007_000032", "subset": "train",**
> "severity": "..."
> },

Added the subset field into the error messages!

zhiltsov-max (dismissed):

zhiltsov-max (approved):

PR #174, opened 2021-03-18
Fixes in ICDAR dataset format

yasakova-anastasia:
### Summary

- Common ICDAR format is split into 3 sub-formats: 
  `icdar_text_segmentation`, `icdar_text_localization`, `icdar_word_recognition`
- Fixed image reading and saving in these formats
- Removed extra task directory on saving

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/plugins/icdar_format/converter.py
+++ b/datumaro/plugins/icdar_format/converter.py
@@ -10,15 +10,11 @@ from datumaro.components.extractor import AnnotationType, CompiledMask
 from datumaro.util.image import save_image
 from datumaro.util.mask_tools import paint_mask
 
-from .format import IcdarPath, IcdarTask
+from .format import IcdarPath
 
 
 class IcdarWordRecognitionConverter(Converter):
     DEFAULT_IMAGE_EXT = IcdarPath.IMAGE_EXT
-
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        self._task = IcdarTask.word_recognition
 
     def apply(self):
         for subset_name, subset in self._extractor.subsets().items():
@@ -42,10 +38,6 @@ class IcdarWordRecognitionConverter(Converter):
 
 class IcdarTextLocalizationConverter(Converter):
     DEFAULT_IMAGE_EXT = IcdarPath.IMAGE_EXT
-
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        self._task = IcdarTask.text_localization
 
     def apply(self):
         for subset_name, subset in self._extractor.subsets().items():
@@ -73,9 +65,6 @@ class IcdarTextLocalizationConverter(Converter):
 
 class IcdarTextSegmentationConverter(Converter):
     DEFAULT_IMAGE_EXT = IcdarPath.IMAGE_EXT
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        self._task = IcdarTask.text_segmentation
 
     def apply(self):
         for subset_name, subset in self._extractor.subsets().items():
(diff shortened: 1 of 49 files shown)


datumaro/plugins/icdar_format/converter.py
```diff
@@ -167,7 +169,8 @@ def apply(self):
                 for task, task_conv in task_converters.items():
                     if item.has_image and self._save_images:
                         self._save_image(item, osp.join(
-                            self._save_dir, subset_name, IcdarPath.IMAGES_DIR,
+                            self._save_dir, IcdarPath.TASK_DIR[task],
```
zhiltsov-max, line 172: Does this mean an image can be saved multiple times, if it contains annotations from few tasks? Maybe we don't need common converter then?

datumaro/plugins/icdar_format/converter.py
```diff
@@ -56,6 +56,7 @@ def write(self, path):
         os.makedirs(path, exist_ok=True)
         for item in self.annotations:
             file = osp.join(path, 'gt_' + item + '.txt')
+            os.makedirs(osp.dirname(file), exist_ok=True)
```
zhiltsov-max, line 59:
```suggestion
            file = osp.join(path, osp.dirname(item),
                'gt_' + osp.basename(item) + '.txt')
            os.makedirs(osp.dirname(file), exist_ok=True)
```

The item can have a relative path.

datumaro/plugins/icdar_format/converter.py
```diff
@@ -13,182 +13,136 @@
...
-        kwargs['tasks'] = IcdarTask.text_segmentation
         super().__init__(*args, **kwargs)
+        self._task = IcdarTask.text_segmentation
+
+    def apply(self):
+        for subset_name, subset in self._extractor.subsets().items():
+            for item in subset:
+                if item.has_image and self._save_images:
+                    self._save_image(item, osp.join(self._save_dir,
+                        IcdarPath.TASK_DIR[self._task],
```
zhiltsov-max, line 91: I don't see a point in saving into `IcdarPath.TASK_DIR[self._task]` anymore.

yasakova-anastasia: Why? When we do one task, our images folder will not be in the tasks folder. This is normal. But, for example, assets for three tasks will have one folder of images.

zhiltsov-max: Such test fails:

```
    def test_can_save_dataset_with_cyrillic_and_spaces_in_filename(self):
        expected_dataset = Dataset.from_iterable([
            DatasetItem(id='кириллица с пробелом',
                image=np.ones((8, 8, 3)), annotations=[
                    Bbox(0, 1, 3, 5),
                ]),
        ])

        for task in [
            IcdarTextLocalizationConverter,
            IcdarTextSegmentationConverter,
            IcdarWordRecognitionConverter,
        ]:
            with self.subTest(subformat=task), TestDir() as test_dir:
                self._test_save_and_load(expected_dataset,
                    partial(task.convert, save_images=True),
                    test_dir)
```

yasakova-anastasia: In the extractor `root = format_root/task`. Maybe also do it in the converter?

zhiltsov-max: This is what I see in https://github.com/openvinotoolkit/datumaro/pull/174/files#diff-8dabf26e19c18cd14fd53c6edd8a4b1d2dda9ebff6029ef2786e570df53ac0daR76-R77:
and https://github.com/openvinotoolkit/datumaro/pull/174/files#diff-8dabf26e19c18cd14fd53c6edd8a4b1d2dda9ebff6029ef2786e570df53ac0daR123-R124
```
            image_path = osp.join(self._path, IcdarPath.IMAGES_DIR,
                item_id + IcdarPath.IMAGE_EXT)
```

yasakova-anastasia: Yes, but the path is modified in `find_sources()`. In the extractor `self._path = format_dir/task` (in localization and segmentation tasks). Maybe in the converter to `__init__` with `self._save_dir` do the same.

It seems to me that the images folder should be in the task folder.

PR #178, opened 2021-03-22
Add ndr cli

seungyoon-woo:
### Summary
This pr includes :
* CLI support for NDR
* Handle large number of data

### How to test

`python -m unittest -v tests/test_ndr.py`

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/plugins/ndr.py
+++ b/datumaro/plugins/ndr.py
@@ -8,19 +8,52 @@ import numpy as np
 from scipy.linalg import orth
 
 from datumaro.components.extractor import Transform, DEFAULT_SUBSET_NAME
+from datumaro.components.cli_plugin import CliPlugin
 
-AlgoList = Enum("AlgoList", ["gradient"]) # other algorithms will be added
+Algorithm = Enum("Algorithm", ["gradient"]) # other algorithms will be added
 
-class NDR(Transform):
+OverSamplingMethod = Enum("OverSamplingMethod", ["random", "similarity"])
+
+UnderSamplingMethod = Enum("UnderSamplingMethod", ["uniform", "inverse"])
+
+class NDR(Transform, CliPlugin):
     """
     Near-duplicated image removal |n
     Removes near-duplicated images in subset |n
+    Example:  control number of outputs to 100 after NDR |n
+    |s|s%(prog)s \ |n
+    |s|s|s|s--working_subset train \ |n
+    |s|s|s|s--algorithm gradient \ |n
+    |s|s|s|s--num_cut 100 \ |n
+    |s|s|s|s--over_sample random \ |n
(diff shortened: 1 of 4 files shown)


zhiltsov-max (approved):

datumaro/plugins/ndr.py
```diff
@@ -71,7 +104,7 @@ def __init__(self, extractor,
         # parameter validation before main runs
         if working_subset == duplicated_subset:
             raise ValueError("working_subset == duplicated_subset")
-        if algorithm not in [algo_name.name for algo_name in AlgoList]:
+        if algorithm not in [algo.name for algo in Algorithm]:
```
zhiltsov-max, line 107: I suggest to use the approach we use in other places like this: accept None, string and Enum in c-tor, convert to Enum and use only it internally.

For example: https://github.com/openvinotoolkit/datumaro/blob/develop/datumaro/plugins/voc_format/converter.py#L96

zhiltsov-max (approved):
Please fix tests

seungyoon-woo:
Squashed all commit into one and delete not related to issue

datumaro/plugins/ndr.py
seungyoon-woo: Can you tell me specific detail of what c-tor is and use internally mean? 
assert algorithm is None or isinstance(algorithm, (str, Algorithm)) 
Is this code correct?
And do I apply change for next 109~112 line?

zhiltsov-max: C-tor = class constructor = `__init__` method.
Internally = in the class implementation. Use only Enum values everywhere, where user interaction is not supposed.

1. Looks right. Also I suggest to substitute a default value, if `None` is passed.
2. It would be nice to be consistent.

seungyoon-woo: But if I allowed string type, I cannot handle algorithm with string w/ wrong algorithm name.
Do you have any alternatives?

tests/test_ndr.py
```diff
@@ -94,28 +104,28 @@ def test_ndr_without_cut(self):
...
         # train : 300, val : 300, test : 300
         np.random.seed(1234)
-        source = self._generate_classification_dataset(config, 10)
+        source = self._generate_dataset(config, 10)
 
         result = ndr.NDR(source, working_subset='train', seed=12145)
 
-        self.assertEqual(2, len(result.get_subset("train")))
-        self.assertEqual(298, len(result.get_subset("duplicated")))
+        self.assertEqual(1, len(result.get_subset("train")))
```
zhiltsov-max, line 111: Why it was changed?

seungyoon-woo: After changing default block shape from (4,4) to (8,8), there was change in train and duplicated subset length

PR #179, opened 2021-03-22
Checking before adding attributes (Market-1501)

yasakova-anastasia:
### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/plugins/market1501_format.py
+++ b/datumaro/plugins/market1501_format.py
@@ -88,7 +88,8 @@ class Market1501Extractor(SourceExtractor):
                     image=image_path)
                 items[item_id] = item
 
-            if pid != Market1501Path.UNKNOWN_ID:
+            if pid != Market1501Path.UNKNOWN_ID or \
+                    camid != Market1501Path.UNKNOWN_ID:
                 attributes = item.attributes
                 attributes['query'] = subdir == Market1501Path.QUERY_DIR
                 attributes['person_id'] = pid
--- a/datumaro/plugins/market1501_format.py
+++ b/datumaro/plugins/market1501_format.py
@@ -87,11 +87,11 @@ class Market1501Extractor(SourceExtractor):
                 item = DatasetItem(id=item_id, subset=self._subset,
                     image=image_path)
                 items[item_id] = item
-
-            attributes = item.attributes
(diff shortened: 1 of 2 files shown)


datumaro/plugins/market1501_format.py
```diff
@@ -88,10 +88,11 @@ def _load_items(self, rootdir):
...
 
-            attributes = item.attributes
-            attributes['query'] = subdir == Market1501Path.QUERY_DIR
-            attributes['person_id'] = pid
-            attributes['camera_id'] = camid
+            if pid != Market1501Path.UNKNOWN_ID:
+                attributes = item.attributes
+                attributes['query'] = subdir == Market1501Path.QUERY_DIR
+                attributes['person_id'] = pid
+                attributes['camera_id'] = camid
```
zhiltsov-max, line 95: Why this information is not needed, when id is unknown?

yasakova-anastasia: I think that we shouldn't include this information. If we initially didn't have attributes in the dataset, we shouldn't add them.

zhiltsov-max: I think, it is impossible for us to know at this point, if these values were present or not. Also, how about the `test` subset?

The second question is why do we exclude all the attributes?

yasakova-anastasia: For example, we cannot check 'query' like that, so we check only one and it decides for all.

> I think, it is impossible for us to know at this point, if these values were present or not. Also, how about the `test` subset?
What do you mean with `test` subset?

zhiltsov-max: I mean the `bounding_box_test` directory from https://www.kaggle.com/pengcw1/market-1501/data, which contains images having -1, but camera ids and sequence ids are meaningful.

yasakova-anastasia: Then we think we have no attributes when `person_id` and `camera_id` are -1.

zhiltsov-max (approved):
I'm still not convinced that this approach is right, but I suggest to merge and look.

yasakova-anastasia:
> I'm still not convinced that this approach is right, but I suggest to merge and look.

Yes, I agree with you

PR #195, opened 2021-03-30
Format fixes in COCO and VOC

### LINKED ISSUE #183: create subset folders when converting voc to coco /w save-images option
When I convert the voc-format data to coco format with the `--save-images` option, I've expected that output images are under their subsets, but I couldn't find subsets.

In the [user manual - Import project](https://github.com/openvinotoolkit/datumaro/blob/develop/docs/user_manual.md#import-project), I can find COCO-like format as below.
```
COCO/
├── annotations/
│   ├── instances_val2017.json
│   ├── instances_train2017.json
├── images/
│   ├── val2017
│   ├── train2017
```

But after converting the dataset to coco format, I've got the format below.
```
COCO/
├── annotations/
│   ├── instances_val2017.json
│   ├── instances_train2017.json
├── images/
```
All the images are under the `images` folder, not under the `images/val2017` or `images/train2017` folder.
I think it should be the same as the first example.

In addition, the 2nd format is also described in the [user manual - Add and remove data](https://github.com/openvinotoolkit/datumaro/blob/develop/docs/user_manual.md#add-and-remove-data).
I think we need to update the user manual, too.

zhiltsov-max:
### Summary

Closes #183 

- Added an option to save images in COCO in flat layout, disable by default
- Removed conflicting mask attributes in VOC extractor

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/plugins/coco_format/converter.py
+++ b/datumaro/plugins/coco_format/converter.py
@@ -478,9 +478,12 @@ class CocoConverter(Converter):
         parser.add_argument('--allow-attributes',
             type=str_to_bool, default=True,
             help="Allow export of attributes (default: %(default)s)")
-        parser.add_argument('--reindex', action='store_true',
-            help="Assign new indices to images and annotations "
-                "(default: %(default)s)")
+        parser.add_argument('--reindex', type=str_to_bool, default=False,
+            help="Assign new indices to images and annotations, "
+                "useful to avoid merge conflicts (default: %(default)s)")
+        parser.add_argument('--merge-images', type=str_to_bool, default=False,
+            help="Save all images into a single "
+                "directory (default: %(default)s)")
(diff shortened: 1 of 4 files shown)


datumaro/plugins/coco_format/converter.py
```diff
@@ -605,6 +607,12 @@ def patch(cls, dataset, patch, save_dir, **kwargs):
             if osp.isfile(image_path):
                 os.unlink(image_path)
 
+            image_path = osp.join(images_dir, subset,
+                conv._make_image_filename(item))
+            if osp.isfile(image_path):
+                os.unlink(image_path)
```
yasakova-anastasia, line 613: The same code is above

zhiltsov-max: It mostly is, but the second block has `subset` part in the path. We need to check and remove both of the images.

Issue #194, opened 2021-03-30
Warning when trying to export yolo dataset after split


anwar1997ff:
Steps to reproduce:
`dataset = Dataset.import_from("/path/to/dataset", 'yolo')`
`dataset.transform('detection_split', [('train', 0.5),('val', 0.5)])`
`dataset.export('/output/path', 'yolo')`

Warning:
`WARNING:root:Skipping subset export 'val'. If specified, the only valid names are 'train', 'valid'`

Which wont save the val data in the dataset.

Replacing "valid" with "val" in datumaro/datumaro/plugins/yolo_format/format.py solved this issue, but not sure if it raises other conflicts

zhiltsov-max:
Hi. If a subset name is different from `train` and `valid` (https://github.com/AlexeyAB/darknet/blob/master/src/detector.c#L30), the training framework will fail to use it. But you're right, splitters need to support different subset names.

jihyeonyi:
Currently, we're restricting the subset name in the detection_split, but it's possible to support different subset names. 
This is the same for the classification_split.
So, I'll change them for both splitting algorithms.

PR #207, opened 2021-04-07
Kate/handling multiple attributes and speed up detection split

jihyeonyi:
### Summary
While handling multi-attributes, we've found that task-specific splitters are mishandling them.
This fix includes,
1. better handling of the multi-attributes case 
2. speed up & OOM issue is fixed for the detection_split

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/plugins/splitter.py
+++ b/datumaro/plugins/splitter.py
@@ -9,6 +9,7 @@ from math import gcd
 from datumaro.components.extractor import (Transform, AnnotationType,
     DEFAULT_SUBSET_NAME)
 from datumaro.components.cli_plugin import CliPlugin
+from datumaro.util import cast
 
 NEAR_ZERO = 1e-7
 
@@ -154,14 +155,15 @@ class _TaskSpecificSplit(Transform, CliPlugin):
         """
 
         # float--> numerical, others(int, string, bool) --> categorical
-        def _is_number(value):
+        def _is_float(value):
             if isinstance(value, str):
-                try:
-                    float(value)
-                    return True
-                except ValueError:
-                    return False
+                casted = cast(value, float)
+                if casted is not None:
+                    if cast(casted, str) == value:
+                        return True
+                return False
             elif isinstance(value, float):
+                cast(value, float)
                 return True
             return False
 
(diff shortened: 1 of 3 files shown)


jihyeonyi:
I've added the code to remove the subset name restriction for classification and detection.

datumaro/plugins/splitter.py
```diff
@@ -70,21 +74,29 @@ def _get_uniq_annotations(dataset):
...
-        if valid is None:
-            valid = ["train", "val", "test"]
+        subsets = set()
+        valid = ["train", "val", "test"]
+        # remove subset name restriction
+        # regarding https://github.com/openvinotoolkit/datumaro/issues/194
         for subset, ratio in splits:
-            assert subset in valid, \
-                "Subset name must be one of %s, but got %s" % (valid, subset)
+            if restrict:
```
zhiltsov-max, line 85: What is the point of this restriction? Is it possible to require more than 3 subsets? Do subsets have special processing depending on the name?

datumaro/plugins/splitter.py
```diff
@@ -129,37 +152,74 @@ def _group_by_attr(items):
...
+
+        # float--> numerical, others(int, string, bool) --> categorical
+        def _is_number(value):
+            if isinstance(value, str):
+                try:
+                    float(value)
+                    return True
+                except ValueError:
+                    return False
+            elif isinstance(value, float):
```
zhiltsov-max, line 164:
```suggestion
            elif isinstance(value, (int, float)):
```

zhiltsov-max: Check if you can get any use for `datumaro.util.cast`.

datumaro/plugins/splitter.py
jihyeonyi: if restrict is Ture, the subset names are restricted into ["train", "val", "test"]. This is for re-identification at the moment.
For detection and classification, this restriction isn't needed so I changed this code. This is regarding issue #194

datumaro/plugins/splitter.py
jihyeonyi: 'int' is regarded as categorical at the moment. 
That's why I checked 'float' only. 
For the `datumaro.util.cast`, I'll check it.

zhiltsov-max (dismissed):

PR #210, opened 2021-04-08
Add panoptic and stuff COCO format

zmaslova:
### Checklist

### License

```python
#
```


datumaro/plugins/coco_format/extractor.py
```diff
@@ -100,6 +109,34 @@ def _load_person_kp_categories(self, loader):
 
         return categories
 
+    @staticmethod
+    def _load_panoptic_config(path):
+        with open(path, 'r') as f:
+            import json
+            return json.load(f)
+
+    def _load_panoptic_categories(self, config, root_path):
```
zhiltsov-max, line 118: From https://cocodataset.org/#format-data I understood that the panoptic task has the same json structure as others and includes color info. Why the current solution uses a separate file?

datumaro/plugins/coco_format/extractor.py
```diff
@@ -124,6 +161,55 @@ def _load_items(self, loader):
...
         return items
 
+    def _load_panoptic_items(self, config, panoptic_images):
+        import numpy as np
+        import PIL.Image
+        from collections import defaultdict
+        items = OrderedDict()
+
+        def rgb2id(color):
+            if isinstance(color, np.ndarray) and len(color.shape) == 3:
```
zhiltsov-max, line 171: Consider using `datumaro.util.mask_tools.unpaint_mask` instead

datumaro/plugins/coco_format/extractor.py
```diff
@@ -124,6 +161,55 @@ def _load_items(self, loader):
...
+
+        for ann in config['annotations']:
+            img_id = ann['image_id']
+            image_path = osp.join(self._images_dir, imgs_info[img_id]['file_name'])
+            image_size = (imgs_info[img_id].get('height'), imgs_info[img_id].get('width'))
+            image = Image(path=image_path, size=image_size)
+            anns = list()
+
+            panoptic_image_path = osp.join(panoptic_images, ann['file_name'])
+            panoptic_format = np.array(PIL.Image.open(panoptic_image_path), dtype=np.uint32)
```
zhiltsov-max, line 189: Consider using `datumaro.util.mask_tools.load_mask` and `.lazy_mask` (which loads through image cache)

datumaro/plugins/coco_format/extractor.py
```diff
@@ -124,6 +161,55 @@ def _load_items(self, loader):
...
+            for segm_info in ann['segments_info']:
+                cat_id = segm_info['category_id']
+                segm_id = segm_info['id']
+                mask = pan == segm_id
+                rle = mask_utils.encode(np.asfortranarray(mask.astype('uint8')))
+                rle['counts'] = rle['counts'].decode('utf8')
+                rle_per_category[cat_id].append(rle)
+
+            for cat_id, rle_list in rle_per_category.items():
+                rle = mask_utils.merge(rle_list)
```
zhiltsov-max, line 201: In Datumaro we provide a mask per instance, they need not to be merged. Mask `id` would be instance id. A bbox, if provided, should be an annotation of type `Bbox` with `group` = `id` of the mask (but having a separate bbox and area attributes seem useless, unless they have special meaning, because there are corresponding methods in the Mask class).

datumaro/plugins/coco_format/extractor.py
```diff
@@ -124,6 +161,55 @@ def _load_items(self, loader):
...
+            anns = list()
+
+            panoptic_image_path = osp.join(panoptic_images, ann['file_name'])
+            panoptic_format = np.array(PIL.Image.open(panoptic_image_path), dtype=np.uint32)
+            pan = rgb2id(panoptic_format)
+            rle_per_category = defaultdict(list)
+            for segm_info in ann['segments_info']:
+                cat_id = segm_info['category_id']
+                segm_id = segm_info['id']
+                mask = pan == segm_id
```
zhiltsov-max, line 195: Consider using the Mask class with a single shared mask image in memory and `lambda m, c: m == c` ([example](https://github.com/openvinotoolkit/datumaro/blob/develop/datumaro/plugins/voc_format/extractor.py#L310)) as a lazy_mask extractor to avoid memory explosion on mask splitting for large number of instances and pixels.

tests/test_coco_format.py
```diff
@@ -146,6 +146,35 @@ def test_can_detect(self):
...
+                0: (0, 0, 0),
+                1: (128, 0, 0),
+            }),
+        })
+
+        dataset = Dataset.import_from(
+            osp.join(DUMMY_DATASET_DIR, 'coco_panoptic'), 'coco')
+
+        #compare_datasets(self, expected_dataset, dataset, require_images=True)
+        compare_categories(self, expected_dataset.categories(), dataset.categories())
```
zhiltsov-max, line 176: Comparison of categories is a part of dataset comparison function. Why datasets are not compared? If there are problems because of mask encoding in pycocotools, make mask size a multiplier of 5. It does not really affect real use cases, but helps with tests.

zhiltsov-max (changes requested):
Thanks for your contribution. How about the `stuff` task?

datumaro/plugins/coco_format/extractor.py
zmaslova: Because it's not actually the same, but yes. quite similar. If you will try to open it with pycocotools, you'll get the same parse error https://github.com/openvinotoolkit/cvat/issues/1143
I couldn't find panoptic json files with colors inside, just additional https://github.com/cocodataset/panopticapi/blob/master/panoptic_coco_categories.json

datumaro/plugins/coco_format/extractor.py
zmaslova: I wanted to make real segmentation masks (as it's here, merged for different instances of one class) and panoptic masks (will be done later, it's important for export task)

zmaslova:
> Thanks for your contribution. How about the `stuff` task?

Haven't try it, but will, ok

datumaro/plugins/coco_format/extractor.py
zhiltsov-max: Indeed, there is no color information even in the official annotations. But, actually, it is not used for mask parsing, because `segment_info.id` is enough here, so, maybe, we could avoid requiring a special file. I just want this format to be able to read the unmodified official dataset as a minimum.

datumaro/plugins/coco_format/extractor.py
zhiltsov-max: What is the use for merged masks of one class? Fully merged masks look attractive, but again, they are only useful for training, and they are interchangeable (by merging) with instance masks. In Datumaro, we decided to use instance masks, unless there is a strong need for compiled masks. And compiled masks can be obtained with `datumaro.components.extractor.CompiledMask`.

datumaro/plugins/coco_format/extractor.py
zmaslova: I failed with using all that pretty lazy mask functions - they did not what I want

datumaro/plugins/coco_format/converter.py
```diff
@@ -637,3 +705,14 @@ class CocoLabelsConverter(CocoConverter):
...
+class CocoPanopticConverter(CocoConverter):
+    def __init__(self, *args, **kwargs):
+        kwargs['tasks'] = CocoTask.panoptic
+        super().__init__(*args, **kwargs)
+
+class CocoStuffConverter(CocoConverter):
+    def __init__(self, *args, **kwargs):
+        kwargs['tasks'] = CocoTask.stuff
+        super().__init__(*args, **kwargs)
+        self._segmentation_mode = SegmentationMode.mask
```
zhiltsov-max, line 718: I think, it would be better to set a kwarg a line above.

datumaro/plugins/coco_format/converter.py
```diff
@@ -451,6 +451,70 @@ def save_annotations(self, item):
...
+            segment_info['area'] = float(ann.get_area())
+            segment_info['bbox'] = list(ann.get_bbox())
+            segment_info['iscrowd'] = int(ann.attributes.get('is_crowd', 0))
+            segments_info.append(segment_info)
+            mask = mask_utils.decode(ann.rle) == 1
+            pan_format[mask] = ann.id
+
+        if segmentation_path:
+            os.makedirs(segmentation_path, exist_ok=True)
+            Image.fromarray(id2rgb(pan_format)).save(
```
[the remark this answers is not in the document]
zmaslova: about 3.5 minutes  in average now

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support for escaping in attribiute values in LabelMe format (<https://github.com/openvinotoolkit/datumaro/issues/49>)
 - Support for Segmentation Splitting (<https://github.com/openvinotoolkit/datumaro/pull/223>)
 - Support for CIFAR-10/100 dataset format (<https://github.com/openvinotoolkit/datumaro/pull/225>)
+- Support import COCO panoptic and stuff format (<https://github.com/openvinotoolkit/datumaro/pull/210>, <https://github.com/openvinotoolkit/cvat/issues/1784>, <https://github.com/openvinotoolkit/cvat/issues/1143>)
```
zhiltsov-max, line 14: Adding support in Datumaro wont automatically add it in CVAT. It will need at least updating a Datumaro revision in CVAT, which is fixed. It will allow importing, but I'd expect a new format in CVAT for coco segmentation.

zmaslova: you're right. fixed

datumaro/plugins/coco_format/converter.py
```diff
@@ -451,6 +451,70 @@ def save_annotations(self, item):
...
+                continue
+
+            segment_info = {}
+            segment_info['id'] = ann.id
+            segment_info['category_id'] = ann.label
+            segment_info['area'] = float(ann.get_area())
+            segment_info['bbox'] = list(ann.get_bbox())
+            segment_info['iscrowd'] = int(ann.attributes.get('is_crowd', 0))
+            segments_info.append(segment_info)
+            mask = mask_utils.decode(ann.rle) == 1
```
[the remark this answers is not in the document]
zmaslova: fixed

datumaro/plugins/coco_format/converter.py
```diff
@@ -576,10 +640,12 @@ def apply(self):
                             '' if self._merge_images else subset_name))
                     else:
                         log.debug("Item '%s' has no image info", item.id)
+                segmentation_path = osp.join(self._ann_dir,
+                    'panoptic' + ('' if self._merge_images else "_"+ subset_name))
                 for task_conv in task_converters.values():
                     task_conv.save_image_info(item,
                         self._make_image_filename(item))
```
zhiltsov-max, line 647: Probably we also need to fix handling of image paths wrt `self._merge_images`.

datumaro/plugins/coco_format/converter.py
```diff
@@ -576,10 +640,12 @@ def apply(self):
                             '' if self._merge_images else subset_name))
                     else:
                         log.debug("Item '%s' has no image info", item.id)
+                segmentation_path = osp.join(self._ann_dir,
```
zhiltsov-max, line 643: I don't think `merge_images` should be applied to annotations.

I think segmentation path should be made a method in this class to avoid an extra parameter in all the subclasses.

datumaro/plugins/coco_format/extractor.py
```diff
@@ -42,9 +42,17 @@ def __init__(self, path, task, merge_instance_polygons=False, subset=None):
...
-        loader = self._make_subset_loader(path)
-        self._load_categories(loader)
-        self._items = list(self._load_items(loader).values())
+        if self._task == CocoTask.panoptic:
+            #panoptic is not added to pycocotools
+            panoptic_config = self._load_panoptic_config(path)
+            panoptic_images = osp.splitext(path)[0]
+
+            self._load_panoptic_categories(panoptic_config, rootpath)
+            self._items = list(self._load_panoptic_items(panoptic_config, panoptic_images).values())
```
zhiltsov-max, line 51: Please limit lines to 80 characters.

datumaro/plugins/coco_format/extractor.py
```diff
@@ -100,6 +107,22 @@ def _load_person_kp_categories(self, loader):
 
         return categories
 
+    @staticmethod
+    def _load_panoptic_config(path):
+        with open(path, 'r') as f:
+            import json
```
zhiltsov-max, line 113: This import can be moved to the top of the file

datumaro/plugins/coco_format/extractor.py
```diff
@@ -124,6 +147,42 @@ def _load_items(self, loader):
 
         return items
 
+    def _load_panoptic_items(self, config, panoptic_images):
+        import numpy as np
```
zhiltsov-max, line 151: This import can be moved to the top of the file

datumaro/plugins/coco_format/extractor.py
```diff
@@ -100,6 +107,22 @@ def _load_person_kp_categories(self, loader):
 
         return categories
 
+    @staticmethod
+    def _load_panoptic_config(path):
+        with open(path, 'r') as f:
+            import json
+            return json.load(f)
+
+    def _load_panoptic_categories(self, config, root_path):
```
zhiltsov-max, line 116: `root_path` is unused

datumaro/plugins/coco_format/extractor.py
```diff
@@ -124,6 +147,42 @@ def _load_items(self, loader):
...
+            anns = list()
+
+            panoptic_image_path = osp.join(panoptic_images, ann['file_name'])
+            panoptic_format = np.array(load_image(panoptic_image_path, dtype=np.uint32))
+            pan = bgr2id(panoptic_format)
+            for segm_info in ann['segments_info']:
+                cat_id = segm_info['category_id']
+                segm_id = segm_info['id']
+                mask = pan == segm_id
+                rle = mask_utils.encode(np.asfortranarray(mask.astype('uint8')))
```
zhiltsov-max, line 175: Please, update the implementation of `datumaro.util.mask_tools.mask_to_rle` if this code works faster. This can be done later.

datumaro/plugins/coco_format/extractor.py
```diff
@@ -124,6 +147,42 @@ def _load_items(self, loader):
...
+
+        for ann in config['annotations']:
+            img_id = int(ann['image_id'])
+            image_path = osp.join(self._images_dir, imgs_info[img_id]['file_name'])
+            image_size = (imgs_info[img_id].get('height'), imgs_info[img_id].get('width'))
+            image = Image(path=image_path, size=image_size)
+            anns = list()
+
+            panoptic_image_path = osp.join(panoptic_images, ann['file_name'])
+            panoptic_format = np.array(load_image(panoptic_image_path, dtype=np.uint32))
```
zhiltsov-max, line 169: load_image is expected to return a numpy array

zmaslova: fixed

datumaro/plugins/coco_format/extractor.py
zmaslova: Yes, will be done in separate PR

datumaro/plugins/coco_format/extractor.py
zmaslova: Fixed

datumaro/plugins/coco_format/extractor.py
zmaslova: Fixed

datumaro/plugins/coco_format/extractor.py
zmaslova: Fixed

datumaro/plugins/coco_format/extractor.py
zmaslova: Fixed

datumaro/plugins/coco_format/converter.py
zmaslova: Redone, is it ok now?

datumaro/plugins/coco_format/converter.py
zmaslova: Don't get it

datumaro/plugins/coco_format/converter.py
```diff
@@ -451,6 +453,67 @@ def save_annotations(self, item):
...
+            return bgr_map
+
+        segments_info = list()
+        masks = []
+        for ann in item.annotations:
+            if ann.type != AnnotationType.mask:
+                continue
+
+            segment_info = {}
+            segment_info['id'] = ann.id
```
zhiltsov-max, line 494: It seems that we need a fallback to generate these values, when they are missing (`None`).

datumaro/plugins/coco_format/converter.py
```diff
@@ -451,6 +453,67 @@ def save_annotations(self, item):
...
+
+        if masks:
+            pan_format = mask_tools.merge_masks(
+                ((m.image, m.id) for m in masks),
+                start=np.zeros(item.image.size, dtype=np.uint32))
+            save_image(osp.join(self._context._segmentation_dir, ann_filename),
+                id2bgr(pan_format), create_dir=True)
+
+        elem = {
+            'id': item.id,
```
zhiltsov-max, line 510: `id` does not exist in the official spec, what is the point in it?

zmaslova: sorry, forgot to remove it. removed now

datumaro/plugins/coco_format/converter.py
zmaslova: done

zhiltsov-max (dismissed):

zhiltsov-max (approved):

PR #225, opened 2021-04-22
Support for CIFAR-10/100 format

yasakova-anastasia:
### Summary

- Added support for CIFAR-10/100 format (classification task)

### Checklist

### License

```python
#
```


datumaro/plugins/cifar_format.py
```diff
@@ -0,0 +1,179 @@
...
+    BATCHES_META = 'batches.meta'
+    TRAIN_ANNOTATION_FILE = 'data_batch_'
+    IMAGES_DIR = 'images'
+    IMAGE_SIZE = 32
+
+CifarLabel = ['airplane', 'automobile', 'bird', 'cat',
+    'deer', 'dog', 'frog', 'horse', 'ship', 'truck']
+
+
+class CifarExtractor(SourceExtractor):
```
zhiltsov-max, line 25: As I understand, it works with python version of the format. I think we need to document it somewhere.

datumaro/plugins/cifar_format.py
```diff
@@ -0,0 +1,179 @@
...
+        for subset_name, subset in self._extractor.subsets().items():
+            filenames = []
+            labels = []
+            data = []
+            image_sizes = {}
+            for item in subset:
+                filenames.append(item.id + self._find_image_ext(item))
+
+                if item.has_image and self._save_images:
+                    self._save_image(item, osp.join(self._save_dir, CifarPath.IMAGES_DIR,
```
zhiltsov-max, line 135: Do we need an option to pack images into a single file?

datumaro/plugins/cifar_format.py
```diff
@@ -0,0 +1,179 @@
...
+                        self._make_image_filename(item)))
+
+                anns = [a.label for a in item.annotations
+                    if a.type == AnnotationType.label]
+                label = None
+                if anns:
+                    label = anns[0]
+                labels.append(label)
+
+                image = item.image.data
```
zhiltsov-max, line 145: An image can be missing, or it can have no data (will be None).

datumaro/plugins/cifar_format.py
```diff
@@ -0,0 +1,179 @@
...
+                if not image.any():
+                    data.append(image)
+                else:
+                    data.append(image.reshape(image.shape[0] * image.shape[1] * \
+                        image.shape[2]).astype(np.uint8))
+                    if image.shape[0] != CifarPath.IMAGE_SIZE or \
+                            image.shape[1] != CifarPath.IMAGE_SIZE:
+                        image_sizes[len(data) - 1] = (image.shape[0], image.shape[1])
```
zhiltsov-max, line 153: Consider using `image.size`, which can be filled even if there is no image data available.

datumaro/plugins/cifar_format.py
```diff
@@ -0,0 +1,179 @@
...
+                label = None
+                if anns:
+                    label = anns[0]
+                labels.append(label)
+
+                image = item.image.data
+                if not image.any():
+                    data.append(image)
+                else:
+                    data.append(image.reshape(image.shape[0] * image.shape[1] * \
```
zhiltsov-max, line 149: Do we need an image here if it is already saved above?

datumaro/plugins/cifar_format.py
```diff
@@ -0,0 +1,179 @@
...
+                # needed for different image sizes
+                annotation_dict['image_sizes'] = [image_sizes.get(p, size)
+                    for p in range(len(data))]
+
+            filename = '%s_batch' % subset_name
+            batch_label = None
+            if subset_name.startswith('train_'):
+                num = subset_name.split('_')[1]
+                filename = CifarPath.TRAIN_ANNOTATION_FILE + num
+                batch_label = 'training batch %s of 5' % num
```
zhiltsov-max, line 171: Is there a fixed total number of batches?

zhiltsov-max (commented):
Nice PR. There are some questions about image handling in converter.

datumaro/plugins/cifar_format.py
yasakova-anastasia: Yes, dataset is divided into five training batches and one test batch.

datumaro/plugins/cifar_format.py
yasakova-anastasia: I think that the ability to convert images to single files can be useful here.

datumaro/plugins/cifar_format.py
yasakova-anastasia: It seems to me that we should always write the image data to this file, and convert the data to images only if `save_images == True`

datumaro/plugins/cifar_format.py
yasakova-anastasia: I think that if the image doesn't have data, we don't need to store the size.

datumaro/plugins/cifar_format.py
zhiltsov-max: What is the point in image duplication?

yasakova-anastasia: The AC has [the ability to convert images](https://github.com/openvinotoolkit/open_model_zoo/blob/9ca5dbeff80464bf5728e8be25daedfe9a9208d7/tools/accuracy_checker/accuracy_checker/annotation_converters/cifar.py#L161-L164). 
But you're right, in our case it's  image duplication.

datumaro/plugins/cifar_format.py
zhiltsov-max: It is not guaranteed, that symbols after `_` will be a number. I suggest adding a check with `datumato.utils.cast` at least.

zhiltsov-max (approved):

Issue #226, opened 2021-04-23
error in 'find_mask_bbox' @ util.mask_tools


jihyeonyi:
I've tried to `get_bbox` of Mask with an image below.

```
Mask(image = [ 
    [ 0, 0, 0, 0, 0 ],
    [ 0, 0, 1, 1, 1 ],
    [ 0, 0, 1, 1, 1 ],
    [ 0, 0, 1, 1, 1 ],
    [ 0, 0, 1, 1, 1 ] 
])
```

I expected (2, 1, 3, 4),  but I got (2, 1, 2, 3)

I think the `find_mask_bbox` function in the util.mask_tools should be corrected in calculating the width and height.

PR #227, opened 2021-04-28
add validation item for instance segmentation

jihyeonyi:
### Summary
Enhance validator to support instance segmentation

### How to test
python -m unittest -v tests/test_validator.py

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/components/validator.py
+++ b/datumaro/components/validator.py
@@ -963,9 +963,13 @@ class SegmentationValidator(_Validator):
             mask_has_error = False
 
             _x, _y, _w, _h = ann.get_bbox()
-            if ann.type == AnnotationType.mask:  # temp.
+
+            # Detete the following block when #226 is resolved
+            # https://github.com/openvinotoolkit/datumaro/issues/226
+            if ann.type == AnnotationType.mask:
                 _w += 1
                 _h += 1
+
             area = ann.get_area()
 
             ann_mask_info = _generate_ann_mask_info(area, _w, _h)
@@ -1039,7 +1043,10 @@ class SegmentationValidator(_Validator):
             mask_label_stats = dist_by_label[label_name]
 
             _x, _y, _w, _h = ann.get_bbox()
-            if ann.type == AnnotationType.mask:  # temp.
+
+            # Detete the following block when #226 is resolved
+            # https://github.com/openvinotoolkit/datumaro/issues/226
+            if ann.type == AnnotationType.mask:
                 _w += 1
                 _h += 1
             area = ann.get_area()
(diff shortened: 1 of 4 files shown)


datumaro/components/validator.py
```diff
@@ -876,14 +912,244 @@ def generate_reports(self, stats):
...
+                'area': area,
+                'width': _w,
+                'height': _h,
+            }
+
+        def _update_mask_stats_by_label(item_key, ann, mask_label_stats):
+            mask_has_error = False
+
+            _x, _y, _w, _h = ann.get_bbox()
+            if ann.type == AnnotationType.mask:  # temp.
```
zhiltsov-max, line 966: Can you make the comment more clear?

jihyeonyi: Once #226 is resolved, this code can be deleted.
I'll leave this as a comment.

zhiltsov-max: Ok. I don't think this fix completely fixes the problem, though, because when we have `[0, 0, 0]`, `[0, 1, 0]` the fix will always return 1.

zhiltsov-max (dismissed):

datumaro/components/validator.py
jihyeonyi: In that case, the bounding box should be 1x1, but currently, it returns 0x0.
```
>>> mask = np.array([[0,0,0], [0,1,0]])
>>> find_mask_bbox(mask)
[1, 1, 0, 0]
```
So, I think this temporal code can fix the problem.

zhiltsov-max: No, I meant that two cases: `[0, 0, 0]` and `[0, 1, 0]` - both will return 1?

zhiltsov-max:
I tried to launch this validator for VOC2012 dataset and got an error:

The output file has:
```
         "items_far_from_mean": {
                        "('2011_000066', 'trainval')": {
                            "0": 
```

And the error is:
```
  File "datumaro/datumaro/cli/contexts/project/__init__.py", line 844, in validate_command
    json.dump(validation_results, f, indent=4, sort_keys=True)
TypeError: Object of type 'int64' is not JSON serializable
```

So probably there is need to convert some values.


Also, I think this kind of output can be improved: `"('2011_003271', 'val')"`

jihyeonyi:
I can't reproduce VOC2012 case. 
However, I guess this is due to the numpy datatypes. So I added numpy encoder.
Could you please check the error is gone?

And for the tuple key like `"('2011_003271', 'val')"`, I agree with you. 
Let's leave this as a future improvement item.

zhiltsov-max:
Yes, the fix helped.

zhiltsov-max (approved):

PR #234, opened 2021-05-06
Support for MNIST dataset format

yasakova-anastasia:
### Summary

- Add support for MNIST and MNIST in CSV dataset formats
- Add documentation for MNIST and MNIST in CSV dataset formats

### Checklist

### License

```python
#
```


zhiltsov-max (changes requested):

zhiltsov-max (approved):

docs/mnist_user_manual.md
```diff
@@ -0,0 +1,171 @@
...
+datum export -f imagenet -o <path/to/output/dir>
+# or
+datum convert -if mnist -i <path/to/mnist> -f imagenet -o <path/to/output/dir>
+```
+
+These commands also work for MNIST in CSV if you use `mnist_csv` instead of `mnist`.
+
+##  Export to MNIST
+
+There are few ways to convert dataset to MNIST format:
```
zhiltsov-max, line 118: Add descriptions for export parameters.

zhiltsov-max (dismissed):

datumaro/plugins/mnist_csv_format.py
```diff
@@ -0,0 +1,169 @@
...
+                        image = image.data.reshape(-1).astype(np.uint8).tolist()
+                        image.insert(0, label)
+                        data.append(image)
+                else:
+                    data.append([label])
+
+                if item.id != str(len(data) - 1):
+                    item_ids[len(data) - 1] = item.id
+
+            anno_file = osp.join(self._save_dir, 'mnist_%s.csv' % subset_name)
```
kirill-sizov, line 128: I think we need to create an save directory if it doesn't exist. What do you think?

kirill-sizov (changes requested):

datumaro/plugins/mnist_csv_format.py
zhiltsov-max: It is guaranteed to exist at this point in all practical use cases, but if this class is used directly, this can be a problem.

kirill-sizov: But when I try to export mnist dataset to unexisting directory with command  `datum export -f mnist_csv -o ./newdir`, I get an error `No such file or directory: '<project_path>/newdir/mnist_train.csv'`. And for other formats, this command works. Or am I not exporting correctly?

yasakova-anastasia: Yes, you're right, this error appears. It will appear in all formats that have no subfolders. This needs to be fixed, but I'm not sure what needs to be created in each format separately. Probably this can be done in [converter](https://github.com/openvinotoolkit/datumaro/blob/962ade03232aefba40c7adef92635cb6557adc59/datumaro/components/converter.py#L41) for all formats.

kirill-sizov (approved):

zhiltsov-max:
I tried to convert PASCAL VOC to MNIST CSV, and obtained 6 files with annotations - for `train`, `trainval` and `val` subsets. `mnist_train.csv` includes lots of `None` values, which, I assume, related to images with no labels (as well as should be in the test subset). Is it OK for the format, or these values need to be something else, like -1, 0, "" etc?

yasakova-anastasia:
@zhiltsov-max , I didn't find any information in the specification how the unlabelled item is described. `None` can cause problems?

zhiltsov-max:
I'd better have all the values of the same type, with `-1` for this case, instead.

PR #252, opened 2021-05-26
Fix spelling errors

IRDonch:
### Summary
Fix various misspellings/typos that I found with codespell.

### How to test
N/A

### Checklist

### License

```python
#
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
-- Support for escaping in attribiute values in LabelMe format (<https://github.com/openvinotoolkit/datumaro/issues/49>)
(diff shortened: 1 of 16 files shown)


zhiltsov-max (approved):
Is it possible to do automatically on CI?

IRDonch:
> Is it possible to do automatically on CI?

I don't think you can do it in a reliable way, because whether a word is actually a misspelling or not can be heavily context-dependent. For example, codespell suggests to replace "ans" with "and" (which I did in this PR), but in another context "ans" could be an abbreviation for "answer" and thus not an error. So I don't think you can use codespell without human review of the results.

IRDonch:
Well, in all fairness... if you're willing to tolerate some false negatives, you could keep an ignore list in the repository, and just add all ambiguous words there. I suppose that would be better than nothing, although it means that contributors have to not blindly trust the spellchecker and add words to the ignore list sometimes. :-)

zhiltsov-max:
I think, we can leave it in a list of recommended extensions for the environment setup in the developer guide.

PR #291, opened 2021-06-10
Add initial support for the Open Images dataset

### LINKED ISSUE #274: Support the Open Images dataset
[Open Images](https://storage.googleapis.com/openimages/web/index.html) is a dataset with ~9 million images and annotations of various types:

* image-level labels;
* object bounding boxes;
* instance segmentation masks;
* object relationships;
* localized narratives (synchronized audio recordings, text transcriptions and mouse traces describing the image).

The first three types are directly supported by Datumaro. The last one is not fully supported, but could be used to create caption-type annotations.

I think it would be reasonable to limit the scope of this issue to annotation types that are already supported. The ones that aren't can be added in the future.

IRDonch:
### Summary
Partly resolves #274. Currently, only reading supported, and only of images and image-level labels.

### How to test
Beyond unit tests, you can download the dataset and run the example listed in the documentation. I would advise only downloading one subset (and not the train subset), otherwise reading/writing will take a very long time.

### Checklist

### License

```python
#
```


datumaro/plugins/open_images_format.py
```diff
@@ -0,0 +1,155 @@
...
+
+        def load_from(path):
+            with self._open_csv_annotation(path) as image_reader:
+                for image_description in image_reader:
+                    image_id = image_description['ImageID']
+                    assert _RE_IMAGE_ID.fullmatch(image_id)
+                    assert image_id not in items_by_id
+
+                    subset = image_description['Subset']
+                    assert subset in {'test', 'train', 'validation'}
```
zhiltsov-max, line 101: We don't limit subset names to these 3, unless format can't represent others (like YOLO, where we have only `train` and `valid`).

datumaro/plugins/open_images_format.py
```diff
@@ -0,0 +1,155 @@
...
+import re
+
+from datumaro.components.extractor import (
+    AnnotationType, DatasetItem, Importer, Label, LabelCategories, Extractor,
+)
+
+# image IDs in Open Images are always hexadecimal numbers
+_RE_IMAGE_ID = re.compile(r'[0-9a-fA-F]+')
+
+class _OpenImagesPath:
```
zhiltsov-max, line 19: These classes are not supposed to be private.

datumaro/plugins/open_images_format.py
```diff
@@ -0,0 +1,155 @@
...
+        self._items.extend(items_by_id.values())
+
+        self._load_labels(items_by_id)
+
+    def _load_labels(self, items_by_id):
+        label_categories = self._categories[AnnotationType.label]
+
+        # TODO: implement reading of machine-annotated labels
+
+        for label_path in self._glob_annotations('*-human-imagelabels.csv'):
```
zhiltsov-max, line 130: Consider `os.listdir` and `glob.iglob`.

datumaro/plugins/open_images_format.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+class OpenImagesImporter(Importer):
+    @classmethod
+    def find_sources(cls, path):
+        for pattern in [
+            _OpenImagesPath.FULL_IMAGE_DESCRIPTION_NAME,
+            *_OpenImagesPath.SUBSET_IMAGE_DESCRIPTION_PATTERNS,
+        ]:
+            if glob.glob(osp.join(glob.escape(path), _OpenImagesPath.ANNOTATIONS_DIR, pattern)):
```
zhiltsov-max, line 152: I don't think calling `glob` in a loop is a good idea. Consider using `cls._find_sources_recursive()` with file name filter.

datumaro/plugins/open_images_format.py
```diff
@@ -0,0 +1,155 @@
...
+                return [{'url': path, 'format': 'open_images'}]
```
zhiltsov-max, line 153: By default, try to implement `Extractor` to work with a single subset annotation file. Inherit base `Extractor` if the default way gives significant performance problems or makes other implementation problems (eg. label mismatching / order, reusing some parsed data etc.).

datumaro/plugins/open_images_format.py
```diff
@@ -0,0 +1,155 @@
...
+
+        for label_path in self._glob_annotations('*-human-imagelabels.csv'):
+            with self._open_csv_annotation(label_path) as label_reader:
+                for label_description in label_reader:
+                    image_id = label_description['ImageID']
+                    item = items_by_id[image_id]
+
+                    confidence = float(label_description['Confidence'])
+
+                    if confidence > 0.5:
```
zhiltsov-max, line 138: By default, we use `<` and `<=`.

zhiltsov-max (changes requested):
Please, consult CONTRIBUTING.md and existing code about the code style.

Don't expect this code will only work with the public version of the dataset. We must support the public version, but if we can introduce reasonable extensions or relaxations to the format, it's better to do this.

datumaro/plugins/open_images_format.py
IRDonch: OK. However, we can't just allow subset names to be _anything_, because they're used as path components to locate image files (e.g. we shouldn't allow `../../../..` to be used as a subset name, since then we would end up accessing data from outside the dataset). Is there some pre-existing functionality in Datumaro to check if a name is path-safe?

datumaro/plugins/open_images_format.py
```diff
@@ -0,0 +1,155 @@
...
+                    image_id = image_description['ImageID']
+                    assert _RE_IMAGE_ID.fullmatch(image_id)
+                    assert image_id not in items_by_id
+
+                    subset = image_description['Subset']
+                    assert subset in {'test', 'train', 'validation'}
+
+                    items_by_id[image_id] = DatasetItem(
+                        id=image_id,
+                        image=osp.join(self._dataset_dir, f'images/{subset}/{image_id}.jpg'),
```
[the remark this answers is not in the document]
IRDonch: Maybe, but how would I find out which extension is used in the dataset? The metadata files don't contain this information.

datumaro/plugins/open_images_format.py
```diff
@@ -0,0 +1,155 @@
...
+        set_parents_from_node(root_node, root_category)
+
+    def _load_items(self):
+        items_by_id = {}
+
+        def load_from(path):
+            with self._open_csv_annotation(path) as image_reader:
+                for image_description in image_reader:
+                    image_id = image_description['ImageID']
+                    assert _RE_IMAGE_ID.fullmatch(image_id)
```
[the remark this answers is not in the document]
IRDonch: The situation with image IDs is the same as with subset names - they are used in file names, so some sort of validation is required. I can loosen the restrictions, though, as the only real constraint is that the IDs must be filename-safe.

datumaro/plugins/open_images_format.py
zhiltsov-max: No, there is no such functionality, but disallowing slashes looks like a reasonable limitation.

datumaro/plugins/open_images_format.py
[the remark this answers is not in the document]
zhiltsov-max: Look at Pascal VOC, for example. Use `datumaro.util.image.find_images()`

datumaro/plugins/open_images_format.py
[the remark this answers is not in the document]
zhiltsov-max: We don't restrict image ids in having slashes, because they can be a relative paths. The only exception I can imagine - we need to check for a leading slash, because this would define an absolute path. I don't think we must restrict them in any else way, because formats can have different ways to store data.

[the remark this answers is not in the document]
IRDonch: I think we would also need to prohibit `..` path components, since that would allow accessing files outside the dataset.

datumaro/plugins/open_images_format.py
IRDonch: I don't understand this. If you can only work with a single subset annotation file, doesn't that mean that the imported dataset can never contain more than one subset? Doesn't this provide less functionality to the user?

datumaro/plugins/open_images_format.py
IRDonch: `iglob` I can understand, but how would `listdir` help me here?

datumaro/plugins/open_images_format.py
IRDonch: OK, updated.

datumaro/plugins/open_images_format.py
```diff
@@ -0,0 +1,155 @@
...
+                    assert image_id not in items_by_id
```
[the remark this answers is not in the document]
IRDonch: I replaced all asserts that are unrelated to name validation with raising appropriate exceptions (I'll deal with the remaining asserts once I rework the name validation itself).

datumaro/plugins/open_images_format.py
IRDonch: Changed to public; mind explaining why, though?

datumaro/plugins/open_images_format.py
zhiltsov-max: Multiple subsets and tasks are supported by `Importer`s, which create enough quantity of `Extractor`s. Then, annotations are merged.

datumaro/plugins/open_images_format.py
```diff
@@ -0,0 +1,155 @@
...
+    def _load_categories(self):
+        label_categories = LabelCategories()
+
+        with self._open_csv_annotation('oidv6-class-descriptions.csv') as class_description_reader:
+            for class_description in class_description_reader:
+                label_categories.add(class_description['LabelName'])
+
+        self._categories[AnnotationType.label] = label_categories
+
+        self._load_label_category_parents()
```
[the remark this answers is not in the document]
IRDonch: OK, made it optional.

datumaro/plugins/open_images_format.py
zhiltsov-max: "Access modifiers" doesn't make much sense for plugins. We only looks for specific symbols in plugins + _technically_, such bookkeeping dataset information might be useful outside.

datumaro/plugins/open_images_format.py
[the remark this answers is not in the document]
IRDonch: Okay, implemented (incidentally, this also removes the need to validate the image IDs).

datumaro/plugins/open_images_format.py
IRDonch: Okay; but doesn't this mean that the shared metadata files (like class descriptions and image descriptions) will have to be redundantly parsed multiple times (for each task)?

zhiltsov-max: They might. Please, read the former comment.

datumaro/plugins/open_images_format.py
```diff
@@ -92,23 +94,32 @@ def set_parents_from_node(node, category):
...
-                    assert _RE_IMAGE_ID.fullmatch(image_id)
-
                     if image_id in items_by_id:
                         raise RepeatedItemError(item_id=image_id)
 
                     subset = image_description['Subset']
                     assert subset in {'test', 'train', 'validation'}
 
+                    if image_id not in image_paths_by_id:
+                        raise ItemWithMissingImageError(
```
zhiltsov-max, line 117: In Datumaro we don't consider this as an error necessarily. We work with annotations in the first place. An input dataset might not have images at all, but it needs to have their metainfo in annotations.

datumaro/plugins/open_images_format.py
IRDonch: Well, I'm not sure if this is what you meant, but I redid the implementation of `_glob_annotations` so that the directory is only listed once.

datumaro/plugins/open_images_format.py
IRDonch: Well, I'm confused again. Should I switch to using `SourceExtractor` or not?

datumaro/plugins/open_images_format.py
IRDonch: OK; I removed the error and instead made the image optional.

datumaro/plugins/open_images_format.py
```diff
@@ -0,0 +1,155 @@
...
+                    load_from(path)
+
+        self._items.extend(items_by_id.values())
+
+        self._load_labels(items_by_id)
+
+    def _load_labels(self, items_by_id):
+        label_categories = self._categories[AnnotationType.label]
+
+        # TODO: implement reading of machine-annotated labels
```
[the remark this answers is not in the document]
IRDonch: I'll definitely think about it, but I was going to leave this for a future PR. :)

datumaro/plugins/open_images_format.py
```diff
@@ -0,0 +1,155 @@
...
+                        subset=subset,
+                    )
+
+        # It's preferable to load the combined image description file,
+        # because it contains descriptions for training images without human-annotated labels
+        # (the file specific to the training set doesn't).
+        # However, if it's missing, we'll try loading subset-specific files instead, so that
+        # this exporter can be used on individual subsets of the dataset.
+        try:
+            load_from(_OpenImagesPath.FULL_IMAGE_DESCRIPTION_NAME)
```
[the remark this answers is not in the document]
IRDonch: > I feel like I'd want to pass an absolute path here. It would be much more self-explanatory.

That would involve duplicating the code that joins the annotation file name with the path of the `annotations` directory. I instead tried to make it clearer by renaming the `load_from` parameter to `annotation_name` (as well as ensuring that it is always a name).

> Maybe, I'd also had this function with a return value.

I don't think that would be any better. The function itself would be longer (as it would need to collect the results into a temporary dictionary), the usages would be longer and more repetitive (since we would need to merge the return values into `items_by_id`), and the runtime would take longer (since the items would first be placed into the temporary dictionary and then merged into another dictionary).

datumaro/plugins/open_images_format.py
zhiltsov-max: By default, you inherit a class from `SourceExtractor`. If you see it works bad, you switch to `Extractor`.

datumaro/plugins/open_images_format.py
```diff
@@ -0,0 +1,174 @@
...
+
+        for label_path in self._glob_annotations('*-human-imagelabels.csv'):
+            with self._open_csv_annotation(label_path) as label_reader:
+                for label_description in label_reader:
+                    image_id = label_description['ImageID']
+                    item = items_by_id[image_id]
+
+                    confidence = float(label_description['Confidence'])
+
+                    if 0.5 < confidence:
```
zhiltsov-max, line 154: Is it a dataset policy to ignore items with low confidence?

datumaro/plugins/open_images_format.py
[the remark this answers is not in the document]
IRDonch: I loosened up the subset name validation rules (and replaced the assert with a custom exception).

datumaro/plugins/open_images_format.py
IRDonch: Well... that's actually what I did. :-) You just didn't see the initial version, because it was already gone by the time I submitted the PR.

datumaro/plugins/open_images_format.py
IRDonch: A confidence below 1/2 indicates that the item is more likely to _not_ belong to the corresponding class than it is to belong to it. In fact, the human-verified label files contain "negative labels", which have a confidence of 0 (when an image definitely doesn't belong to a certain class).

The Datumaro `Label` class, however, doesn't have a confidence measure. If you just load all labels and discard the confidence information, it will not be possible to distinguish positive labels from negative labels, which will make the label annotations effectively useless.

I tried to extract as much information from the annotation files as is possible with the current Datumaro architecture, which meant only loading the positive labels. I think potentially Datumaro could be extended to support label confidence levels or negative labels, at which point it'll be possible to load all labels from OID.

datumaro/plugins/open_images_format.py
IRDonch: I loosened up the subset name validation rules (and replaced the assert with a custom exception).

datumaro/plugins/open_images_format.py
```diff
@@ -0,0 +1,155 @@
...
+
+class _OpenImagesPath:
+    ANNOTATIONS_DIR = 'annotations'
+    FULL_IMAGE_DESCRIPTION_NAME = 'image_ids_and_rotation.csv'
+    SUBSET_IMAGE_DESCRIPTION_PATTERNS = (
+        '*-images-with-rotation.csv',
+        '*-images-with-labels-with-rotation.csv',
+    )
+
+class OpenImagesExtractor(Extractor):
```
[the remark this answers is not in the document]
IRDonch: A quick examination suggests that the older versions of the dataset have the same format, except some files are named differently (and not all types of information are available), so I should be able to support them with minor changes. I'll experiment with it.

datumaro/plugins/open_images_format.py
zhiltsov-max: Arbitrary item and annotation attributes are currently stored in the `attributes` field of these types. Probably, you should add `confidence` as a `score` label attribute, to be consistent with other formats.

zhiltsov-max:
We use squash-merges, you can have any commits you need in a PR.

datumaro/plugins/open_images_format.py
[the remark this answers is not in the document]
zhiltsov-max: How about adding `id` and `subset` name checks into the core codebase? Having a check for abspath and for leading dots looks more or less generally-applicable. As well as for subset name being a valid filename / dirname.

datumaro/plugins/open_images_format.py
IRDonch: OK, I did it. I'm assuming this means that I should no longer drop low-confidence (including 0 confidence) labels.

datumaro/plugins/open_images_format.py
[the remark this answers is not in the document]
IRDonch: It's a good idea, but can we maybe defer it until after this PR? It sounds like a lengthy task in itself, since I would need to make sure that these checks don't interfere with loading the currently-supported datasets.

In addition, I'm not so sure that we can use the same check for every format. Some rules can probably be made universal (e.g. no empty string, no control characters), but some datasets might be okay with, for example, slashes in IDs. (There is sort of a precedent with Open Images itself, since it has _label_ names that contain slashes).

datumaro/plugins/open_images_format.py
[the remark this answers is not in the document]
IRDonch: Okay, here's what I discovered: versions 4, 5 and 6 use the same format with minor differences. Versions 2 and 3 use a different layout, while version 1 uses yet another different layout.

Given this, I think the most practical thing to do is to support versions 4+. Which is what I did.

datumaro/plugins/open_images_format.py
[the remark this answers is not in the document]
zhiltsov-max: I'd vote for another independent PR with this functionality. Slashes in ids are ok, because ids are relative paths inside dataset by design. Everything else - probably, not ok.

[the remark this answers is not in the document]
IRDonch: Created an issue for this: #313.

zhiltsov-max (approved):

PR #301, opened 2021-06-18
Support for CIFAR-100

yasakova-anastasia:
### Summary

- Added support for CIFAR-100
- Added documentation for CIFAR-10/100

### Checklist

### License

```python
#
```


tests/test_cifar_format.py
```diff
@@ -125,6 +125,56 @@ def test_can_save_and_load_empty_image(self):
...
+            DatasetItem(id='image_4', subset='test',
+                image=np.ones((32, 32, 3)),
+                annotations=[Label(1, attributes={
+                    'coarse_label': 0
+                })]
+            )
+        ], categories=[['label_0', 'superclass_0'], ['label_1', 'superclass_0']])
+
+        with TestDir() as test_dir:
+            CifarConverter.convert(source_dataset, test_dir, save_images=True)
```
zhiltsov-max, line 149: How is it different from CIfar10 tests? Can we choose between formats?

yasakova-anastasia: In this test, the label has a parent and each label has a `coarse_label` attribute. No, we cannot choose the format. But if the annotation file contains fine labels and coarse labels, then it will be cifar-100. If this is not the case, then there is no point in using this format.

yasakova-anastasia:
I see two options for presenting CIFAR-100:
1. As it is now. But because the file with labels isn't sorted, the class doesn't match the superclass (the label doesn't store its own superclass).
2. Add to the format the parameters of what type of labels to use: fine, coarse, fine and coarse. If we use fine or coarse, then there will be one label for each item. If we use fine and coarse, then there will be two labels for each item. Then we should probably separate CIFAR-10 and CIFAR-100.

tests/test_cifar_format.py
zhiltsov-max: I feel, we need to make `coarse/fine` information automatically inferred from `categories` and add an option to converter to use 10 or 100 format.

zhiltsov-max (commented):
Please update the format docs.

yasakova-anastasia:
> Please update the format docs.

Added documentation for Cifar.

zhiltsov-max:
There is still no information about differences between 10 and 100 versions and how to export in a specific version.

yasakova-anastasia:
> There is still no information about differences between 10 and 100 versions and how to export in a specific version.

Fixed.

zhiltsov-max (approved):

PR #308, opened 2021-06-22
Reduce duplication of dependency information

IRDonch:
### Summary

Currently, all runtime dependencies are listed in both `requirements.txt` and `setup.py`, except that they sometimes have differing version constraints. This is confusing and increases the amount of work a maintainer needs to do when updating the dependency list.

Fix it by moving all runtime dependencies to a separate file (`requirements-core.txt`) and referencing that file from both `setup.py` and `requirements.txt`. Combine all version constraints from both original files.

Unfortunately, some things still need to be duplicated:

* The OpenCV dependency varies depending on environment variables, and that logic can't be replicated with a requirements file. So OpenCV has to be listed separately in `setup.py` and `requirements.txt`.

* Cython is a build dependency, not a runtime dependency, so it shouldn't go into `requirements-core.txt`. It would technically be possible to move it to a separate requirements file and reference from there, but it's just one dependency, so it doesn't seem worth it.

### How to test
Try to run `pip install -r requirements.txt`. (The installation through `pip install .` is covered by CI.)

### Checklist

- ~~[ ] I have updated the [documentation](
  https://github.com/openvinotoolkit/datumaro/tree/develop/docs) accordingly~~ No update should be needed
- ~~[ ] I have added tests to cover my changes~~
- ~~[ ] I have [linked related issues](
  https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)~~

### License

```python
#
```

### DIFF
--- a/requirements-core.txt
+++ b/requirements-core.txt
@@ -1,0 +1,21 @@
+attrs>=19.3.0
+defusedxml>=0.6.0
+GitPython>=3.0.8
+lxml>=4.4.1
+matplotlib>=3.3.1
+numpy>=1.17.3
+Pillow>=6.1.0
+
+# Avoid 2.0.2 Linux binary distribution because of
+# a conflict in numpy versions with TensorFlow:
+# - TF is compiled with numpy 1.19 ABI
+# - pycocotools is compiled with numpy 1.20 ABI
+# Using a previous version allows to force package rebuilding.
+#
+# https://github.com/openvinotoolkit/datumaro/issues/253
+pycocotools>=2.0.0,!=2.0.2; platform_system != "Windows"
+
+pycocotools-windows; platform_system == "Windows"
+PyYAML>=5.3.1
+scikit-image>=0.15.0
+tensorboardX>=1.8
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Tensorflow AVX check is made optional in API and is disabled by default (<https://github.com/openvinotoolkit/datumaro/pull/305>)
 - Extensions for images in ImageNet_txt are now mandatory (<https://github.com/openvinotoolkit/datumaro/pull/302>)
+- Several dependencies now have lower bounds (<https://github.com/openvinotoolkit/datumaro/pull/308>)
 
 ### Deprecated
 -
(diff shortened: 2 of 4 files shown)


requirements-core.txt
```diff
@@ -0,0 +1,21 @@
+attrs>=19.3.0
+defusedxml>=0.6.0
+GitPython>=3.0.8
+lxml>=4.4.1
+matplotlib>=3.3.1
+numpy>=1.17.3
+Pillow>=6.1.0
+
+# Avoid 2.0.2 Linux binary distribution because of
```
zhiltsov-max, line 9: Technically, in requirements.txt we could just use --no-binary, and I'd prefer this solution. But it is not supported in setup.py.

zhiltsov-max (commented):
What do you think about adding a couple of extra requirement files? 
`-testing` with pytest
`-ci` with coverage
`-plugins` with TF, TF-gpu, pandas?

zhiltsov-max (commented):
Also, you asked about fixed dependency versions in setup.py - consult this doc https://packaging.python.org/discussions/install-requires-vs-requirements/.

IRDonch:
> What do you think about adding a couple of extra requirement files?
> `-testing` with pytest
> `-ci` with coverage
> `-plugins` with TF, TF-gpu, pandas?

I think the most sensible scheme would be as follows:

* `requirements-core.txt` - core runtime dependencies (the file I created in this PR)
* For each plugin that has additional dependencies, `requirements-<plugin>.txt`. The `setup.py` equivalent would be an entry in `extras-require`.
* `requirements-dev.txt` - development dependencies (pytest, pylint, etc.). I don't think we need to split this into test, coverage or whatever.
* `requirements.txt` - master file that includes all of the above.

That said, the above scheme only makes sense if we want to keep runtime dependencies in `requirements.txt` files, which I'm not sure about. A simpler scheme would be to keep them only in `setup.py` and have only a `requirements-dev.txt` and a master `requirements.txt` file that just contains:

```
-e .[plugin1,plugin2]
-r requirements-dev.txt
```

IRDonch:
> Also, you asked about fixed dependency versions in setup.py

I... don't think I did, though? :-) I don't think `setup.py` should have fixed dependency versions, and the document you linked agrees with me.

IRDonch:
So, what now? Should I change anything?

zhiltsov-max (approved):

zhiltsov-max:
Do you want to add anything into CONTRIBUTING.md?

IRDonch:
> Do you want to add anything into CONTRIBUTING.md?

Not really. I don't think anything I did here conflicts with anything that `CONTRIBUTING.md` says.

PR #316, opened 2021-06-29
Add python bandit checks.

dvkruchinin:
### Summary

Adding a check using linters. PyLint, Bandit, Remark.

### Checklist

### License

```python
#
```


zhiltsov-max:
Please update to the latest develop to fix build.

requirements.txt
```diff
@@ -2,4 +2,6 @@ Cython>=0.27.3 # include before pycocotools
 -r requirements-core.txt --no-binary=pycocotools # https://github.com/openvinotoolkit/datumaro/issues/253
 opencv-python-headless>=4.1.0.25
 pandas>=1.1.5
-pytest>=5.3.5
\ No newline at end of file
+pytest>=5.3.5
+pylint==2.7.0
+bandit==1.7.0
```
zhiltsov-max, line 7:
```suggestion
pylint>=2.7.0
bandit>=1.7.0
```

.remarkrc.js
```diff
@@ -0,0 +1,18 @@
+exports.settings = { bullet: '*', paddedTable: false };
+
+exports.plugins = [
+    'remark-frontmatter',
+    'remark-gfm',
+    'remark-preset-lint-recommended',
+    'remark-preset-lint-consistent',
+    ['remark-lint-list-item-indent', 'space'],
+    ['remark-lint-no-dead-urls', false], // Does not work because of github protection system
+    ['remark-lint-maximum-line-length', 120],
```
zhiltsov-max, line 10:
```suggestion
    ['remark-lint-maximum-line-length', 80],
```

Does it ignore long URLs?

json_to_html.py
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env python
```
zhiltsov-max, line 1: Maybe it's better to be put to `/ci/` or something? I don't think this file should be in the root.

dvkruchinin:
> Please update to the latest develop to fix build.

Done.

requirements.txt
dvkruchinin: Applied locally.

.remarkrc.js
dvkruchinin: Applied locally.

json_to_html.py
dvkruchinin: Moved to ```ci``` folder.

zhiltsov-max: Maybe we need to exclude it from `setup.py`. As well as `docs/`. They are not modules, though.

zhiltsov-max:
Is it possible to print check summary in the console log?

json_to_html.py
dvkruchinin: Added these directories by analogy with ```tests```

dvkruchinin:
> Is it possible to print check summary in the console log?

What is the best way to do it? Report to the console, html? These options have not been considered before. Therefore, it is difficult to say now how readable they will be. Need to check it out.

zhiltsov-max:
> What is the best way to do it?

I think, it would be useful to print warnings in some condensed format. But it is not critical. If it requires big efforts, let's ignore it.

.github/workflows/pylint.yml
```diff
@@ -0,0 +1,44 @@
...
+          PR_FILES=$(curl -s -X GET -G $URL | jq -r '.[] | select(.status != "removed") | .filename')
+          for files in $PR_FILES; do
+            extension="${files##*.}"
+            if [[ $extension == 'py' ]]; then
+              changed_files_pylint+=" ${files}"
+            fi
+          done
+
+          if [[ ! -z ${changed_files_pylint} ]]; then
+            sudo apt-get --no-install-recommends install -y build-essential curl python3-dev python3-pip python3-venv
```
IRDonch, line 21: You should use the `actions/setup-python` action to set up Python. Also, `curl` has to already be installed at this point, or else the `curl` run above would have failed.

.github/workflows/pylint.yml
```diff
@@ -0,0 +1,44 @@
+name: Linter
```
IRDonch, line 1: All workflows should be named differently.

.github/workflows/pylint.yml
```diff
@@ -0,0 +1,44 @@
...
+            . .env/bin/activate
+            pip install -U pip wheel setuptools
+            pip install pylint-json2html
+            pip install $(egrep "pylint.*" ./requirements.txt)
+            mkdir -p pylint_report
+
+            echo "Pylint version: "`pylint --version | head -1`
+            echo "The files will be checked: "`echo ${changed_files_pylint}`
+            pylint ${changed_files_pylint} --output-format=json > ./pylint_report/pylint_checks.json || exit_code=`echo $?` || true
+            pylint-json2html -o ./pylint_report/pylint_checks.html ./pylint_report/pylint_checks.json
```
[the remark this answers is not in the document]
dvkruchinin: I asked above how to do it better. CVAT uses this approach.

json_to_html.py
IRDonch: > Maybe we need to exclude it from `setup.py`. As well as `docs/`. Even though they are not modules yet.

FWIW, I think we should move to a `src`-based layout (i.e., create an `src` directory, move `datumaro` there and modify `setup.py` to only look for packages there) to prevent unwanted files from being included in the distribution by accident.

.github/workflows/pylint.yml
dvkruchinin: It seems to me that this will only complicate the reading of the config. Besides, there is no difference. But if it is so necessary, I will do it through ```actions```.

.github/workflows/pylint.yml
dvkruchinin: Not obligatory. There are no tools used to work with workflow through the name. The display of the job name in the checkers field will also depend on this.

.github/workflows/pylint.yml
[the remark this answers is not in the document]
IRDonch: Well, IMO the way to do it better is just... to run pylint and not do anything with the output.

.github/workflows/pylint.yml
IRDonch: There _is_ a difference. With `actions/setup-python`, you can install a specific version of Python, while here you install whatever the default version of Python is for the current OS, which could change at any time and break everything.

.github/workflows/pylint.yml
IRDonch: OK, fair enough.

.github/workflows/pylint.yml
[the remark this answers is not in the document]
dvkruchinin: Okay, I'll rework this point. Should I do the same for all linters? The results of the check will be output to the console.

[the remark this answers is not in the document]
IRDonch: IMO, yes, that would be the best. @zhiltsov-max Do you have any objections to this?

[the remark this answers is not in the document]
zhiltsov-max: Maybe, I'd keep a file with the complete output, but having a simple report in logs could save some time.

.github/workflows/pylint.yml
dvkruchinin: Reworked.

.github/workflows/pylint.yml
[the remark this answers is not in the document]
dvkruchinin: Applied

json_to_html.py
dvkruchinin: Since it was decided to form a linter report in the console, a file for generating an html report for the remark linter is not needed. I need to return setup.py to its state in the develop branch?

dvkruchinin:
Combined with #321

.github/workflows/bandit.yml
```diff
@@ -0,0 +1,16 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+
+      - name: Run checks
+        run: |
+          pip install --user -r <(grep "^bandit" ./requirements.txt)
+          echo "Bandit version: "`bandit --version | head -1`
```
zhiltsov-max, line 15: I won't say it is a problem, but this line looks a little bit extra, because the previous line outputs the version installed.

datumaro/components/operations.py
```diff
@@ -1292,7 +1292,7 @@ def _default_hash(item):
             log.warning("Item (%s, %s) has no image "
                 "info, counted as unique", item.id, item.subset)
             return None
-        return hashlib.md5(item.image.data.tobytes()).hexdigest()
+        return hashlib.md5(item.image.data.tobytes()).hexdigest() # nosec - disable B303:md5 check
```
zhiltsov-max, line 1295: Probably, I'd better have comments about the reasons to ignore, instead of ignored rules. 

```suggestion
        # ignore B303 (md5 check), because it is not used in security context
        return hashlib.md5(item.image.data.tobytes()).hexdigest() # nosec
```

zhiltsov-max (changes requested):

datumaro/components/operations.py
IRDonch: Why not just change it to use a non-obsolete hash algorithm?

zhiltsov-max: We need any fast hashing algo here, which one would you suggest?

IRDonch: SHA-256? 🤷

zhiltsov-max: Looking at https://stackoverflow.com/questions/10070293/choosing-a-hash-function-for-best-performance, I can conclude, that md5 is a good baseline. There are few other algos, which up to 2 times faster, everything else is worse. CRC32 might be good candidate, but it has worse entropy, so we can expect more collisions.

IRDonch: Okay, fair enough.

datumaro/plugins/cifar_format.py
```diff
@@ -4,7 +4,7 @@
 
 import os
 import os.path as osp
-import pickle
+import pickle # nosec - disable B403:import_pickle check
```
zhiltsov-max, line 7: This is an RCE, and it must be fixed.

datumaro/util/tf_util.py
```diff
@@ -16,7 +16,7 @@ def check_import():
 
     from .os_util import check_instruction_set
 
-    result = subprocess.run([sys.executable, '-c', 'import tensorflow'],
+    result = subprocess.run([sys.executable, '-c', 'import tensorflow'], # nosec - disable B603:subprocess_without_shell_equals_true check
```
zhiltsov-max, line 19:
```suggestion
    # Disable B603:subprocess_without_shell_equals_true - the command line is controlled
    result = subprocess.run([sys.executable, '-c', 'import tensorflow'], # nosec
```

zhiltsov-max (approved):

PR #319, opened 2021-06-29
Fix COCO Panoptic

### LINKED ISSUE #317: Potential integer overflow
In the code below we see that `img` can be of type `uint8`, then `img` is multiplied by 2<sup>8</sup> and 2<sup>16</sup>, but `uint8` cannot store such large numbers.
https://github.com/openvinotoolkit/datumaro/blob/5209d42348b001b2f07a917d5dd7b76f3469c7ce/datumaro/util/mask_tools.py#L116-L119

kirill-sizov:
### Summary
Fixed #317.

### How to test
Add the original [COCO Panoptic dataset](https://cocodataset.org/#download) to the Datumaro project and then export it from the project.

### Checklist

### License

```python
#
```


datumaro/util/mask_tools.py
```diff
@@ -114,8 +114,8 @@ def make_binary_mask(mask):
     return mask.astype(bool)
 
 def bgr2index(img):
-    if img.dtype.kind not in {'b', 'i', 'u'}:
-        img = img.astype(np.uint8)
+    if img.dtype.kind not in {'b', 'i', 'u'} or img.dtype.itemsize < 4:
+        img = img.astype(np.uint32)
```
zhiltsov-max, line 118: Please check the calling contexts if they relied on number range narrowing from the previous conversion.

kirill-sizov: I found only one context in which this function is called: function `bgr2index` is called in only one place to convert the bgr mask to the COCO index representation, and after this conversion, the calling context does not depend on the narrowing of the range:
https://github.com/openvinotoolkit/datumaro/blob/e8371e917a525d3026b547c717a1c2dfacd0a921/datumaro/plugins/coco_format/extractor.py#L187-L191
In the code above, we can see that the `mask` is redefined by the return value of the `bgr2index` function. But the `bgr2index` function can return a mask with values greater than 255, so that means we have to expand the mask's value type.

zhiltsov-max (approved):
Nice observation!

PR #245, opened 2021-05-18
CVAT-3D Milestone-6 PCD and Kitti formats

manasars:
Added feature for supporting import and export of PCD and Kitti format files in CVAT.

- [x] I submit my changes into the `develop` branch
- [x] I have added description of my changes into [CHANGELOG](https://github.com/openvinotoolkit/datumaro/blob/develop/CHANGELOG.md)

### License

- [x] I submit _my code changes_ under the same [MIT License](
  https://github.com/opencv/cvat/blob/develop/LICENSE) that covers the project.

Issue #310, opened 2021-06-24
Update type definitions wrt new attrs features


zhiltsov-max:
Review and update `extractor.py` and other files.

> Changed in version 21.1.0: init=False injects \_\_attrs_init\_\_ 

Should fix our class patching in Bbox.

zhiltsov-max:
There are more things we can do:
- utilize the new library API (https://www.attrs.org/en/stable/api.html#next-generation-apis): define, field, frozen
- remove `repr_ns` in nested classes

IRDonch:
FWIW, the attrs changelog promises that in the next release you'll be able to "`import attrs`", so I think we should wait and see what that's like before doing a mass refactoring. My assumption is that it'll be a package with just the new generation API in it, so it'll be easier to remove old API usage if we switch to it.

PR #320, opened 2021-06-29
Fix cuboids / 3d / M6

zhiltsov-max:
### Summary

Fixes #245 

- Rewritten supervisely pointcloud format
- Rewritten kitti raw 3d format
- `point_cloud` format renamed to `sly_pointcloud`. Regular (KITTI PCD) point clouds require a separate format.
- Added `Cuboid3d` annotation type
- Added point clouds and related images in `DatasetItem`
- Added docs for new formats

TBD:
- [ ] Update datumaro format
- [ ] Implement cuboid comparison

### Checklist

### License

```python
#
```


datumaro/components/converter.py
```diff
@@ -62,12 +66,17 @@ def _make_image_filename(self, item, *, name=None, subdir=None):
         subdir = subdir or ''
         return osp.join(subdir, name + self._find_image_ext(item))
 
+    def _make_pcd_filename(self, item, *, name=None, subdir=None):
+        name = name or item.id
+        subdir = subdir or ''
+        return osp.join(subdir, name + '.pcd')
```
IRDonch, line 72: This function mostly replicates `_make_image_filename` - perhaps you could factor the common part out?

datumaro/components/converter.py
```diff
@@ -76,15 +85,26 @@ def _save_image(self, item, path=None, *,
             self._make_image_filename(item, name=name, subdir=subdir))
         path = osp.abspath(path)
 
-        src_ext = item.image.ext.lower()
-        dst_ext = osp.splitext(osp.basename(path))[1].lower()
+        item.image.save(path)
+
+    def _save_pcd(self, item=None, path=None, *,
```
IRDonch, line 90: Similarly, this function is very similar to `_save_image` - maybe some common part could be extracted here too?

datumaro/components/converter.py
```diff
@@ -76,15 +85,26 @@ def _save_image(self, item, path=None, *,
...
 
         os.makedirs(osp.dirname(path), exist_ok=True)
-        if src_ext == dst_ext and osp.isfile(item.image.path):
-            if item.image.path != path:
-                shutil.copyfile(item.image.path, path)
-        elif src_ext == dst_ext and isinstance(item.image, ByteImage):
+        if item.pcd and osp.isfile(item.pcd):
+            if item.pcd != path:
+                shutil.copyfile(item.pcd, path)
+        elif isinstance(item.pcd, bytes):
```
IRDonch, line 108: This branch should be first, because otherwise it's possible that the `bytes` value gets interpreted as a path instead (very unlikely, but still).

datumaro/components/extractor.py
```diff
@@ -126,7 +126,8 @@ def __iter__(self):
 @attrs
 class Label(Annotation):
     _type = AnnotationType.label
-    label = attrib(converter=int)
+    label = attrib(converter=int, default=0)
```
IRDonch, line 129: Giving `label` a default value seems questionable - why would one need to create a `Label` without the label index?

datumaro/components/extractor.py
```diff
@@ -359,6 +360,69 @@ def as_polygon(self):
     def get_area(self):
         return 0
 
+
+@attrs
+class Cuboid3d(Annotation):
```
IRDonch, line 365: Isn't a cuboid inherently 3D? Seems like a redundant name.

datumaro/components/extractor.py
```diff
@@ -359,6 +360,69 @@ def as_polygon(self):
...
+        return self._points[6:9]
+
+    @scale.setter
+    def _set_scale(self, value):
+        self.scale[:] = \
+            [round(p, _COORDINATE_ROUNDING_DIGITS) for p in value]
+
+assert not hasattr(Cuboid3d, '__attrs_init__') # hopefully, it will be supported
+setattr(Cuboid3d, '__attrs_init__', Cuboid3d.__init__)
+setattr(Cuboid3d, '__init__', Cuboid3d.__actual_init__)
```
IRDonch, line 424: Seems like you can accomplish the desired behavior without manual manipulation by using `init=False` (see https://www.attrs.org/en/stable/init.html#hooking-yourself-into-initialization).

datumaro/components/extractor.py
```diff
@@ -517,9 +581,40 @@ class DatasetItem:
         type=str, validator=not_empty)
     annotations = attrib(factory=list, validator=default_if_none(list))
     subset = attrib(converter=lambda v: v or DEFAULT_SUBSET_NAME, default=None)
+
+    # Currently unused
     path = attrib(factory=list, validator=default_if_none(list))
 
+    # TODO: introduce "media" field with type info. Replace image and pcd.
     image = attrib(type=Image, default=None)
+    pcd = attrib(type=bytes, default=None)
```
IRDonch, line 590: Perhaps for clarity it would be better to use a longer name, e.g. `point_cloud`?

datumaro/components/extractor.py
```diff
@@ -517,9 +581,40 @@ class DatasetItem:
...
+    related_images = attrib(type=list, default=None)
+
+    def __attrs_post_init__(self):
+        assert not (self.has_image and self.has_pcd), \
+            "Can't set both image and point cloud info"
+
+    @related_images.validator
+    def _related_image_validator(self, attribute, related_images):
```
IRDonch, line 598: This looks more like a converter than a validator.

datumaro/components/extractor.py
```diff
@@ -517,9 +581,40 @@ class DatasetItem:
...
+                image = Image(path=image)
+            assert isinstance(image, Image), type(image)
+
+            self.related_images.append(image)
+
+    @pcd.validator
+    def _pcd_validator(self, attribute, pcd):
+        if pcd:
+            assert isinstance(pcd, (bytes, str)) or callable(pcd), type(pcd)
+            self.pcd = pcd
```
IRDonch, line 616: I don't think a validator is supposed to set the attribute itself (wouldn't that cause a recursive loop anyway?). You just need to check the type.

datumaro/components/operations.py
```diff
@@ -825,15 +832,7 @@ class _ShapeMerger(AnnotationMerger, _ShapeMatcher):
...
-            label, label_score = self.find_cluster_label(cluster)
-            shape, shape_score = self.merge_cluster_shape(cluster)
-
-            shape.z_order = max(cluster, key=lambda a: a.z_order).z_order
-            shape.label = label
-            shape.attributes['score'] = label_score * shape_score \
-                if label is not None else shape_score
-
-            merged.append(shape)
+            merged.append(self.merge_cluster(cluster))
```
IRDonch, line 835: Seems like this whole function could now be reduced to `return list(map(self.merge_cluster, clusters))`.

datumaro/components/extractor.py
```diff
@@ -359,6 +360,69 @@ def as_polygon(self):
...
+    # will be overridden by attrs, then will be overridden again by us
+    # attrs' method will be renamed to __attrs_init__
+    def __init__(self, position, rotation=None, scale=None, **kwargs):
+        assert len(position) == 3, position
+        if not rotation:
+            rotation = [0] * 3
+        if not scale:
+            scale = [1] * 3
+        kwargs.pop('points', None)
+        self.__attrs_init__(points=[*position, *rotation, *scale], **kwargs)
```
IRDonch, line 389: `position` is a point, but `rotation` and `scale` surely aren't. Why do you store them in one list?

IRDonch (changes requested):
I have reviewed the core part so far.

datumaro/components/extractor.py
zhiltsov-max: No, there are 2d cuboids, for example, in CVAT. https://openvinotoolkit.github.io/cvat/docs/manual/advanced/annotation-with-cuboids/creating-the-cuboid/. 3d extend them, of course, but it looks like they need another hierarchy with 3d objects.

datumaro/components/extractor.py
zhiltsov-max: Right, copy-pasted.

datumaro/components/extractor.py
zhiltsov-max: It was added in newer versions of attrs, so it needs a general refactoring. I've opened an issue for this: #310

datumaro/components/extractor.py
IRDonch: Okay, I guess that makes sense. 

(IMO, a cuboid is still 3D, and the thing in CVAT should be called something else, e.g. "flat cuboid", but it's subjective, so I won't insist on it.)

datumaro/util/test_utils.py
```diff
@@ -155,6 +208,9 @@ def test_save_and_load(test, source_dataset, converter, test_dir, importer,
     if target_dataset is None:
         target_dataset = source_dataset
 
-    if not compare:
+    if not compare and kwargs.get("dimension") is Dimensions.dim_3d:
+        compare = compare_datasets_3d
+        del kwargs["dimension"]
+    elif not compare:
```
IRDonch, line 214: It seems like the sole purpose of the `dimension` argument is to select a different default comparison function, which raises the question: why even add that argument? It's not any more difficult for the client of the function to pass `compare=compare_datasets_3d` than it is to pass `dimension=Dimensions.dim_2d`.

docs/formats/sly_pointcloud_user_manual.md
```diff
@@ -0,0 +1,183 @@
...
+- `createdAt` (write, string),
+- `updatedAt` (write, string),
+- `labelerLogin` (write, string), responsible for the corresponding fields
+  in the annotation file.
+- `frame` (read/write, integer). Indicates frame number of the image.
+- arbitrary attributes
+
+## Import Supervisely Point Cloud dataset
+
+An example datast in Supervisely Point Cloud format is available for download:
```
IRDonch, line 39:
```suggestion
An example dataset in Supervisely Point Cloud format is available for download:
```

docs/formats/sly_pointcloud_user_manual.md
```diff
@@ -0,0 +1,183 @@
...
+
+<https://drive.google.com/u/0/uc?id=1BtZyffWtWNR-mk_PHNPMnGgSlAkkQpBl&export=download>
```
IRDonch, line 41: Where does this example come from?

docs/formats/sly_pointcloud_user_manual.md
```diff
@@ -0,0 +1,183 @@
...
+information.
+
+## Export to other formats
+
+Datumaro can convert Supervisely Point Cloud dataset into any other
+format [Datumaro supports](../user_manual.md#supported-formats).
+
+Such conversion will only be successful if the output
+format can represent the type of dataset you want to convert,
+e.g. 3d point clouds annotations can be
```
IRDonch, line 86:
```suggestion
e.g. 3D point cloud annotations can be
```

docs/formats/sly_pointcloud_user_manual.md
```diff
@@ -0,0 +1,183 @@
...
+saved in `KITTI Raw` format, but not as `COCO keypoints`.
```
IRDonch, line 87: The inline code style seems inappropriate here. These aren't identifiers or code fragments.

docs/formats/sly_pointcloud_user_manual.md
```diff
@@ -0,0 +1,183 @@
...
+
+There are few ways to convert Supervisely Point Cloud dataset
```
IRDonch, line 89:
```suggestion
There are a few ways to convert Supervisely Point Cloud dataset
```

docs/formats/sly_pointcloud_user_manual.md
```diff
@@ -0,0 +1,183 @@
...
+        pcd='path/to/pcd2.pcd', related_images=['image2.png'],
+        attributes={'frame': 1}
+    ),
+], categories=['cat', 'dog'])
+
+dataset.export('my_dataset/', format='sly_pointcloud', save_images=True,
+    allow_undeclared_attrs=True)
+```
+
+More examples of working with KITTI Raw dataset from code can be found in
```
IRDonch, line 182:
```suggestion
More examples of working with Supervisely Point Cloud dataset from code can be found in
```

datumaro/plugins/sly_pointcloud_format/extractor.py
```diff
@@ -0,0 +1,187 @@
...
+    AnnotationType, Cuboid3d, DatasetItem, Importer, LabelCategories,
+    SourceExtractor,
+)
+from datumaro.util.image import find_images
+
+from .format import PointCloudPath
+
+
+class SuperviselyPointcloudExtractor(SourceExtractor):
+    NAME = 'sly_pointcloud'
```
IRDonch, line 19:
```suggestion
class SuperviselyPointCloudExtractor(SourceExtractor):
    NAME = 'sly_point_cloud'
```

"Pointcloud" is not a word, and you consistently spell it as two words in the documentation, so it should be treated as two words in the code, as well.

datumaro/components/extractor.py
zhiltsov-max: For memory and processing efficiency. You're not completely true about rotation and scale, because they all are N-d vectors. It's like 4x4 quaternions.

datumaro/components/extractor.py
zhiltsov-max: You're right, but converters have lesser capabilities than validators, so I decided to avoid using converters in general.

docs/formats/sly_pointcloud_user_manual.md
zhiltsov-max: Actually, I don't know. @nmanovic, how did it get in SOW?

datumaro/util/test_utils.py
zhiltsov-max: The only benefit I see is the higher level of abstraction when the enum is used.

datumaro/components/converter.py
zhiltsov-max: Probably, we need to delete it at all.

datumaro/components/converter.py
zhiltsov-max: Refactored

datumaro/components/converter.py
zhiltsov-max: I have no idea currently how to do it better. Probably, it will be done in future, when generic `media` field is introduced in `DatasetItem`.

datumaro/components/extractor.py
zhiltsov-max: Done

datumaro/components/extractor.py
zhiltsov-max: Refactored

datumaro/components/extractor.py
zhiltsov-max: Validators and converters only work on initialization. Refactored.

datumaro/components/operations.py
zhiltsov-max: Done

docs/formats/sly_pointcloud_user_manual.md
zhiltsov-max: "a few" would mean more than just "few". And we have exactly 2.

PR #322, opened 2021-07-01
Fix pylint issues. Added pylint checking.

dvkruchinin:
### Summary
Fix pylint issues for which there are no questions. Added pylint checking.

### Checklist

### License

```python
#
```


.github/workflows/pylint.yml
```diff
@@ -0,0 +1,16 @@
...
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+
+      - name: Run checks
+        run: |
+          pip install --user $(egrep "pylint.*" ./requirements.txt)
+          echo "Pylint version: "`pylint --version | head -1`
+          pylint -r n `find . -name "*.py" -type f | awk -F "/" '{print $2}' | uniq`
```
IRDonch, line 16: This fails to scan some of the files, such as those inside `datumaro/plugins/openvino_plugin/samples`. I would suggest using `git ls-files -z '*.py' | xargs -0 pylint -r n` instead.

.github/workflows/pylint.yml
```diff
@@ -0,0 +1,16 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+
+      - name: Run checks
+        run: |
+          pip install --user $(egrep "pylint.*" ./requirements.txt)
```
IRDonch, line 14:
```suggestion
          pip install --user -r <(grep "^pylint" ./requirements.txt)
```

(Using `-r` permits normal requirements file syntax, such as comments).

dvkruchinin:
> FYI, regarding the E1120 errors complaining about `os.path.join` calls, I have filed a false positive report for Pylint: [PyCQA/pylint#4554](https://github.com/PyCQA/pylint/issues/4554). So you can suppress them and link the issue in the comment.

Ok. Done.

.github/workflows/pylint.yml
dvkruchinin: Yes, indeed. Thanks. Fixed it locally.

.github/workflows/pylint.yml
dvkruchinin: Fixed it locally.

datumaro/plugins/accuracy_checker_plugin/details/ac.py
```diff
@@ -3,9 +3,6 @@
 #
 # SPDX-License-Identifier: MIT
 
-from datumaro.util.tf_util import import_tf
```
IRDonch, line 6: I think this import is supposed to go first, because we need to make sure that we import TF before Accuracy Checker does. Can you suppress the Pylint warning just for this import?

dvkruchinin: It was possible to do this for the specified file.

IRDonch: You need to do the same in `representation.py`.

dvkruchinin: Done

datumaro/cli/contexts/project/__init__.py
```diff
@@ -8,8 +8,8 @@
 import os
 import os.path as osp
 import shutil
-import numpy as np
 from enum import Enum
```
zhiltsov-max, line 11: There is no such in PyLint, but please put `from x import y` before `import z`

datumaro/cli/contexts/project/diff.py
```diff
@@ -9,11 +9,10 @@
 import logging as log
 import os
 import os.path as osp
-
+import warnings
 import cv2
```
zhiltsov-max, line 13:
```suggestion

import cv2
```

datumaro/cli/contexts/project/__init__.py
IRDonch: I suggest that we stick to Pylint issues in this PR. We can use `isort` to reorder all imports automatically later.

zhiltsov-max (approved):

PR #326, opened 2021-07-05
Remove Pylint unused-import warning suppressions

IRDonch:
### Summary

This makes all static analyzers happy (rather than just pylint) and ensures
we don't miss any unexpected unused-import issues in the affected files.

### How to test

N/A

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/util/__init__.py
+++ b/datumaro/util/__init__.py
@@ -1,17 +1,18 @@
 
-# Copyright (C) 2019-2020 Intel Corporation
+# Copyright (C) 2019-2021 Intel Corporation
 #
 # SPDX-License-Identifier: MIT
 
 from contextlib import ExitStack
-from distutils.util import strtobool as str_to_bool # pylint: disable=unused-import
+import distutils.util
 from functools import partial, wraps
 from itertools import islice
 from typing import Iterable, Tuple
 import attr
 
-
 NOTSET = object()
+
+str_to_bool = distutils.util.strtobool
 
 def find(iterable, pred=lambda x: True, default=None):
     return next((x for x in iterable if pred(x)), default)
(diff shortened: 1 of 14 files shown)


datumaro/util/image.py
```diff
@@ -18,7 +17,7 @@ class _IMAGE_BACKENDS(Enum):
 _IMAGE_BACKEND = None
 _image_loading_errors = (FileNotFoundError, )
 try:
-    import cv2
+    importlib.import_module('cv2')
     _IMAGE_BACKEND = _IMAGE_BACKENDS.cv2
 except ImportError:
     import PIL
```
zhiltsov-max, line 23: Probably, `PIL` import also deserves a change.

zhiltsov-max (commented):
I'm not sure I like the way it is implemented. It looks like a fight with linting tools.

datumaro/util/image.py
IRDonch: The `PIL` import is not unused, so I can't change it.

IRDonch:
> I'm not sure I like the way it is implemented. It looks like a fight with linting tools.

Why? I think it's a fairly logical convention that imported names are only for use within the imported module, and if you want them to be a part of the module's interface, you should re-export them explicitly.

At least in the case of `datumaro/util/__init__.py`, I consider my version easier to understand than the original (for humans as well as linters). It definitely took me some headscratching to understand the purpose of the original import.

datumaro/util/__init__.py
```diff
@@ -1,18 +1,19 @@
...
 #
 # SPDX-License-Identifier: MIT
 
 from contextlib import ExitStack
-from distutils.util import strtobool as str_to_bool # pylint: disable=unused-import
+import distutils.util
 from functools import partial, wraps
 from itertools import islice
 from typing import Iterable, Tuple
 import attr
```
zhiltsov-max, line 11:
```suggestion
from functools import partial, wraps
from itertools import islice
from typing import Iterable, Tuple
import distutils.util

import attr
```

zhiltsov-max (approved):

PR #343, opened 2021-07-09
Factor out the images.meta loading code from YoloExtractor

IRDonch:
### Summary
It looks like the same thing will be needed for Open Images, so I'm
moving it to a common module.

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/util/image.py
+++ b/datumaro/util/image.py
@@ -4,7 +4,9 @@
 
 from enum import Enum, auto
 from io import BytesIO
-from typing import Any, Callable, Iterable, Iterator, Optional, Tuple, Union
+from typing import (
+    Any, Callable, Dict, Iterable, Iterator, Optional, Tuple, Union,
+)
 import importlib
 import os
 import os.path as osp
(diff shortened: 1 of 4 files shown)


datumaro/util/image.py
```diff
@@ -380,3 +382,32 @@ def save(self, path):
...
+
+    if not osp.isfile(image_meta_path):
+        raise Exception("Can't read image meta file '%s'" % image_meta_path)
+
+    image_meta = {}
+
+    with open(image_meta_path, encoding='utf-8') as f:
+        for line in f:
+            image_name, h, w = line.strip().rsplit(maxsplit=2)
+            image_meta[image_name] = (int(h), int(w))
```
zhiltsov-max, line 411: Probably, we need to skip empty (and maybe commented) lines here too.
```suggestion
            line = line.strip()
            if not line or line.startswith('#'):
                continue
            
            image_name, h, w = line.strip().rsplit(maxsplit=2)
            image_meta[image_name] = (int(h), int(w))
```

IRDonch: I implemented an approach based on `shlex.split` - that handles comments and allows other fields to be added in the future. What do you think?

zhiltsov-max: Maybe it is a good solution, but I find it hard to understand the grammar it describes from the official description (both in POSIX and non-POSIX modes), so I'm having troubles with predicting the possible consequences.

IRDonch: It's just (very simplified) shell syntax. The main consequence is that file names with spaces in them have to be quoted.

zhiltsov-max: Consider updating few other formats to use `shlex`, such syntax is used in few extra dataset files from datumaro - VOC, CamVid and several others.

zhiltsov-max (approved):

datumaro/util/image.py
IRDonch: I will take a look later, but right now I want to get back to Open Images.

FWIW, I found four likely candidates for updating:

```console
$ rg -Fl "line[0] == '#'"
datumaro/plugins/cityscapes_format.py
datumaro/plugins/voc_format/format.py
datumaro/plugins/camvid_format.py
datumaro/plugins/kitti_format/format.py
```

Not sure about LFW.

Issue #142, opened 2021-03-04
Provide a mode to ignore, fix, stop on or report errors in imported datasets


zhiltsov-max:
When a dataset is being imported, its source files can contain errors in annotations. Currently, importing will fail, but there are other  strategies possible. For example, an error can be ignored or fixed automatically. This would make dataset import more tolerant, allowing to partially import corrupted data.

zhiltsov-max:
Related https://github.com/openvinotoolkit/cvat/issues/955
Related https://github.com/openvinotoolkit/cvat/issues/1529

PR #352, opened 2021-07-13
Open Images: implement bounding box support

IRDonch:
### Summary

This continues the implementation of #274.

### How to test

You can download some of the bounding boxes for OID, create `images.meta` as described in the documentation and try to import the dataset.

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/plugins/open_images_format.py
+++ b/datumaro/plugins/open_images_format.py
@@ -12,6 +12,7 @@ import logging as log
 import os
 import os.path as osp
 import re
+import types
 
 from attr import attrs
 
@@ -97,6 +98,15 @@ class OpenImagesPath:
         'IsDepiction',
         'IsInside',
     )
+
+    BBOX_BOOLEAN_ATTRIBUTES = (
+        types.SimpleNamespace(datumaro_name='occluded', oid_name='IsOccluded'),
+        types.SimpleNamespace(datumaro_name='truncated', oid_name='IsTruncated'),
+        types.SimpleNamespace(datumaro_name='is_group_of', oid_name='IsGroupOf'),
+        types.SimpleNamespace(datumaro_name='is_depiction', oid_name='IsDepiction'),
+        types.SimpleNamespace(datumaro_name='is_inside', oid_name='IsInside'),
+    )
+
 
 class OpenImagesExtractor(Extractor):
     def __init__(self, path):
@@ -286,14 +296,20 @@ class OpenImagesExtractor(Extractor):
                     y_min = float(bbox_description['YMin']) * height
                     y_max = float(bbox_description['YMax']) * height
 
-                    confidence = float(bbox_description['Confidence'])
-                    # TODO: other attributes?
+                    attributes = {
+                        'score': float(bbox_description['Confidence']),
+                    }
+
+                    for bool_attr in OpenImagesPath.BBOX_BOOLEAN_ATTRIBUTES:
+                        int_value = int(bbox_description[bool_attr.oid_name])
+                        if int_value >= 0:
+                            attributes[bool_attr.datumaro_name] = bool(int_value)
 
                     item.annotations.append(Bbox(
                         label=label_index,
                         x=x_min, y=y_min,
                         w=x_max - x_min, h=y_max - y_min,
-                        attributes={'score': confidence},
+                        attributes=attributes,
                     ))
 
 
@@ -452,4 +468,9 @@ class OpenImagesConverter(Converter):
                                 'YMin': annotation.y / height,
                                 'XMax': (annotation.x + annotation.w) / width,
                                 'YMax': (annotation.y + annotation.h) / height,
+                                **{
+                                    bool_attr.oid_name:
+                                        int(annotation.attributes.get(bool_attr.datumaro_name, -1))
+                                    for bool_attr in OpenImagesPath.BBOX_BOOLEAN_ATTRIBUTES
+                                },
                             })
--- a/datumaro/plugins/open_images_format.py
+++ b/datumaro/plugins/open_images_format.py
@@ -12,6 +12,7 @@ import logging as log
 import os
 import os.path as osp
 import re
+import shlex
 import types
 
 from attr import attrs
@@ -25,7 +26,9 @@ from datumaro.components.extractor import (
     LabelCategories,
 )
 from datumaro.components.validator import Severity
-from datumaro.util.image import find_images
+from datumaro.util.image import (
+    DEFAULT_IMAGE_META_FILE_NAME, Image, find_images, load_image_meta_file,
+)
 from datumaro.util.os_util import split_path
 
 # A regex to check whether a subset name can be used as a "normal" path
@@ -122,6 +125,14 @@ class OpenImagesExtractor(Extractor):
 
         self._categories = {}
         self._items = []
+
+        try:
+            self._image_meta = load_image_meta_file(osp.join(
+                path, OpenImagesPath.ANNOTATIONS_DIR,
+                DEFAULT_IMAGE_META_FILE_NAME
+            ))
+        except Exception:
+            self._image_meta = {}
 
         self._load_categories()
         self._load_items()
@@ -223,9 +234,17 @@ class OpenImagesExtractor(Extractor):
                     if _RE_INVALID_SUBSET.fullmatch(subset):
                         raise UnsupportedSubsetNameError(item_id=image_id, subset=subset)
 
+                    image_path = image_paths_by_id.get(image_id)
+
+                    image = None
+                    if image_path is not None:
+                        image = Image(
+                            path=image_path, size=self._image_meta.get(image_id),
+                        )
+
                     items_by_id[image_id] = DatasetItem(
                         id=image_id,
-                        image=image_paths_by_id.get(image_id),
+                        image=image,
                         subset=subset,
                     )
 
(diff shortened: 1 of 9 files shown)


zhiltsov-max:
Also add a test like in https://github.com/openvinotoolkit/datumaro/pull/346

IRDonch:
> Also add a test like in #346

I have examined the implementation in #346, and it would be tricky to implement something like that for OID, because the annotation files often don't have fixed names - just possible suffixes. So it wouldn't be possible to specifically remove the annotation files for a particular subset that was removed. Would it work if I just removed all existing annotation files at the beginning of `apply`?

zhiltsov-max:
Honestly, it is tricky every time. We know from the patch which items were added, removed or modified and which subsets were modified. During writing it is also easy to find out modified subsets that became empty now. Will this information be enough to remove unnecessary files in this format?

IRDonch:
> We know from the patch which items were added, removed or modified and which subsets were modified. During writing it is also easy to find out modified subsets that became empty now. Will this information be enough to remove unnecessary files in this format?

The problem is that we don't know which files which annotations/metadata came from. For the training subset, it could be `train-images-with-labels-with-rotation.csv`, `oidv6-train-images-with-labels-with-rotation.csv`, `image_ids_and_rotation.csv`, or (if the user tries to really exercise the flexibility of the loader) even `foobar-images-with-rotation.csv`. So I don't see a better solution than just nuking the annotation directory.

Although I guess for safety I could do the nuking _after_ the new files are written, instead of before.

IRDonch:
Anyway, I think patching support is a distinct issue from bounding box support, so I think I'm going to work on it separately.

datumaro/plugins/open_images_format.py
```diff
@@ -238,6 +289,53 @@ def _load_labels(self, items_by_id):
                     item.annotations.append(Label(
                         label=label_index, attributes={'score': confidence}))
 
+    def _load_bboxes(self, items_by_id):
+        label_categories = self._categories[AnnotationType.label]
+
+        for bbox_path in self._glob_annotations(
+            '*' + OpenImagesPath.BBOX_DESCRIPTION_FILE_SUFFIX
```
zhiltsov-max, line 296: Probably, we don't need '*' here and can only look for the subsets we loaded earlier. This also should resolve the problem with patching, because each subset file should contain annotations only from that subset (cases, when this is not true can produce a warning, but it should not change reading and patching behaviour).

datumaro/plugins/open_images_format.py
```diff
@@ -238,6 +289,53 @@ def _load_labels(self, items_by_id):
...
+                for bbox_description in bbox_reader:
+                    image_id = bbox_description['ImageID']
+                    item = items_by_id[image_id]
+
+                    label_name = bbox_description['LabelName']
+                    label_index, _ = label_categories.find(label_name)
+                    if label_index is None:
+                        raise UndefinedLabel(
+                            item_id=item.id, subset=item.subset,
+                            label_name=label_name, severity=Severity.error)
```
zhiltsov-max, line 308: I think, the severity argument should be separate from the error class. Probably, I see reporting a dataset problem as an `Extractor` (`Converter`) - base class - method, which could have such severity parameter. Then, it could be handled according to the current error control policy - e.g. a warning could be promoted to an error or ignored. #142

datumaro/plugins/open_images_format.py
```diff
@@ -238,6 +289,53 @@ def _load_labels(self, items_by_id):
...
+
+                    if item.has_image:
+                        height, width = item.image.size
```
zhiltsov-max, line 311: Check if size is available with `item.image.has_size`.

datumaro/plugins/open_images_format.py
```diff
@@ -78,6 +86,30 @@ class OpenImagesPath:
...
+        'IsDepiction',
+        'IsInside',
+    )
+
+    BBOX_BOOLEAN_ATTRIBUTES = (
+        types.SimpleNamespace(datumaro_name='occluded', oid_name='IsOccluded'),
+        types.SimpleNamespace(datumaro_name='truncated', oid_name='IsTruncated'),
+        types.SimpleNamespace(datumaro_name='is_group_of', oid_name='IsGroupOf'),
+        types.SimpleNamespace(datumaro_name='is_depiction', oid_name='IsDepiction'),
+        types.SimpleNamespace(datumaro_name='is_inside', oid_name='IsInside'),
```
zhiltsov-max, line 110: Having `is_` only in some boolean attributes looks a bit misaligned. But let's keep it.

datumaro/plugins/open_images_format.py
```diff
@@ -376,3 +476,42 @@ def _save_subsets(self):
...
+                        elif annotation.type is AnnotationType.bbox:
+                            if bbox_description_writer is None:
+                                bbox_description_writer = annotation_writers.enter_context(
+                                    self._open_csv_annotation(
+                                        subset_name + OpenImagesPath.BBOX_DESCRIPTION_FILE_SUFFIX,
+                                        OpenImagesPath.BBOX_DESCRIPTION_FIELDS))
+                                bbox_description_writer.writeheader()
+
+                            if item.has_image:
+                                image_meta[item.id] = (height, width) = item.image.size
```
zhiltsov-max, line 488: Again, check if size is available.

datumaro/plugins/open_images_format.py
```diff
@@ -376,3 +476,42 @@ def _save_subsets(self):
...
+                            })
+
+        if image_meta:
+            image_meta_file_path = osp.join(
+                self._save_dir, OpenImagesPath.ANNOTATIONS_DIR,
+                DEFAULT_IMAGE_META_FILE_NAME)
+
+            with open(image_meta_file_path, 'w') as image_meta_file:
+                for image_name, (height, width) in image_meta.items():
+                    print(shlex.quote(image_name), height, width, file=image_meta_file)
```
zhiltsov-max, line 517: I have a feeling we need to provide a function to write arbitrary lines in this format in addition to the reading function. This will simplify the client code and abstracts it from direct value encoding.

docs/formats/open_images_user_manual.md
```diff
@@ -15,7 +15,25 @@ on [its website](https://storage.googleapis.com/openimages/web/download.html).
 Datumaro supports versions 4, 5 and 6.
 
 Datumaro currently supports only the human-verified image-level label
-annotations from this dataset.
+annotations and bounding box annotations from this dataset.
+
+One attribute is supported on the labels:
+
+- `score`: the confidence level from 0 to 1. A score of 0 indicates that
```
zhiltsov-max, line 22:
```suggestion
- `score` (float): the confidence level from 0 to 1. A score of 0 indicates that
```

docs/formats/open_images_user_manual.md
```diff
@@ -89,6 +110,36 @@ Open Images dataset directory should have the following structure:
...
+To load bounding box annotations,
+Datumaro needs to know the sizes of the corresponding images.
+By default, it will determine these sizes by loading each image from disk,
+which will make the loading process slow.
+
+You can speed up this process
+by extracting the image size information in advance
+and recording it in an `images.meta` file.
+This file must be located in the `annotations` directory,
+and must contain one line per image, with the following structure:
```
zhiltsov-max, line 124:
```suggestion
If images are not available locally or you want to speed up the process,
you can provide an `images.meta` file.
Such file must be located in the `annotations` directory,
and must contain one line per image, with the following structure:
```

datumaro/plugins/open_images_format.py
IRDonch: Yeah, I'm not a big fan of it either. The main reason I kept the "is" is that "group" and "depiction" are nouns rather than adjectives (and "inside" can be a noun as well), so it would be a bit unnatural to have them as boolean attributes.

datumaro/plugins/open_images_format.py
IRDonch: Done.

datumaro/plugins/open_images_format.py
IRDonch: `has_size` won't help here, because it doesn't attempt to load the size from the image file if it isn't already loaded. Added a check for `image.size is not None` instead.

datumaro/plugins/open_images_format.py
IRDonch: Done.

docs/formats/open_images_user_manual.md
IRDonch: I think just off-handedly mentioning the possibility that the images might not be available doesn't connect well with the other paragraphs, so I rewrite this whole section to incorporate this scenario.

docs/formats/open_images_user_manual.md
IRDonch: OK, updated all attribute descriptions to be consistent with that format.

datumaro/plugins/open_images_format.py
IRDonch: I don't understand - do you want me to change something in the code? I didn't implement the `UndefinedLabel` class, I'm just using it.

datumaro/plugins/open_images_format.py
IRDonch: I don't agree. It seems like it would require a lot of hardcoding (which subsets have which filenames for each annotation type) for no clear benefit. Furthermore, this wouldn't work with custom datasets with custom subset names, since we can't predict what file names will be used for those.

But speaking of patching, I think I have found a good way to implement it. What we need is to remember the set of annotation files that we wrote, and then remove all annotation files that are not in that set, but that match one of the annotation file patterns. This is fairly simple to implement, and it's safe, because if a file matches an annotation file pattern, but we haven't created it, then that means that it must describe a subset that disappeared (or perhaps an annotation type that disappeared).

datumaro/plugins/open_images_format.py
zhiltsov-max: Probably, this error class is not supposed to be used outside dataset validators. But we can leave it.

datumaro/plugins/open_images_format.py
zhiltsov-max: Ok, let's put it to another PR and try to find solution later.

zhiltsov-max (approved):

datumaro/plugins/open_images_format.py
IRDonch: Yep, working on it.

Issue #348, opened 2021-07-12
Dataset patches do not remove empty (e.g. renamed) subsets


zhiltsov-max:
Steps to reproduce:
1. Create a project
2. Import a dataset
3. Rename a subset (e.g. `datum transform -t random_split`)

Depending on the format, the exported dataset will contain annotations from renamed-from and renamed-to subsets. This leads to duplication of annotations in different subsets, which are then found and merged together on importing.

PR #363, opened 2021-07-16
Open Images: remove unneeded annotations when saving in-place

IRDonch:
### Summary

This makes patching work correctly when a subset or annotation type is removed. As a side effect, I also factored out the functionality for lazily creating CSV files into a separate class.

Related to #348.

### How to test

N/A

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/plugins/open_images_format.py
+++ b/datumaro/plugins/open_images_format.py
@@ -382,10 +382,19 @@ class _LazyCsvDictWriter:
 
 
 class _AnnotationWriter:
+    _POSSIBLE_ANNOTATION_FILE_PATTERNS = (
+        # class description files don't need to be listed,
+        # because they are always written
+        OpenImagesPath.FULL_IMAGE_DESCRIPTION_FILE_NAME,
+        *OpenImagesPath.SUBSET_IMAGE_DESCRIPTION_FILE_PATTERNS,
+        '*' + OpenImagesPath.LABEL_DESCRIPTION_FILE_SUFFIX,
+        '*' + OpenImagesPath.BBOX_DESCRIPTION_FILE_SUFFIX,
+        DEFAULT_IMAGE_META_FILE_NAME,
+    )
+
     def __init__(self, root_dir):
         self._annotations_dir = osp.join(root_dir, OpenImagesPath.ANNOTATIONS_DIR)
         self._written_annotations = set()
-        self._potential_annotation_patterns = set()
 
         os.makedirs(self._annotations_dir, exist_ok=True)
 
@@ -408,22 +417,26 @@ class _AnnotationWriter:
         return _LazyCsvDictWriter(
             lambda: self.open_csv(file_name, field_names))
 
-    def add_potential_annotation_pattern(self, pattern):
-        self._potential_annotation_patterns.add(pattern)
-
     def remove_unwritten(self):
         for file_name in os.listdir(self._annotations_dir):
             if file_name not in self._written_annotations and any(
                 fnmatch.fnmatch(file_name, pattern)
-                for pattern in self._potential_annotation_patterns
+                for pattern in self._POSSIBLE_ANNOTATION_FILE_PATTERNS
             ):
-                os.unlink(file_name)
+                os.unlink(osp.join(self._annotations_dir, file_name))
 
 class OpenImagesConverter(Converter):
     DEFAULT_IMAGE_EXT = '.jpg'
 
     def apply(self):
         self._save(_AnnotationWriter(self._save_dir))
+
+    @classmethod
+    def patch(cls, dataset, patch, save_dir, **options):
+        converter = cls(dataset, save_dir, **options)
+        annotation_writer = _AnnotationWriter(save_dir)
+        converter._save(annotation_writer)
+        annotation_writer.remove_unwritten()
 
     def _save(self, annotation_writer):
         self._save_categories(annotation_writer)
--- a/datumaro/plugins/open_images_format.py
+++ b/datumaro/plugins/open_images_format.py
@@ -349,38 +349,100 @@ class OpenImagesImporter(Importer):
 
         return []
 
+
+class _LazyCsvDictWriter:
+    """
+    For annotation files, we only learn that the file is required after
+    we find at least one occurrence of the corresponding annotation type.
+    However, it's convenient to create the writer ahead of time, so that
+    it can be used in a `with` statement.
+
+    This class behaves like a csv.DictWriter, but it only creates the file
+    once the first row is written.
+    """
+
+    def __init__(self, writer_manager_factory):
+        self._writer_manager_factory = writer_manager_factory
+        self._writer_manager = None
+        self._writer = None
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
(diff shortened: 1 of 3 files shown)


datumaro/plugins/open_images_format.py
```diff
@@ -349,38 +349,113 @@ def find_sources(cls, path):
...
+    DEFAULT_IMAGE_EXT = '.jpg'
 
-    def _save_categories(self):
-        with self._open_csv_annotation(
+    def apply(self):
+        self._save(_AnnotationWriter(self._save_dir))
+
+    @classmethod
+    def patch(cls, dataset, patch, save_dir, **options):
+        converter = cls(dataset, save_dir, **options)
```
zhiltsov-max, line 436: The purpose of patching is not only updating, but also avoiding to read the whole input dataset (which is supposed to be loaded lazily), if it is possible. Passing input dataset can be changed to passing only modified subsets in most cases. Consider doing it here as well.

tests/test_open_images_format.py
```diff
@@ -110,6 +111,42 @@ def test_can_save_and_load_image_with_arbitrary_extension(self):
 
             compare_datasets(self, dataset, parsed_dataset, require_images=True)
 
+    @mark_requirement(Requirements.DATUM_274)
+    def test_inplace_save_writes_only_updated_data(self):
+        dataset = Dataset.from_iterable([
+            DatasetItem('a', subset='s', image=np.ones((2, 1, 3)),
```
zhiltsov-max, line 117: Consider having an unchanged subset and unchanged items in modified subsets.

datumaro/plugins/open_images_format.py
```diff
@@ -349,38 +349,113 @@ def find_sources(cls, path):
...
+        annotation_writer = _AnnotationWriter(save_dir)
+        converter._save(annotation_writer)
+        annotation_writer.remove_unwritten()
```
zhiltsov-max, line 439: Look at other implementations. Probably, I've found a more or less generic heuristic for removal of subsets - [here](https://github.com/openvinotoolkit/datumaro/pull/365/files#diff-13eb8f857f20c388f388dd802efed6d9f85486bc3294cb55b930cf7200766544R310-R315).

IRDonch: This approach only works if you can accurately predict the file name from which the deleted subsets' annotations were loaded. Which, for this format, we can't do.

tests/test_open_images_format.py
IRDonch: Alright, implemented (I also gave the subsets better names).

datumaro/plugins/open_images_format.py
IRDonch: Unfortunately, that doesn't seem to be possible, because building the updated `images.meta` file requires all subsets. Should I switch to using per-subset meta files?

datumaro/plugins/open_images_format.py
zhiltsov-max: Ok, then it looks like the best what we can do is remove all the .csv and deleted images by their names. Then reexport the whole dataset.

datumaro/plugins/open_images_format.py
zhiltsov-max: Answered below

datumaro/plugins/open_images_format.py
IRDonch: This I can do, but wouldn't it be safer to only delete the old annotation files _after_ we're done writing the new ones (which is what I'm doing in the current version)? That way there won't be data loss if, for example, an unexpected exception occurs after the old files are deleted.

> and deleted images by their names

There's actually an unexpected problem with that. If an item is deleted, then we don't know what extension its image had, so we can't actually delete it. As far as I can see, this problem affects your implementations, as well.

zhiltsov-max: > This I can do, but wouldn't it be safer to only delete the old annotation files after we're done writing the new ones (which is what I'm doing in the current version)? That way there won't be data loss if, for example, an unexpected exception occurs after the old files are deleted.

The only way of protection from such problems I can come up with is writing to new files, then renaming to the old names. Because once you opened a file with `open(..., 'w')`, the old data is lost.

> we don't know what extension its image had, so we can't actually delete it. As far as I can see, this problem affects your implementations, as well.

Yes, I know about this. A way to fix this could be using of image file search by name, and it was implemented in some format, but it looks like overkill right now. Maybe, it could be a next update for the patching functionality.

IRDonch: > The only way of protection from such problems I can come up with is writing to new files, then renaming to the old names.

Okay. I could implement this quite easily (since I had already centralized the writing of annotations), so I did.

> Yes, I know about this. A way to fix this could be using of image file search by name, and it was implemented in some format, but it looks like overkill right now. Maybe, it could be a next update for the patching functionality.

Okay. I implemented your solution for deleting images for now.

datumaro/plugins/open_images_format.py
```diff
@@ -398,12 +398,22 @@ def __init__(self, root_dir):
...
-        f = open(
-            osp.join(self._annotations_dir, file_name),
-            'w', encoding='utf-8', newline=newline)
-        return f
+
+        file_path = osp.join(self._annotations_dir, file_name)
+
+        # Write to a temporary file first, to avoid data loss if we're patching
+        # an existing dataset and the process is interrupted.
+        temp_file_path = file_path + '.tmp'
```
zhiltsov-max, line 410: I wouldn't say it is a problem, but using https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp would be more pragmatic. Consider using id with `datumaro.utils.error_cleanup`.

zhiltsov-max (approved):

datumaro/plugins/open_images_format.py
IRDonch: > I wouldn't say it is a problem, but using https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp would be more pragmatic.

I've considered it, but `mkstemp` (and its various derivatives) has one quality that makes it unsuitable for this use case: it always creates the temporary file with `600` permissions, which are not necessarily the same permissions as one would get by creating the file normally. And correcting the permissions after the fact is surprisingly involved (the default permissions depend on the process umask, and you can't actually fetch the current umask without changing it, and changing it would affect other threads...).

Furthermore, the only benefit of using it would be avoiding filename conflicts, but the only way such conflicts would occur would be if either:

a. two datasets were being written to the same directory at the same time, which I'm pretty sure would break horribly anyway, or
b. a `.tmp` file already existed in the `annotations` directory and contained valuable data, which seems unlikely (there are certainly no `.tmp` files in the original Open Images dataset).

> Consider using it with datumaro.utils.error_cleanup.

That doesn't seem to exist. Perhaps it's called something else?

zhiltsov-max: > That doesn't seem to exist. Perhaps it's called something else?

I meant this: https://github.com/openvinotoolkit/datumaro/blob/develop/datumaro/util/__init__.py#L192

An example: https://github.com/openvinotoolkit/datumaro/blob/develop/datumaro/cli/contexts/project/__init__.py#L551

PR #373, opened 2021-07-21
Enable the Pylint undefined-variable checker

IRDonch:
### Summary

Undefined variables are a serious problem, and easy to check for, so do that and fix all existing occurrences.

~~WRT the `error_rollback` changes: there was no error there, but Pylint would complain, because it didn't know where the `on_error` variable was coming from. Defining `on_error` explicitly is the workaround for that. While it's less elegant, it makes the code more transparent, both for human users and static analyzers (such as Pylint itself, or VS Code's syntax checker). Also, IMO, the benefits of the checker far outweigh the drawback of having to implement this workaround.~~

### Checklist

### License

```python
#
```

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -86,6 +86,7 @@ enable=
     E0301, # non-iterator-returned
     E0302, # unexpected-special-method-signature
     E0601, # used-before-assignment
+    E0602, # undefined-variable
     E0603, # undefined-all-variable
     E0604, # invalid-all-object
     E0701, # bad-except-order
--- a/datumaro/cli/commands/merge.py
+++ b/datumaro/cli/commands/merge.py
@@ -6,6 +6,7 @@ from collections import OrderedDict
 import argparse
 import json
 import logging as log
+import os
 import os.path as osp
 
 from datumaro.components.errors import MergeError, QualityError
--- a/datumaro/plugins/accuracy_checker_plugin/details/representation.py
+++ b/datumaro/plugins/accuracy_checker_plugin/details/representation.py
@@ -34,7 +34,7 @@ def import_prediction(pred):
     elif isinstance(pred, (ac.DetectionPrediction, ac.ActionDetectionPrediction)):
         return (dm.Bbox(x0, y0, x1 - x0, y1 - y0, int(label_id),
                 attributes={'score': float(score)})
-            for label, score, x0, y0, x1, y1 in zip(pred.labels, pred.scores,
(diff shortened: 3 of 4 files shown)


datumaro/cli/contexts/model.py
```diff
@@ -46,6 +46,8 @@ def build_add_parser(parser_ctor=argparse.ArgumentParser):
 
     return parser
 
+on_error = None
```
zhiltsov-max, line 49: Having to declare a variable negates a half of the usability of this solution from my point of view. Maybe we can includea "default" name in `good-names` of `.pylintrc`? Probably, we could also rename `implicit` to `implicit_var` to make it more clear.

datumaro/util/__init__.py
```diff
@@ -196,16 +196,12 @@ def wrapped_func(*args, **kwargs):
             if implicit:
                 fglobals = func.__globals__
 
-                has_arg = arg_name in fglobals
-                old_val = fglobals.get(arg_name)
+                old_val = fglobals[arg_name]
```
zhiltsov-max, line 199: Now it fails with a cryptic error if a variable is not defined. If we accept this solution, I suggest to add docs or raise a more clear error with a possible solution.

datumaro/cli/contexts/model.py
IRDonch: How about we just dispense with the implicit variables and modify the semantics of `error_rollback` so that it can be used as follows:

```python
@error_rollback
def foobar():
    Rollback.current.add(my_rollback)
```

?

zhiltsov-max: I have no objections about such solution. But then we need either some class property (which can be done via metaclasses somehow), or we introduce a static function - with stack lookup (to obtain a function identifier). We need to have a separate rollback stacks for inner functions. It can be done even with a single global list (stack) of exit callbacks via some markers. Then, if we have an unhandled error, we execute functions (in the reverse order to the addition) until we meet a marker. Maybe I miss some elegant solution?

IRDonch: I don't see why such complexity is needed. We can just set `Rollback.current` before the underlying function is called, and restore it back to the previous value once it's done. It's the same thing we do now, just with a class member variable instead of a module-level variable.

zhiltsov-max: Looks good. It is similar to what I described in the second half.

IRDonch:
The `on_error` refactoring was implemented in #380. There are only non-controversial fixes left here. 🙂

zhiltsov-max (approved):

PR #380, opened 2021-07-23
Redesign the implicit mode of error_rollback

IRDonch:
### Summary
Instead of adding a global `on_error` variable in the module of the calling function, use a thread-local variable in the `Rollback` class. This has the following advantages:

* Linters and IDEs no longer complain about undefined variables.
* The `Rollback.current` accessor is type-annotated, allowing IDEs to provide autocompletion for methods.
* Using a thread-local variable eliminates potential conflicts if Datumaro is used in multiple threads.
* The `implicit` parameter is no longer required, since we can default to implicit mode if the variable name is unspecified.

Inspired by the discussion in #373.

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/cli/contexts/model.py
+++ b/datumaro/cli/contexts/model.py
@@ -9,7 +9,7 @@ import os.path as osp
 import shutil
 
 from datumaro.components.project import Environment
-from datumaro.util import error_rollback
+from datumaro.util import error_rollback, on_error_do
 
 from ..util import CliException, MultilineFormatter, add_subparser
 from ..util.project import (
@@ -46,7 +46,7 @@ def build_add_parser(parser_ctor=argparse.ArgumentParser):
 
     return parser
 
-@error_rollback('on_error', implicit=True)
+@error_rollback
 def add_command(args):
     project = load_project(args.project_dir)
 
@@ -73,7 +73,7 @@ def add_command(args):
         model_dir = osp.join(project.config.project_dir,
             project.local_model_dir(args.name))
         os.makedirs(model_dir, exist_ok=False)
-        on_error.do(shutil.rmtree, model_dir, ignore_errors=True)
+        on_error_do(shutil.rmtree, model_dir, ignore_errors=True)
 
         try:
             cli_plugin.copy_model(model_dir, model_args)
--- a/datumaro/util/__init__.py
+++ b/datumaro/util/__init__.py
@@ -2,11 +2,12 @@
 #
 # SPDX-License-Identifier: MIT
 
-from contextlib import ExitStack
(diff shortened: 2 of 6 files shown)


datumaro/cli/contexts/model.py
```diff
@@ -73,7 +73,7 @@ def add_command(args):
         model_dir = osp.join(project.config.project_dir,
             project.local_model_dir(args.name))
         os.makedirs(model_dir, exist_ok=False)
-        on_error.do(shutil.rmtree, model_dir, ignore_errors=True)
+        Rollback.current().do(shutil.rmtree, model_dir, ignore_errors=True)
```
zhiltsov-max, line 76: Looks good. Do you have any ideas how can we make it more terse? Probably, I'm OK with the current solution, but I'd prefer calling a single function. Maybe, we could change class name to lower case and join `current().do()` to a single function to make it clear

IRDonch: What if I just add something like this to `datumaro.util`:

```python
def on_error_do(...):
    Rollback.current().add(...)
```

Then the usage can be as concise as possible.

> Apparently, in 3.7 we could create a module with an attribute: https://stackoverflow.com/a/57263518

Not a fan of this approach. If you implement it like that, things like autocomplete or go-to-definition won't work anymore, and they are half the reason I'm even making this change in the first place.

zhiltsov-max: > `def on_error_do(...):`

Maybe. It's a bit narrower than other ways (in meaning of API accessibility), but it covers all the actual use cases.

IRDonch: OK, I implemented it.

zhiltsov-max (approved):

datumaro/util/__init__.py
```diff
@@ -186,24 +189,31 @@ def __exit__(self, type=None, value=None, \
...
+    def as_current(self):
+        previous = getattr(self._thread_locals, 'current', None)
+        self._thread_locals.current = self
+        try:
+            yield
+        finally:
+            self._thread_locals.current = previous
+
+# shorthand for common cases
+def on_error_do(callback, *args, ignore_errors=False):
```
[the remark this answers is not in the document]
IRDonch: I don't think functions that take keyword arguments of their own should take generic kwargs to forward elsewhere, because then you might provoke bugs where arguments that are meant to be forwarded are used directly by the function instead. (And FWIW, I don't think `Rollback.add` should do it either.)

Also, more practically, there aren't any current uses of `on_error` that use keyword argument forwarding, so there was no reason for me to add it.

Issue #313, opened 2021-06-24
Add global constraints on dataset item IDs and subset names


IRDonch:
There are some constraints that should be placed on all item IDs and subset names, to ensure that they can be safely and unambiguously printed to the terminal or used in text-based formats. Specifically:

* they should not be empty strings;
* they should not contain control characters.

PR #388, opened 2021-07-27
Open Images: add segmentation mask support

IRDonch:
### Summary

Yet another step to implementing #274.

### Checklist

### License

```python
#
```


datumaro/plugins/open_images_format.py
```diff
@@ -15,29 +16,32 @@
 import types
 
 from attr import attrs
+import cv2
+import numpy as np
 
+from datumaro.cli.util import make_file_name
```
zhiltsov-max, line 22: I don't think importing from `cli` is good here. Probably, this function is better be moved to the core utils (and it is done in the versioning PR).

datumaro/plugins/open_images_format.py
```diff
@@ -50,9 +54,26 @@ class UnsupportedSubsetNameError(DatasetError):
...
+@attrs(auto_attribs=True)
+class UnsupportedBoxIdError(DatasetError):
+    box_id: str
+
+    def __str__(self):
+        return "Item %s has a mask with an unsupported box ID %r." % (
+            self.item_id, self.box_id)
+
+@attrs(auto_attribs=True)
+class UnsupportedMaskPathError(DatasetError):
```
zhiltsov-max, line 66: Consider moving these errors to `errors.py`.

datumaro/plugins/open_images_format.py
```diff
@@ -337,6 +372,89 @@ def _load_bboxes(self, items_by_id):
...
+    def _load_masks(self, items_by_id):
+        label_categories = self._categories[AnnotationType.label]
+
+        for mask_path in self._glob_annotations(
+            '*' + OpenImagesPath.MASK_DESCRIPTION_FILE_SUFFIX
+        ):
+            with self._open_csv_annotation(mask_path) as mask_reader:
+                for mask_description in mask_reader:
+                    mask_path = mask_description['MaskPath']
+                    if _RE_INVALID_PATH_COMPONENT.fullmatch(mask_path):
```
zhiltsov-max, line 384: How about `osp.abspath().startswith(dataset_root + osp.sep)`?

datumaro/plugins/open_images_format.py
```diff
@@ -591,6 +723,58 @@ def _save_subsets(self, annotation_writer):
...
+
+                            if box_id_str:
+                                if _RE_INVALID_PATH_COMPONENT.fullmatch(box_id_str):
+                                    raise UnsupportedBoxIdError(item_id=item.id, box_id=box_id_str)
+                            else:
+                                # find a box ID that isn't used in any other annotations
+                                while True:
+                                    box_id_str = format(next_box_id, "08x")
+                                    next_box_id += 1
+                                    if box_id_str not in existing_box_ids:
```
zhiltsov-max, line 739: Are the ids unique for the whole dataset?

datumaro/plugins/open_images_format.py
```diff
@@ -591,6 +723,58 @@ def _save_subsets(self, annotation_writer):
...
+                            mask_file_name = '%s_%s_%s.png' % (
+                                item.id, make_file_name(label_name), box_id_str,
+                            )
+
+                            box = annotation.attributes.get('box')
+                            box_coords = {}
+
+                            if box is not None:
+                                if item.has_image and item.image.size is not None:
+                                    image_meta[item.id] = (height, width) = item.image.size
```
zhiltsov-max, line 752: I discourage using multiple assignment pattern.

datumaro/plugins/open_images_format.py
```diff
@@ -337,6 +372,89 @@ def _load_bboxes(self, items_by_id):
...
+                        continue
+
+                    height, width = item.image.size
+
+                    # The original OID has box coordinates for all masks, but
+                    # a dataset converted from another dataset might not.
+                    if all(mask_description[f] for f in [
+                        'BoxXMin' ,'BoxXMax', 'BoxYMin', 'BoxYMax',
+                    ]):
+                        box = attributes['box'] = {}
```
zhiltsov-max, line 427: Is there any reason to include it as an attribute? Can these values be distinct from what is in the mask? If possible, let's stick to simple type attributes.
If you want to have a bbox attached to a mask, the current approach is to have it as another annotation in the list with the same group as the corresponding mask and no label. It complicates processing a bit, though, check COCO converter or https://github.com/openvinotoolkit/datumaro/blob/develop/datumaro/util/annotation_util.py#L15. I'd vote for exclusion of this information completely, because it can be computed without great overhead, and masks have such method.

zhiltsov-max (changes requested):

datumaro/plugins/open_images_format.py
IRDonch: > Is there any reason to include it as an attribute? Can these values be distinct from what is in the mask?

To quote the Open Images website:

"Note that this is not the bounding box of the mask, but the starting box from which the mask was annotated. These coordinates can be used to relate the mask data with the boxes data."

> If you want to have a bbox attached to a mask, the current approach is to have it as another annotation in the list with the same group as the corresponding mask and no label.

Okay, that seems quite doable. I can assign a distinct group to every box annotation and then find the matching box annotation for every mask annotation when loading them.

I could, of course, just not load this information at all, but according to the dataset description, it's _not_ something just can be computed, so it seems useful to have.

datumaro/plugins/open_images_format.py
IRDonch: Why?

datumaro/plugins/open_images_format.py
IRDonch: The box IDs are only unique per image, which is why `existing_box_ids` is recreated for each image.

I guess technically they could be unique per image _and_ label, but there are no duplicate (image, box_id) pairs in the whole dataset, so it's not a particularly relevant distinction.

datumaro/plugins/open_images_format.py
IRDonch: That's too forgiving. Something like `../../masks/validation/foobar.png` can pass a check like that, but it should not be accepted as a valid path.

datumaro/plugins/open_images_format.py
IRDonch: > Consider moving these errors to components/errors.py.

I think that would be done best in the scope of #313.

> BTW, as of 21.1.0 of attrs, auto_attribs can be replaced with define. and it looks like a thing I'd want to start using in the project.

I agree, but _that_ would be done best in the scope of #310. 🙂

datumaro/plugins/open_images_format.py
IRDonch: Yeah, fair enough. I'll probably steal your move patch (unless you want to submit it as a separate PR).

datumaro/plugins/open_images_format.py
zhiltsov-max: Basically, for the same reason we don't write lines like `do_one(); do_another(); x = foo(); y = bar();`. This make the code clearer and reduces load on the reader.

IRDonch: It's not quite the same thing, because it reduces duplication. But eh, I can change it if you want.

datumaro/plugins/open_images_format.py
IRDonch: > Okay, that seems quite doable. I can assign a distinct group to every box annotation and then find the matching box annotation for every mask annotation when loading them.

Implemented this.

datumaro/plugins/open_images_format.py
IRDonch: Removed.

datumaro/plugins/open_images_format.py
IRDonch: Patch stolen. 🙂

datumaro/plugins/open_images_format.py
```diff
@@ -330,13 +377,114 @@ def _load_bboxes(self, items_by_id):
...
+                    # The box IDs are rather useless, because the _box_ annotations
+                    # don't include them, so they cannot be used to match masks to boxes.
+                    # However, it is still desirable to record them, because they are
+                    # included in the mask file names, so in order to save each mask to the
+                    # file it was loaded from when saving in-places, we need to know
+                    # the original box ID.
+                    box_id = mask_description['BoxID']
+                    if _RE_INVALID_PATH_COMPONENT.fullmatch(box_id):
+                        raise UnsupportedBoxIdError(item_id=item.id, box_id=box_id)
+                    attributes['box_id'] = box_id
```
zhiltsov-max, line 436: Maybe these ids should be mask ids?

tests/test_open_images_format.py
```diff
@@ -34,12 +34,18 @@ def test_can_save_and_load(self):
...
+                        Bbox(label=1, x=2, y=3, w=6, h=1, group=1, attributes={
                             'score': 0.7,
                             'occluded': True, 'truncated': False,
                             'is_group_of': True, 'is_depiction': False,
                             'is_inside': False,
                         }),
+                        Mask(label=0, image=np.eye(8)),
+                        Mask(label=1, image=np.ones((8, 8)), group=1, attributes={
+                            'box_id': '00000000',
+                            'box': {'x': 2, 'y': 3, 'w': 6, 'h': 1},
```
zhiltsov-max, line 46:
```suggestion
```

datumaro/plugins/open_images_format.py
IRDonch: They're called box IDs in the dataset; I don't think we need to rename them. 🤷

tests/test_open_images_format.py
IRDonch: Fixed, thanks.

zhiltsov-max (approved):

PR #394, opened 2021-07-30
Add a generic implementation for patching

### LINKED ISSUE #348: Dataset patches do not remove empty (e.g. renamed) subsets

zhiltsov-max:
### Summary

Closes #348

- Added a generic implementation of dataset patching

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/components/converter.py
+++ b/datumaro/components/converter.py
@@ -38,11 +38,17 @@ class Converter(CliPlugin):
     @classmethod
     @error_rollback
     def patch(cls, dataset, patch, save_dir, **options):
-        # Is not any better in performance than just writing a dataset,
-        # but in case of patching (i.e. writing to the previous location),
-        # it allows to avoid many problems with removing and replacing
-        # existing files. Surely, this approach also has problems with
-        # removal of the given directory.
+        # This solution is not any better in performance than just
+        # writing a dataset, but in case of patching (i.e. writing
+        # to the previous location), it allows to avoid many problems
+        # with removing and replacing existing files. Surely, this
+        # approach also has problems with removal of the given directory.
+        # Problems can occur if we can't remove the directory,
+        # or want to reuse the given directory. It can happen if it
+        # is mounted or (sym-)linked.
+        # Probably, a better solution could be to wipe directory
+        # contents and write new data there. Note that directly doing this
+        # also doesn't work, because images may be needed for writing.
         tmpdir = save_dir + '.tmp' + uuid.uuid4().hex
         os.makedirs(tmpdir, exist_ok=False)
         on_error_do(shutil.rmtree, tmpdir, ignore_errors=True)
--- a/datumaro/components/converter.py
+++ b/datumaro/components/converter.py
@@ -7,10 +7,12 @@ import logging as log
 import os
 import os.path as osp
 import shutil
+import uuid
 
 from datumaro.components.cli_plugin import CliPlugin
 from datumaro.components.dataset import DatasetPatch
 from datumaro.components.extractor import DatasetItem
+from datumaro.util import error_rollback, on_error_do
 from datumaro.util.image import Image
 
 
@@ -34,8 +36,22 @@ class Converter(CliPlugin):
         return converter.apply()
 
     @classmethod
+    @error_rollback
     def patch(cls, dataset, patch, save_dir, **options):
-        return cls.convert(dataset, save_dir, **options)
+        # Is not any better in performance than just writing a dataset,
+        # but in case of patching (i.e. writing to the previous location),
+        # it allows to avoid many problems with removing and replacing
+        # existing files. Surely, this approach also has problems with
+        # removal of the given directory.
+        tmpdir = save_dir + '.tmp' + uuid.uuid4().hex
+        os.makedirs(tmpdir, exist_ok=False)
+        on_error_do(shutil.rmtree, tmpdir, ignore_errors=True)
+
+        retval = cls.convert(dataset, tmpdir, **options)
+        shutil.rmtree(save_dir, ignore_errors=True)
+        os.replace(tmpdir, save_dir)
+
+        return retval
 
     def apply(self):
         raise NotImplementedError("Should be implemented in a subclass")
--- a/datumaro/components/converter.py
+++ b/datumaro/components/converter.py
@@ -2,12 +2,12 @@
 #
 # SPDX-License-Identifier: MIT
 
+from tempfile import mkdtemp
 from typing import Union
 import logging as log
 import os
 import os.path as osp
 import shutil
-import uuid
 
 from datumaro.components.cli_plugin import CliPlugin
 from datumaro.components.dataset import DatasetPatch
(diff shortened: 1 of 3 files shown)


datumaro/components/converter.py
```diff
@@ -34,8 +36,22 @@ def convert(cls, extractor, save_dir, **options):
...
 
     @classmethod
+    @error_rollback
     def patch(cls, dataset, patch, save_dir, **options):
-        return cls.convert(dataset, save_dir, **options)
+        # Is not any better in performance than just writing a dataset,
+        # but in case of patching (i.e. writing to the previous location),
+        # it allows to avoid many problems with removing and replacing
+        # existing files. Surely, this approach also has problems with
+        # removal of the given directory.
```
IRDonch, line 45: I think this should be elaborated on. What are the problems with removal of the given directory?

datumaro/components/converter.py
```diff
@@ -34,8 +36,22 @@ def convert(cls, extractor, save_dir, **options):
...
+        tmpdir = save_dir + '.tmp' + uuid.uuid4().hex
```
IRDonch, line 46: Using `uuid` here seems like an unnecessary indirection (and it also loses a few bits of entropy, since a UUID has some fixed bits). You can just generate the required number of random bytes directly.

datumaro/components/converter.py
```diff
@@ -34,8 +36,22 @@ def convert(cls, extractor, save_dir, **options):
...
+        os.makedirs(tmpdir, exist_ok=False)
```
IRDonch, line 47:
```suggestion
        os.mkdir(tmpdir)
```

This is a rare occasion on which creating parent directories is unnecessary.

datumaro/components/converter.py
```diff
@@ -34,8 +36,22 @@ def convert(cls, extractor, save_dir, **options):
...
+        on_error_do(shutil.rmtree, tmpdir, ignore_errors=True)
+
+        retval = cls.convert(dataset, tmpdir, **options)
+        shutil.rmtree(save_dir, ignore_errors=True)
```
IRDonch, line 51: Ignoring errors here seems like a bad idea. If any errors occur, the following `os.replace` call will fail anyway, so you might as well fail early.

datumaro/components/converter.py
```diff
@@ -34,8 +36,22 @@ def convert(cls, extractor, save_dir, **options):
         return converter.apply()
 
     @classmethod
+    @error_rollback
     def patch(cls, dataset, patch, save_dir, **options):
-        return cls.convert(dataset, save_dir, **options)
+        # Is not any better in performance than just writing a dataset,
+        # but in case of patching (i.e. writing to the previous location),
+        # it allows to avoid many problems with removing and replacing
+        # existing files. Surely, this approach also has problems with
```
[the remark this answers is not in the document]
zhiltsov-max: > For example, what if the user tries to save one dataset into the directory of another dataset?

I'd say it's not our business. But, for completeness sake, there is always a check for empty directory and a corresponding option for overriding in the CLI.

datumaro/components/converter.py
zhiltsov-max: I see problems can occur if we can't remove a directory, or want to reuse the given directory. For example, if it is mounted or (sym-)linked. Ideally, I think, it would be better to wipe directory contents and write new data there.

datumaro/components/converter.py
zhiltsov-max: We need any unused name here. Technically, what `TemporaryDirectory` does, but with normal rights. Maybe we need to copy them from the given directory? And there is also possible the [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) problem.

datumaro/components/converter.py
zhiltsov-max: In normal cases yes, but if user removed something... I don't think this question really should be considered here.

datumaro/components/converter.py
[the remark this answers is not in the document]
IRDonch: > I'd say it's not our business.

Why not? Don't you think it's inconsistent that saving a dataset in its original directory wipes that directory, but saving it in another directory doesn't?

datumaro/components/converter.py
IRDonch: Okay, so reflect that in your comment.

datumaro/components/converter.py
IRDonch: I guess that's a fair point.

datumaro/components/converter.py
IRDonch: All of what you said makes sense, but it doesn't really address my point, which is that you could've used `os.random(16)` to get more entropy for the same file name length (and without introducing an extra import).

> Technically, what TemporaryDirectory does, but with normal rights. Maybe we need to copy them from the given directory?

It might be a good idea to use `mkdtemp` and copy the permissions from the original directory. That way, you won't need to deal with the randomness by yourself.

> And there is also possible the TOCTOU problem.

TOCTOU can only be a problem if there's a check, and I don't think there is one here. 🙂

datumaro/components/converter.py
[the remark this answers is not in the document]
zhiltsov-max: Potentially, there can be extra files in the dataset directory (like readmes, descriptions and stuff), so I don't think removing the directory in patching is also a good solution. But it at least something working and doable.

datumaro/components/converter.py
zhiltsov-max: I tried to re implement this with `mkdtemp`.

datumaro/components/converter.py
zhiltsov-max: Changed

datumaro/components/converter.py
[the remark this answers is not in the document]
IRDonch: I realize that, but I think the behavior should be consistent between scenarios. I think it would be surprising for the user to see that some write operations leave the existing files as-is, while some wipe them out.

Perhaps the right solution would be for the user to opt-in to the wipe behavior. You could add a command-line switch, and in the absence of that switch refuse to patch unless the dataset has a specialized patch implementation.

datumaro/components/converter.py
```diff
@@ -34,8 +36,31 @@ def convert(cls, extractor, save_dir, **options):
...
+        # Probably, a better solution could be to wipe directory
+        # contents and write new data there. Note that directly doing this
+        # also doesn't work, because images may be needed for writing.
+        tmpdir = mkdtemp(dir=osp.dirname(save_dir),
+            prefix=osp.basename(save_dir), suffix='.tmp')
+        on_error_do(shutil.rmtree, tmpdir, ignore_errors=True)
+
+        retval = cls.convert(dataset, tmpdir, **options)
+
+        if osp.isdir(save_dir):
```
IRDonch, line 58: I don't think this check should be done here. If it fails, then the directory will be left with `mkdtemp`'s default permissions (700), which would be surprising. Instead, you should do it at the start of the method, and if it fails, then you can just do the conversion normally (without creating a temporary directory).

datumaro/components/converter.py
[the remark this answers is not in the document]
zhiltsov-max: Adding an extra mystical cli parameter doesn't look right for me. I propose to leave this question till we face this problem IRL.

PR #401, opened 2021-08-05
Add patch command

zhiltsov-max:
### Summary

- Added `datum patch` command, which allows to update a dataset from another dataset
- Added command docs
- Added `ProjectLabels` transform to align dataset labels
- Added label existence check in `LabelCategories` by name
- Added an error message on attempt to re-download a source with no hash
- Improved COCO export docs
- Fixed return value type in `Dataset.is_modified`
- Fixed missing error on saving project readonly project datasets inplace by path
- Fixed class inheritance for `DatasetMergeError` (`item_id` should not be mandatory)
- Fixed sensitive args in `datum diff`
- Fixed label remapping for secondary categories in `RemapLabels`

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/components/errors.py
+++ b/datumaro/components/errors.py
@@ -210,9 +210,13 @@ class WrongGroupError(DatasetQualityError):
 class DatasetMergeError(DatasetError):
     sources = attrib(converter=set, factory=set, kw_only=True)
 
-    def __init__(self, msg=None, *, sources=None):
+    def _my__init__(self, msg=None, *, sources=None):
         super().__init__(msg)
         self.__attrs_init__(sources=sources or set())
+
+# Pylint will raise false positive warnings for derived classes,
+# when __init__ is defined directly
+setattr(DatasetMergeError, '__init__', DatasetMergeError._my__init__)
 
 @attrs
 class MismatchingImageInfoError(DatasetMergeError):
--- a/datumaro/components/errors.py
+++ b/datumaro/components/errors.py
@@ -164,10 +164,8 @@ class NoMatchingFormatsError(DatasetImportError):
         return "Failed to detect dataset format automatically: " \
             "no matching formats found"
 
-
-@attrs
 class DatasetError(DatumaroError):
-    item_id = attrib()
+    pass
 
 class CategoriesRedefinedError(DatasetError):
     def __str__(self):
@@ -175,16 +173,18 @@ class CategoriesRedefinedError(DatasetError):
 
 @attrs
 class RepeatedItemError(DatasetError):
+    item_id = attrib()
+
     def __str__(self):
         return f"Item {self.item_id} is repeated in the source sequence."
 
 
-@attrs
 class DatasetQualityError(DatasetError):
     pass
 
 @attrs
 class AnnotationsTooCloseError(DatasetQualityError):
+    item_id = attrib()
     a = attrib()
     b = attrib()
     distance = attrib()
@@ -195,6 +195,7 @@ class AnnotationsTooCloseError(DatasetQualityError):
 
 @attrs
 class WrongGroupError(DatasetQualityError):
+    item_id = attrib()
     found = attrib(converter=set)
     expected = attrib(converter=set)
     group = attrib(converter=list)
@@ -205,12 +206,17 @@ class WrongGroupError(DatasetQualityError):
             (self.item_id, self.found, self.expected, self.group)
 
 
-@attrs
+@attrs(init=False)
 class DatasetMergeError(DatasetError):
     sources = attrib(converter=set, factory=set, kw_only=True)
 
+    def __init__(self, msg=None, *, sources=None):
+        super().__init__(msg)
+        self.__attrs_init__(sources=sources or set())
+
 @attrs
 class MismatchingImageInfoError(DatasetMergeError):
+    item_id = attrib()
     a = attrib()
     b = attrib()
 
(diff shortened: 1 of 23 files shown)


datumaro/plugins/coco_format/converter.py
```diff
@@ -520,13 +520,27 @@ def _split_tasks_string(s):
...
             - '{sm.polygons.name}': save polygons,|n
             |s|smerge and convert masks, prefer polygons|n
             - '{sm.mask.name}': save masks,|n
-            |s|smerge and convert polygons, prefer masks
+            |s|smerge and convert polygons, prefer masks|n
+            |n
+            The '--reindex' option allows to control if the images and
+            annotations must be given new indices. It can be useful, when
+            you want to preserve the original indices in the produced dataset.
+            Consider enabling this option when converting from other formats.|n
```
IRDonch, line 534: The advice to enable this option seems strange when you've changed the default to true. Maybe change it to "keeping this option enabled"?

datumaro/plugins/transforms.py
```diff
@@ -564,6 +567,66 @@ def transform_item(self, item):
...
+
+        src_label_cat = self._extractor.categories().get(AnnotationType.label)
+        self._make_label_id_map(src_label_cat, dst_labels)
+
+        src_mask_cat = self._extractor.categories().get(AnnotationType.mask)
+        if src_mask_cat is not None:
+            assert src_label_cat is not None
+            dst_mask_cat = MaskCategories(attributes=src_mask_cat.attributes)
+            dst_mask_cat.colormap = {
+                id: src_mask_cat[id]
```
IRDonch, line 594: Shouldn't `id` be mapped here? Otherwise you're just recreating the source `MaskCategories` again.

datumaro/plugins/transforms.py
```diff
@@ -564,6 +567,66 @@ def transform_item(self, item):
...
+                for id, _ in enumerate(src_label_cat.items)
```
IRDonch, line 595:
```suggestion
                for id in range(len(src_label_cat.items))
```

datumaro/plugins/transforms.py
```diff
@@ -564,6 +567,66 @@ def transform_item(self, item):
...
+        if src_point_cat is not None:
+            assert src_label_cat is not None
+            dst_point_cat = PointsCategories(attributes=src_point_cat.attributes)
+            dst_point_cat.items = {
+                id: src_point_cat[id]
+                for id, _ in enumerate(src_label_cat.items)
+                if id in src_point_cat and (self._map_id(id) or id == 0)
+            }
+            self._categories[AnnotationType.points] = dst_point_cat
+
```
IRDonch, line 610: IMO, there should be a check here that no categories of unknown types are present. Otherwise, if a new category type is added in the future and this class is not updated, then old label IDs could leak into the transformed dataset.

datumaro/plugins/transforms.py
```diff
@@ -564,6 +567,66 @@ def transform_item(self, item):
...
+    def _make_label_id_map(self, src_label_cat, dst_label_cat):
+        self._map_id = lambda src_id: \
+            dst_label_cat.find(src_label_cat[src_id].name)[0]
```
IRDonch, line 613: This looks like it'll be inefficient if there are many different labels. I think you should build a dictionary mapping old label IDs to new ones.

datumaro/plugins/transforms.py
```diff
@@ -564,6 +567,66 @@ def transform_item(self, item):
...
+        self._categories[AnnotationType.label] = dst_label_cat
+
+    def categories(self):
+        return self._categories
+
+    def transform_item(self, item):
+        annotations = []
+        for ann in item.annotations:
+            if getattr(ann, 'label') is not None:
```
IRDonch, line 622: `getattr(ann, 'label')` is the same as `ann.label`. You probably meant to specify a `None` default value.

datumaro/util/annotation_util.py
```diff
@@ -214,12 +215,30 @@ def smooth_line(points, segments):
...
+    Returns:
+      function to map labels: src id -> dst id
+      dict: src id -> dst id
+      dict: src id -> src label
+      dict: dst id -> dst label
+    """
+
     source_labels = { id: label.name
-        for id, label in enumerate(src_labels or LabelCategories().items)
+        for id, label in enumerate(src_labels or LabelCategories())
```
IRDonch, line 238: Seems like this could be further simplified to:

```suggestion
        for id, label in enumerate(src_labels or ())
```

IRDonch (changes requested):

datumaro/plugins/transforms.py
zhiltsov-max: It is already a mapping. 2 dictionary lookups should be cheap enough.

datumaro/components/errors.py
```diff
@@ -164,27 +167,27 @@ def __str__(self):
         return "Failed to detect dataset format automatically: " \
             "no matching formats found"
 
-
-@attrs
 class DatasetError(DatumaroError):
-    item_id = attrib()
+    pass
```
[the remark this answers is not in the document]
zhiltsov-max: > I don't see it being used anywhere except as a base class.

It is supposed to be a base class. Mostly useful for exception catching, even though there is no use for it yet.

datumaro/plugins/transforms.py
zhiltsov-max: > there should be a check

Probably, a good point.

> could leak

Input categories are not copied in these transforms, so no.

IRDonch: > Input categories are not copied in these transforms, so no.

Okay, fair point.

However, the annotations _are_ copied. So perhaps we should make sure to only copy annotations of known types so that we can make sure that annotations aren't copied without the appropriate categories?

zhiltsov-max: Currently, addition of new annotation type is a rare event, and addition of a new categories is even more rare. I suppose, when it happen, it will require a number of similar changes across the code.

PR #434, opened 2021-08-26
imagenet_txt_format: add support for a custom labels file path

IRDonch:
### Summary

This will be needed to enable Accuracy Checker integration, since Accuracy Checker is separately configured with the annotation and class file paths.

As a prerequisite, also change the `Importer` interface so that importers can access extra parameters when searching for source files. I introduced a new `find_sources_with_params` function instead of adding a new parameter to `find_sources`, so that compatibility with old importers could be preserved.

### How to test
N/A

### Checklist

### License

```python
#
```


tests/test_imagenet_txt_format.py
```diff
@@ -146,6 +148,22 @@ def test_can_import(self):
...
+    @mark_requirement(Requirements.DATUM_GENERAL_REQ)
+    def test_can_import_with_custom_labels_file(self):
+        expected_dataset = Dataset.from_iterable([
+            DatasetItem(id='1', subset='train',
+                annotations=[Label(0)]
+            ),
+        ], categories={
+            AnnotationType.label: LabelCategories.from_iterable(
+                'alt_label_%s' % label for label in range(10)),
+        })
```
zhiltsov-max, line 160:
```suggestion
        ], categories=['alt_label_%s' % label for label in range(10)])
```

datumaro/plugins/imagenet_txt_format.py
```diff
@@ -17,7 +17,10 @@ class ImagenetTxtPath:
     IMAGE_DIR = 'images'
 
 class ImagenetTxtExtractor(SourceExtractor):
-    def __init__(self, path, labels=None, image_dir=None, subset=None):
+    def __init__(
+        self, path, *,
+        labels=None, labels_file=None, image_dir=None, subset=None,
+    ):
```
zhiltsov-max, line 23:
```suggestion
    def __init__(self, path, *,
            labels=None, labels_file=None, image_dir=None, subset=None):
```

zhiltsov-max (commented):
Could you check if it works through CLI? I suspect, it is the first real application for import parameters.

tests/assets/imagenet_txt_dataset_custom_labels/train.txt
```diff
@@ -0,0 +1 @@
+1.jpg 0
```
zhiltsov-max, line 1: We used subdirectories for subformats and extra examples in the test assets. Probably, for this format it should also be done this way. Look at coco and voc directories.

datumaro/plugins/imagenet_txt_format.py
```diff
@@ -29,9 +32,12 @@ def __init__(self, path, labels=None, image_dir=None, subset=None):
         self.image_dir = osp.join(osp.dirname(path), image_dir)
 
         if labels is None:
-            labels = osp.join(osp.dirname(path), ImagenetTxtPath.LABELS_FILE)
-            labels = self._parse_labels(labels)
+            if not labels_file:
+                labels_file = osp.join(osp.dirname(path), ImagenetTxtPath.LABELS_FILE)
+            labels = self._parse_labels(labels_file)
         else:
+            assert labels_file is None, "labels and labels_file must not both be set"
```
zhiltsov-max, line 39:
```suggestion
            assert not labels_file, "'labels' and 'labels_file' can't be used simultaneously"
```

tests/test_imagenet_txt_format.py
```diff
@@ -146,6 +148,22 @@ def test_can_import(self):
...
+
+        dataset = Dataset.import_from(DUMMY_DATASET_WITH_CUSTOM_LABELS_DIR, 'imagenet_txt',
+            labels_file=osp.join(DUMMY_DATASET_WITH_CUSTOM_LABELS_DIR, 'synsets-alt.txt'))
```
zhiltsov-max, line 163: Please, wrap lines at 80 chars, if possible.

datumaro/plugins/imagenet_txt_format.py
```diff
@@ -81,10 +87,13 @@ def _load_items(self, path):
 
 class ImagenetTxtImporter(Importer):
```
zhiltsov-max, line 88:
```suggestion
class ImagenetTxtImporter(Importer, CliPlugin):
    @classmethod
    def build_cmdline_parser(cls, **kwargs):
        parser = super().build_cmdline_parser(**kwargs)
        parser.add_argument('-l', '--labelmap', dest='labels_file',
            help="Path to the labelmap (synsets) file")

        return parser
```

datumaro/plugins/imagenet_txt_format.py
```diff
@@ -29,9 +32,12 @@ def __init__(self, path, labels=None, image_dir=None, subset=None):
         self.image_dir = osp.join(osp.dirname(path), image_dir)
 
         if labels is None:
-            labels = osp.join(osp.dirname(path), ImagenetTxtPath.LABELS_FILE)
-            labels = self._parse_labels(labels)
+            if not labels_file:
+                labels_file = osp.join(osp.dirname(path), ImagenetTxtPath.LABELS_FILE)
+            labels = self._parse_labels(labels_file)
```
zhiltsov-max, line 37: Probably, it would be useful to support passing just the filename instead of the full path.

tests/test_imagenet_txt_format.py
IRDonch: Applied.

datumaro/plugins/imagenet_txt_format.py
IRDonch: Applied.

datumaro/plugins/imagenet_txt_format.py
IRDonch: Applied.

tests/test_imagenet_txt_format.py
IRDonch: Done.

datumaro/plugins/imagenet_txt_format.py
IRDonch: Implemented (with the caveat that it's a relative path and not just the filename).

datumaro/plugins/imagenet_txt_format.py
IRDonch: Why should the key be called `--labelmap` when the parameter is `labels_file`?

tests/assets/imagenet_txt_dataset_custom_labels/train.txt
IRDonch: Okay, rearranged the assets.

datumaro/plugins/imagenet_txt_format.py
zhiltsov-max: Call it how you see fits better. Label map is kind of a general term for such things.

datumaro/plugins/imagenet_txt_format.py
```diff
@@ -29,9 +30,13 @@ def __init__(self, path, labels=None, image_dir=None, subset=None):
...
 
         if labels is None:
-            labels = osp.join(osp.dirname(path), ImagenetTxtPath.LABELS_FILE)
-            labels = self._parse_labels(labels)
+            labels_file = osp.join(osp.dirname(path),
+                labels_file if labels_file else ImagenetTxtPath.LABELS_FILE)
+            labels = self._parse_labels(labels_file)
         else:
+            assert not labels_file, \
+                "'labels' and 'labels_file' can't be used simultaneously"
```
zhiltsov-max, line 38: Can we overload the argument?

datumaro/plugins/imagenet_txt_format.py
zhiltsov-max: I suggest `labels_file or ImagenetTxtPath.LABELS_FILE`.

How about checking if there is just a file at the given path, then, if no, falling back to prepending a dataset directory?

IRDonch: > How about checking if there is just a file at the given path, then, if no, falling back to prepending a dataset directory?

I don't think that would lead to useful semantics, because then you would not be able to reliably use a dataset-relative path as the value of the parameter. If you wrote `ImagenetTxtExtractor(..., labels_file='foobar.txt')`, then it could unexpectedly load `foobar.txt` from the current directory if such a file existed there.

The semantics should be either "path relative to the current directory" or "path relative to the dataset directory", not both. In fact, I'm considering changing the code back to use the former semantics, since it's more consistent with how paths are usually handled in APIs.

IRDonch: > In fact, I'm considering changing the code back to use the former semantics, since it's more consistent with how paths are usually handled in APIs.

Did that.

datumaro/plugins/imagenet_txt_format.py
IRDonch: Okay, did it.

datumaro/plugins/imagenet_txt_format.py
IRDonch: I looked at existing occurrences of "label map" in the codebase, and it seems like that's usually used to describe maps of labels to colors (or vice versa). Whereas in this case we just have a list of labels, so I ended up going with `--labels_file`. I also didn't add a short option, since I don't expect this option to be frequently used.

datumaro/plugins/imagenet_txt_format.py
```diff
@@ -79,12 +82,23 @@ def _load_items(self, path):
         return items
 
 
-class ImagenetTxtImporter(Importer):
+class ImagenetTxtImporter(Importer, CliPlugin):
+    @classmethod
+    def build_cmdline_parser(cls, **kwargs):
+        parser = super().build_cmdline_parser(**kwargs)
+        parser.add_argument('--labels_file', dest='labels',
```
zhiltsov-max, line 89:
```suggestion
        parser.add_argument('--labels-file', dest='labels',
```

datumaro/plugins/imagenet_txt_format.py
zhiltsov-max: Assess which variant you like better:

`datum add -f imagenet_txt /full/path/to/dataset -- --labels-file /full/path/to/dataset/labels.txt`
`datum add -f imagenet_txt /full/path/to/dataset -- --labels-file labels.txt`

My preference is the second. It also allows to relocate the dataset without breaking import.

Imagine, there is an URL instead of the path.
`datum add -f imagenet_txt https://my.storage/path/to/dataset/ -- --labels-file ???`

IRDonch: > Assess which variant you like better:

I don't think there's a clear winner here, really. Sure, it's nice to not repeat the dataset path, but then you can have situations like this:

```
datum add -f imagenet_txt a/b/c -- --labels-file d/e/f.txt
```

where there are two paths on the command line, and the first one is relative to the current directory, while the second one is relative to the first one. That seems like surprising behavior to me. I could see a user trying to load a labels file from the current directory and failing because of this.

> It also allows to relocate the dataset without breaking import.

What do you mean? If the dataset is moved, then the import will be broken regardless, since the path to the dataset will be out of date, won't it?

> Imagine, there is an URL instead of the path.

It is a compelling use case, but how would that work in the first place? You need to be able to get the list of files to load ImageNet, and you can't get it over HTTP.

---

All in all, I'm not married to either approach, so if you think that interpreting the path relative to the dataset path is the better option, I can do that. I just wanted to air my concerns first.

IRDonch: Okay, I changed my mind. I just saw that the _image_ directory is already interpreted relative to the dataset. In that case, it obviously makes sense to do the same for the labels file.

datumaro/plugins/imagenet_txt_format.py
IRDonch: Fixed.

datumaro/plugins/imagenet_txt_format.py
IRDonch: And done.

zhiltsov-max: > where there are two paths on the command line, and the first one is relative to the current directory, while the second one is relative to the first one. That seems like surprising behavior to me. I could see a user trying to load a labels file from the current directory and failing because of this.

That's why I suggested to do 2 file checks above - one for the local/absolute path, and another for the dataset-relative one. Actually, I don't see much sense in pointing to a file outside the dataset directory.

> What do you mean? If the dataset is moved, then the import will be broken regardless, since the path to the dataset will be out of date, won't it?

I meant it in the scope of copying the dataset inside the project directory. Once it is copied, it is relocated, and then it can be relocated further (together with the project). The dataset is also going to be in an arbitrary places after it is downloaded, or materialized in the project cache.

> It is a compelling use case, but how would that work in the first place? You need to be able to get the list of files to load ImageNet, and you can't get it over HTTP.

OK, it is a good point. Let's change HTTP to any CSP then - s3, azure, etc.

zhiltsov-max (approved):

PR #439, opened 2021-08-31
Factor out annotation-related functionality into a separate module

IRDonch:
### Summary
Annotations are only tangentially related to extractors, so I don't think they belong in `extractor.py`. `extractor.py` is also a large module, so removing a chunk of it makes it easier to navigate.

### Checklist

### License

```python
#
```


IRDonch:
@zhiltsov-max This PR is incomplete (I haven't updated the examples in the documentation yet). Let me know if you agree with this refactoring and if you do, I'll finish it.

zhiltsov-max:
Thought of it. It does not really matter until we have stable and documented API, but this breaks the library API.

IRDonch:
> It does not really matter until we have stable and documented API, but this breaks the library API.

I could add compatibility exports if that's your concern. Although I kinda figured that with the versioning changes we're pretty much breaking all compatibility anyway. :P

IRDonch:
So what should I do? Should I add the compatibility exports?

zhiltsov-max:
Let's put off the PR till we introduce the stable API after the versioning release.

IRDonch:
> Let's put off the PR till we introduce the stable API after the versioning release.

I don't think that's a good course of action; it would be much harder to bring the PR up-to-date at that point (which could be long into the future) than it would be to just add compatibility exports now and then remove them later. Plus, by merging this now we can improve the navigability of `extractor.py` in the interim (I initiated this refactoring in the first place because I kept getting lost in it).

zhiltsov-max:
Ok, then add compatibility exports and it is ready.

IRDonch:
OK, done and updated the documentation.

zhiltsov-max (approved):

PR #443, opened 2021-09-02
lazy_image: fix cache collisions leading to unrelated data being returned

### LINKED ISSUE #409: Rare test failures on windows CI
It doesn't fail every time.

Example: https://github.com/openvinotoolkit/datumaro/pull/407/checks?check_run_id=3359354762

```
__________________________ BytesImageTest.test_ctors __________________________

self = <tests.test_images.BytesImageTest testMethod=test_ctors>

    def test_ctors(self):
        with TestDir() as test_dir:
            path = osp.join(test_dir, 'path.png')
            image = np.ones([2, 4, 3])
            image_bytes = encode_image(image, 'png')
    
            for args in [
                { 'data': image_bytes },
                { 'data': lambda _: image_bytes },
                { 'data': lambda _: image_bytes, 'ext': '.jpg' },
                { 'data': image_bytes, 'path': path },
                { 'data': image_bytes, 'path': path, 'size': (2, 4) },
                { 'data': image_bytes, 'path': path, 'size': (2, 4) },
                { 'path': path },
                { 'path': path, 'size': (2, 4) },
            ]:
                with self.subTest(**args):
                    img = ByteImage(**args)
                    # pylint: disable=pointless-statement
                    self.assertEqual('data' in args, img.has_data)
                    if img.has_data:
>                       self.assertEqual(img, image)

tests\test_images.py:126: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
datumaro\util\image.py:366: in __eq__
    return super().__eq__(other)
datumaro\util\image.py:311: in __eq__
    return self.has_data and np.array_equal(self.data, other)
datumaro\util\image.py:282: in data
    data = self._data()
datumaro\util\image.py:227: in __call__
    image = self.loader(self.path)
datumaro\util\image.py:338: in <lambda>
    loader = lambda _: decode_image(self.get_bytes())
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

image_bytes = <object object at 0x0000025118C7B6D0>
dtype = <class 'numpy.float32'>

    def decode_image(image_bytes, dtype=np.float32):
        if _IMAGE_BACKEND == _IMAGE_BACKENDS.cv2:
            import cv2
>           image = np.frombuffer(image_bytes, dtype=np.uint8)
E           TypeError: a bytes-like object is required, not 'object'

datumaro\util\image.py:155: TypeError
```

IRDonch:
### Summary
Currently, the key used to look up the cached image is based on a hash of a tuple containing `id(self)`, `path`, and `loader`. This means there are two situations in which a `lazy_image` can look up the wrong data:

* If there previously existed another `lazy_image` object whose `self` and `loader` had the same object IDs as the current `lazy_image`'s `self` and `loader`. This is possible, because deleted objects' IDs can be reused.

* If a hash collision occurs between the current `lazy_image`'s tuple and some other's.

Fix it by using a weak reference to `self` as the key instead. Different weak references will only compare equal if they point to the same object.

This will only work correctly if the loader and path of a `lazy_image` are not modified after creation. I don't think there are any use cases for modifying them (and there are no instances of that happening in the codebase), so it shouldn't be an issue. To reduce the temptation of client code to modify these fields, mark them as private.

Modifying the `cache` field should not cause issues, but just in case, make it private as well.

Fixes #409 (probably; I never managed to reliably reproduce it)

### Checklist

### License

```python
#
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Deprecation warning in `open_images_format.py`
   (<https://github.com/openvinotoolkit/datumaro/pull/440>)
+- `lazy_image` returning unrelated data sometimes
+  (<https://github.com/openvinotoolkit/datumaro/issues/409>)
 
 ### Security
 - TBD
--- a/datumaro/util/image.py
+++ b/datumaro/util/image.py
@@ -12,6 +12,7 @@ import os
 import os.path as osp
 import shlex
 import shutil
+import weakref
 
 import numpy as np
 
@@ -205,28 +206,28 @@ class lazy_image:
     def __init__(self, path, loader=None, cache=None):
         if loader is None:
             loader = load_image
-        self.path = path
-        self.loader = loader
+        self._path = path
+        self._loader = loader
 
         # Cache:
         # - False: do not cache
         # - None: use the global cache
         # - object: an object to be used as cache
         assert cache in {None, False} or isinstance(cache, object)
-        self.cache = cache
+        self._cache = cache
 
     def __call__(self):
         image = None
-        image_id = hash(self) # path is not necessary hashable or a file path
+        cache_key = weakref.ref(self)
 
-        cache = self._get_cache(self.cache)
+        cache = self._get_cache(self._cache)
         if cache is not None:
-            image = cache.get(image_id)
+            image = cache.get(cache_key)
 
         if image is None:
-            image = self.loader(self.path)
+            image = self._loader(self._path)
             if cache is not None:
-                cache.push(image_id, image)
+                cache.push(cache_key, image)
         return image
 
     @staticmethod
@@ -236,9 +237,6 @@ class lazy_image:
         elif cache == False:
             return None
         return cache
-
-    def __hash__(self):
-        return hash((id(self), self.path, self.loader))
 
 class Image:
     def __init__(self, data: Union[None, Callable, np.ndarray] = None,


zhiltsov-max:
> This will only work correctly if the loader and path of a lazy_image are not modified after creation.

Maybe, make it frozen / tuple?


I don't see that in the current version different objects can be matched to the same cache entry, because all keys are unique. I won't insist on this requirement (especially, we discovered earlier that cache should have a small number of objects), however, I see such property quite useful.
If you want to implement this, consider using `hash(type(self), path, loader)`. I suspect, there will be similar loader for other media types.

IRDonch:
> Maybe, make it frozen / tuple?

That would require it being an `attrs`-based class, and I don't think it's worthwhile to make it so. Marking the attributes as private should be sufficient to let client code know not to mess with them.

> I don't see that in the current version different objects can be matched to the same cache entry, because all keys are unique.

That is true, but it was also true in the previous version (if you discount hash collisions). I haven't really changed the semantics here.

> I won't insist on this requirement (especially, we discovered earlier that cache should have a small number of objects), however, I see such property quite useful.

I have pondered upon how one might implement this, and it seems like it would be pretty complicated.

In general, you can't decide, given an abstract pair of (`path`, `loader`), whether the result of calling `loader(path)` will be the same as it was before. If `path` is an actual disk path, the file at that path might have already changed. Therefore, to be able to reuse a cache entry of another `lazy_image`, we would have to have some way of asking the loader whether the underlying resource has changed.

I think this _could_ be done, but that's more effort than I signed up for. I'm just trying to fix the bug. 🙂 

> If you want to implement this, consider using `hash(type(self), path, loader)`.

FWIW, that would lead to the reoccurrence of this bug. Any kind of compound key would have to be a tuple of objects, not a number.

zhiltsov-max (approved):

zhiltsov-max:
I'm pretty sure this topic will be revisited later, once we start working with large objects, which aren't cheap to load.

PR #444, opened 2021-09-02
Replace Rollback with Scope

zhiltsov-max:
### Summary

Replaces error rollback utilities with scope utilities. The new class extends the previous situation to be like try-except-finally and allows to enter context managers. This helps to maintain flat code indentation and helps with the objects that require closing / disposal in any cases (i.e. resource managers). The `Project` class is going to be such a manager - this is needed to resolve problems with file removal on Windows.

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/util/scope.py
+++ b/datumaro/util/scope.py
@@ -25,7 +25,7 @@ class Scope:
         callback: Callable[[], Any]
         ignore_errors: bool = True
 
-        def __exit__(self, exc_type=None, exc_value=None, exc_traceback=None):
+        def __exit__(self, exc_type, exc_value, exc_traceback):
             try:
                 self.callback()
             except Exception:
@@ -34,7 +34,7 @@ class Scope:
 
     @attrs
     class ErrorHandler(ExitHandler):
-        def __exit__(self, exc_type=None, exc_value=None, exc_traceback=None):
+        def __exit__(self, exc_type, exc_value, exc_traceback):
             if exc_type:
                 return super().__exit__(exc_type=exc_type, exc_value=exc_value,
                     exc_traceback=exc_traceback)
@@ -54,7 +54,7 @@ class Scope:
         will be ignored.
         """
 
-        return self._register_callback(self.ErrorHandler,
+        self._register_callback(self.ErrorHandler,
             ignore_errors=ignore_errors,
             callback=callback, args=args, kwargs=kwargs)
 
@@ -65,7 +65,7 @@ class Scope:
         Registers a function to be called on scope exit.
         """
 
-        return self._register_callback(self.ExitHandler,
+        self._register_callback(self.ExitHandler,
             ignore_errors=ignore_errors,
             callback=callback, args=args, kwargs=kwargs)
 
@@ -93,16 +93,17 @@ class Scope:
         self.enabled = False
 
     def close(self):
-        self.__exit__()
+        self.__exit__(None, None, None)
 
-    def __enter__(self):
+    def __enter__(self) -> 'Scope':
         return self
 
-    def __exit__(self, exc_type=None, exc_value=None, exc_traceback=None):
+    def __exit__(self, exc_type, exc_value, exc_traceback):
         if not self.enabled:
             return
 
         self._stack.__exit__(exc_type, exc_value, exc_traceback)
+        self._stack.pop_all() # prevent issues on repetitive calls
 
     @classmethod
     def current(cls) -> 'Scope':
--- a/datumaro/util/scope.py
+++ b/datumaro/util/scope.py
@@ -11,7 +11,6 @@ from attr import attrs
 
 from datumaro.util import optional_arg_decorator
 
-
 T = TypeVar('T')
 
 class Scope:
--- a/datumaro/util/scope.py
+++ b/datumaro/util/scope.py
@@ -4,13 +4,15 @@
 
 from contextlib import ExitStack, contextmanager
 from functools import partial, wraps
-from typing import Any, ContextManager, Dict, Optional
+from typing import Any, ContextManager, Dict, Optional, TypeVar
 import threading
 
 from attr import attrs
 
 from datumaro.util import optional_arg_decorator
 
+
+T = TypeVar('T')
 
 class Scope:
     """
@@ -83,9 +85,11 @@ class Scope:
         self._exit_stack.callback(handler)
         return name
 
-    def add(self, cm: ContextManager) -> Any:
+    def add(self, cm: ContextManager[T]) -> T:
         """
         Enters a context manager and adds it to the exit stack.
+
+        Returns: cm.__enter__() result
         """
 
         return self._exit_stack.enter_context(cm)
@@ -152,17 +156,19 @@ def scoped(func, arg_name=None):
     return wrapped_func
 
 # Shorthands for common cases
-def on_error_do(callback, *args, ignore_errors=False):
+def on_error_do(callback, *args, ignore_errors=False,
+        fwd_kwargs=None, **kwargs):
     return Scope.current().on_error_do(callback, *args,
-        ignore_errors=ignore_errors)
+        ignore_errors=ignore_errors, fwd_kwargs=fwd_kwargs, **kwargs)
 on_error_do.__doc__ = Scope.on_error_do.__doc__
 
-def on_exit_do(callback, *args, ignore_errors=False):
+def on_exit_do(callback, *args, ignore_errors=False,
+        fwd_kwargs=None, **kwargs):
     return Scope.current().on_exit_do(callback, *args,
-        ignore_errors=ignore_errors)
+        ignore_errors=ignore_errors, fwd_kwargs=fwd_kwargs, **kwargs)
 on_exit_do.__doc__ = Scope.on_exit_do.__doc__
 
-def add(cm: ContextManager):
+def add(cm: ContextManager[T]) -> T:
     return Scope.current().add(cm)
 add.__doc__ = Scope.add.__doc__
 
--- a/datumaro/util/scope.py
+++ b/datumaro/util/scope.py
@@ -147,5 +147,5 @@ def on_exit_do(callback, *args, ignore_errors=False, kwargs=None):
     return Scope.current().on_exit_do(callback, *args,
         ignore_errors=ignore_errors, kwargs=kwargs)
 
-def add(cm: ContextManager[T]) -> T:
+def scope_add(cm: ContextManager[T]) -> T:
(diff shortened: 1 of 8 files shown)


zhiltsov-max:
A draft for automatic cm closing:

``` python
def test_cm_closing(self):
        import inspect
        this_f = inspect.currentframe()
        begin_values = set(id(v) for v in this_f.f_locals.values())

        f2 = inspect.getframeinfo(this_f)
        q = 5

        end_values = set(id(v) for v in this_f.f_locals.values())
        ids_to_release = end_values - begin_values
        for v in this_f.f_locals.values():
            if id(v) in ids_to_release:
                if hasattr(v, 'close'):
                    v.close()
                elif hasattr(v, '__exit__'):
                    v.__exit__()
```

datumaro/util/scope.py
```diff
@@ -0,0 +1,171 @@
...
+def on_exit_do(callback, *args, ignore_errors=False):
+    return Scope.current().on_exit_do(callback, *args,
+        ignore_errors=ignore_errors)
+on_exit_do.__doc__ = Scope.on_exit_do.__doc__
+
+def add(cm: ContextManager):
+    return Scope.current().add(cm)
+add.__doc__ = Scope.add.__doc__
+
+def current():
```
IRDonch, line 169: I don't think `add` and `current` are good names for module-scope functions. They're too ambiguous if imported using `from ... import` (and I don't think you want to do `import datumaro.util.scope; datumaro.util.scope.add(...)`).

`current` in particular doesn't seem necessary at all. `Scope.current()` is already short enough.

IRDonch (changes requested):

datumaro/util/scope.py
```diff
@@ -0,0 +1,171 @@
...
+        Registers a function to be called on scope exit because of an error.
+        Equivalent to the "except" block of "try-except".
+        """
+
+        if args or kwargs or fwd_kwargs:
+            if fwd_kwargs:
+                kwargs.update(fwd_kwargs)
+            callback = partial(callback, *args, **kwargs)
+
+        name = name or hash(callback)
```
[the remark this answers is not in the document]
zhiltsov-max: You're probably right about collisions, but I'm not sure about allowing complete repetitions in both stacks. It doesn't look right for me.

datumaro/util/scope.py
zhiltsov-max: I want these two functions to be used only as `import datumaro.util.scope as scope; scope.add(); scope.current()`. Dropping all the free functions and making the `Scope` class a decorator with static public methods is an option, but it means mixing 3 interfaces in a single entity, which I don't want to do.

datumaro/util/scope.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def __exit__(self, exc_type=None, exc_value=None, exc_traceback=None):
+        if not self.enabled:
+            return
+
+        try:
+            if exc_type:
+                self._error_stack.__exit__(exc_type, exc_value, exc_traceback)
+        finally:
+            self._exit_stack.__exit__(exc_type, exc_value, exc_traceback)
```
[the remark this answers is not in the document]
zhiltsov-max: So the idea is to wrap each callback in a function and check the exception type inside?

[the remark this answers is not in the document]
IRDonch: You're already wrapping each callback in a `Handler` object - you just need to use `push` instead of `callback` and adjust the arguments of `Handler.__call__` accordingly.

datumaro/util/scope.py
IRDonch: `scope.current()` still seems pointless - `Scope.current()` is the same number of characters and doesn't require an extra wrapper function.

As for `add`, IMO, it would be more convenient to rename it to `scope_add` - then you could use the same style of import for it as for everything else.

tests/test_util.py
```diff
@@ -1,127 +1,143 @@
...
 from datumaro.util.test_utils import TestDir
 
 from .requirements import Requirements, mark_requirement
 
 
-class TestRollback(TestCase):
+class TestException(Exception):
+    pass
+
+class TestScope(TestCase):
```
IRDonch, line 17:
```suggestion
class ScopeTest(TestCase):
```

tests/test_util.py
```diff
@@ -1,127 +1,143 @@
...
-        def cb():
-            nonlocal success
-            success = False
+    def test_calls_on_no_error(self):
+        error_cb = mock.MagicMock()
+        exit_cb = mock.MagicMock()
 
-        with Rollback() as on_error:
-            on_error.do(cb)
+        with suppress(TestException), Scope() as scope:
```
IRDonch, line 23: The suppression shouldn't be needed here'

tests/test_util.py
```diff
@@ -1,127 +1,143 @@
...
-                raise Exception('err')
-        except Exception: # nosec - disable B110:try_except_pass check
-            pass
-        finally:
-            self.assertTrue(success)
+    def test_calls_both_stacks_on_error(self):
+        error_cb = mock.MagicMock()
+        exit_cb = mock.MagicMock()
+
+        with suppress(TestException), Scope() as scope:
```
IRDonch, line 35: Instead of `suppress`, you should use `assertRaises` to make sure that the scope does not absorb the exception.

datumaro/util/scope.py
```diff
@@ -0,0 +1,171 @@
...
+    class Handler:
+        callback: Any
+        enabled: bool = True
+        ignore_errors: bool = False
+
+        def __call__(self):
+            if self.enabled:
+                try:
+                    self.callback()
+                except: # pylint: disable=bare-except
```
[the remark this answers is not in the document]
zhiltsov-max: Done

datumaro/util/scope.py
```diff
@@ -0,0 +1,171 @@
...
+class Scope:
+    """
+    A context manager that allows to register error and exit callbacks.
+    """
+
+    _thread_locals = threading.local()
+
+    @attrs(auto_attribs=True)
+    class Handler:
+        callback: Any
```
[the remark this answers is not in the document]
zhiltsov-max: Done

datumaro/util/scope.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def __init__(self):
+        self._handlers = {}
+        self._error_stack = ExitStack()
+        self._exit_stack = ExitStack()
+        self.enabled = True
+
+    def on_error_do(self, callback, *args, name: Optional[str] = None,
+            enabled: bool = True, ignore_errors: bool = False,
+            fwd_kwargs: Optional[Dict[str, Any]] = None, **kwargs) -> str:
```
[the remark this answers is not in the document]
zhiltsov-max: Done. I don't like the thing that positionals are just passed, but kwargs need a special named arg. But having a names arg for positionals looks clumsy.

datumaro/util/scope.py
```diff
@@ -0,0 +1,171 @@
...
+            callback = partial(callback, *args, **kwargs)
+
+        name = name or hash(callback)
+        assert name not in self._handlers, "Callback is already registered"
+
+        handler = self.Handler(callback,
+            enabled=enabled, ignore_errors=ignore_errors)
+        self._handlers[name] = handler
+        self._exit_stack.callback(handler)
+        return name
```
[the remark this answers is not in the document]
zhiltsov-max: Done

datumaro/util/scope.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def add(self, cm: ContextManager) -> Any:
```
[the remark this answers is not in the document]
zhiltsov-max: Done

datumaro/util/scope.py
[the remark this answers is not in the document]
zhiltsov-max: Done

datumaro/util/scope.py
zhiltsov-max: I'm not sure about `scope_add`, but I don't dislike it enough.

tests/test_util.py
zhiltsov-max: Done

tests/test_util.py
zhiltsov-max: Fixed

tests/test_util.py
zhiltsov-max: Fixed

zhiltsov-max:
@IRDonch, I refactored and simplified the class.

datumaro/util/scope.py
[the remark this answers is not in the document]
IRDonch: Well, looks like you removed name support entirely, so I guess the point is moot now.

datumaro/util/scope.py
```diff
@@ -0,0 +1,150 @@
...
+                    exc_traceback=exc_traceback)
+
+
+    def __init__(self):
+        self._stack = ExitStack()
+        self.enabled = True
+
+    def on_error_do(self, callback: Callable,
+            *args, kwargs: Optional[Dict[str, Any]] = None,
+            ignore_errors: bool = False):
```
IRDonch, line 49:
```suggestion
            ignore_errors: bool = False) -> None:
```

(and ditto for the other `None`-returning functions)

datumaro/util/scope.py
```diff
@@ -0,0 +1,150 @@
...
+    """
+
+    _thread_locals = threading.local()
+
+    @attrs(auto_attribs=True)
+    class ExitHandler:
+        callback: Callable[[], Any]
+        ignore_errors: bool = True
+
+        def __exit__(self, exc_type=None, exc_value=None, exc_traceback=None):
```
IRDonch, line 28:
```suggestion
        def __exit__(self, exc_type, exc_value, exc_traceback):
```

`__exit__` is always called with three arguments, so the default values are unnecessary.

datumaro/util/scope.py
zhiltsov-max: What is the point in this?

IRDonch: To help typecheckers find errors.

zhiltsov-max: So you suggest to add it to all function without return value?

IRDonch: I would say yes (if there are annotations on the function's arguments). Otherwise, the return type defaults to `Any`.

zhiltsov-max: I'm not convinced it is worthy investment of efforts in most cases. I'm open for a PR on this, but I don't think it is needed for the functions, which names are not suppose any return value.

PR #446, opened 2021-09-06
Add 'add-background' parameter to VOC converter

yasakova-anastasia:
### Summary

- Added 'add-background' parameter to VOC converter

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/plugins/voc_format/converter.py
+++ b/datumaro/plugins/voc_format/converter.py
@@ -93,12 +93,15 @@ class VocConverter(Converter):
         parser.add_argument('--tasks', type=cls._split_tasks_string,
             help="VOC task filter, comma-separated list of {%s} "
                 "(default: all)" % ', '.join(t.name for t in VocTask))
+        parser.add_argument('--add-background', type=str_to_bool, default=True,
+            help="Allow export dataset with/without adding background label "
+                "(default: %(default)s)")
 
         return parser
 
-    def __init__(self, extractor, save_dir,
-            tasks=None, apply_colormap=True, label_map=None,
-            allow_attributes=True, keep_empty=False, **kwargs):
+    def __init__(self, extractor, save_dir, tasks=None,
+            apply_colormap=True, label_map=None, allow_attributes=True,
+            keep_empty=False, add_background=True, **kwargs):
         super().__init__(extractor, save_dir, **kwargs)
 
         assert tasks is None or isinstance(tasks, (VocTask, list, set))
@@ -113,6 +116,7 @@ class VocConverter(Converter):
         self._apply_colormap = apply_colormap
         self._allow_attributes = allow_attributes
         self._keep_empty = keep_empty
+        self._add_background = add_background
 
         if label_map is None:
             label_map = LabelmapType.source.name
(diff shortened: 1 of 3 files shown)


site/content/en/docs/formats/pascal_voc.md
```diff
@@ -212,6 +212,9 @@ datum import -o project -f voc -i ./VOC2012
 datum export -p project -f voc -- --tasks detection,classification
 ```
 
+- `--add-background ADD_BACKGROUND` - allow export dataset with/without
+adding background label (by default: `True`)
```
zhiltsov-max, line 216: Add a recommendation about how and when to use the parameter.

datumaro/plugins/voc_format/converter.py
```diff
@@ -559,16 +563,17 @@ def _load_categories(self, label_map_source):
...
+            bg_label = find(label_map.items(), lambda x: x[1][0] == (0, 0, 0))
+            if bg_label is not None:
+                bg_label = bg_label[0]
+            else:
+                bg_label = 'background'
+                if bg_label not in label_map:
+                    has_colors = any(v[0] is not None for v in label_map.values())
+                    color = (0, 0, 0) if has_colors else None
+                    label_map[bg_label] = [color, [], []]
+            label_map.move_to_end(bg_label, last=False)
```
zhiltsov-max, line 576: Please clarify the logic behind this parameter. How is it supposed to work, when there are masks?

tests/test_voc_format.py
```diff
@@ -1127,3 +1127,32 @@ def categories(self):
...
+                        'truncated': False,
+                        'difficult': False,
+                        'occluded': False,
+                    })
+                ]),
+        ], categories={
+            AnnotationType.label: LabelCategories.from_iterable(
+                ['a', 'b']),
+            AnnotationType.mask: MaskCategories(
+                colormap=VOC.generate_colormap(2)),
```
zhiltsov-max, line 1146: Probably, you need to remove the background color to test it. Consider using `MaskCategories.generate(include_background=False)`

datumaro/plugins/voc_format/converter.py
yasakova-anastasia: I fixed one point here. This parameter is only useful when there is no background label. If `True`, we add background, otherwise we don't add it. If there are masks, we need to use `add_background = True` (by default, `True`).

tests/test_voc_format.py
yasakova-anastasia: If the background label will always be added for masks, do we need to change something for the mask categories?

datumaro/plugins/voc_format/converter.py
zhiltsov-max: > If there are masks, we need to use ...

Who? The user? Code? Do we need to throw errors in specific cases?

tests/test_voc_format.py
zhiltsov-max: I don't understand the question. Now, the background label is `a` in the `expected` dataset.

site/content/en/docs/formats/pascal_voc.md
zhiltsov-max: I still don't understand how and when I, as a user, supposed to use this option. In the original VOC format there is no dedicated label map at all, so the description is confusing. Add where?

yasakova-anastasia:
I will close this PR. I'll come back to this question when I'm doing experiments on conversion

PR #481, opened 2021-10-01
Sk/fix kitti

kirill-sizov:
### Summary
* Allowed to load not original KITTI labels for detection task
* Fixed labels `id` for exporting masks

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/plugins/kitti_format/format.py
+++ b/datumaro/plugins/kitti_format/format.py
@@ -75,6 +75,7 @@ class KittiPath:
     MASK_EXT = '.png'
 
     LABELMAP_FILE = 'label_colors.txt'
+    LABELS_LIST_FILE = 'labels.txt'
 
     DEFAULT_TRUNCATED = 0.0 # 0% truncated
     DEFAULT_OCCLUDED = 0    # fully visible
@@ -143,3 +144,8 @@ def write_label_map(path, label_map):
             else:
                 color_rgb = ''
             f.write('%s %s\n' % (color_rgb, label_name))
+
+def write_labels_list(path, labels):
+    with open(path, 'w', encoding='utf-8') as f:
+        for label in labels:
+            f.write('%s\n' % label)
--- a/tests/test_kitti_format.py
+++ b/tests/test_kitti_format.py
@@ -2,6 +2,7 @@ from collections import OrderedDict
 from functools import partial
 from unittest import TestCase
 import os.path as osp
+from attr import attributes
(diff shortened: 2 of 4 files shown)


site/content/en/docs/formats/kitti.md
```diff
@@ -48,6 +48,8 @@ KITTI segmentation dataset directory should have the following structure:
 <!--lint disable fenced-code-flag-->
 ```
 └─ Dataset/
+    ├── label_colors.txt # optional, color map for non-original segmentation labels
+    ├── labels.txt # optional, file for non-original detection labels
```
zhiltsov-max, line 52: Can we use only one file?

kirill-sizov: Yes, we can, but the format support the ability to have different labels for detection and segmentation tasks

datumaro/plugins/kitti_format/format.py
```diff
@@ -143,3 +144,8 @@ def write_label_map(path, label_map):
             else:
                 color_rgb = ''
             f.write('%s %s\n' % (color_rgb, label_name))
+
+def write_labels_list(path, labels):
```
zhiltsov-max, line 148: I suppose this function is not used.

site/content/en/docs/formats/kitti.md
zhiltsov-max: What happens, if there is 2 files?

datumaro/plugins/kitti_format/format.py
kirill-sizov: Yes, it is, so I changed this solution.

tests/test_kitti_format.py
```diff
@@ -459,21 +454,62 @@ def __iter__(self):
...
+                        attributes={'is_crowd': False}, label=2, id=2)
+                ]
+            )
+        ], categories=make_kitti_categories(source_label_map))
+
+        expected_label_map = {
+            'background': (0, 0, 0),
+            'label_0': (20, 20, 20),
+            'label_1': (10, 10, 10)
+        }
```
zhiltsov-max, line 496: Why the labels are sorted here?

kirill-sizov: Because in this solution label indexes may change after the import-export procedure. I agree that this is a disadvantage of such a solution. But I think this is better than requiring the user to fill in an additional file with a list of tags.

PR #489, opened 2021-10-04
imagenet_txt_format: implement automatic category creation

IRDonch:
### Summary
I implemented this as part of prototyping Accuracy Checker integration. AC is able to load ImageNet without a labels file, and while you can notionally do that with Datumaro by supplying an empty label list, it doesn't work in practice. As soon as Datumaro encounters the first annotation, it fails, since the label ID fails the range check.

Solve this by adding an option to automatically create new categories with generic names if an out-of-range label ID is encountered.

Also, add a way to specify an empty label list from the command line.

### Checklist

### License

```python
#
```


datumaro/plugins/imagenet_txt_format.py
```diff
@@ -87,17 +102,34 @@ def build_cmdline_parser(cls, **kwargs):
...
+            action='store_const', const=(),
+            help="Do not load a file with label descriptions")
+        parser.add_argument('--auto-labels', action='store_true',
+            help="Automatically create generic label descriptions "
+                "if an annotation refers to a label "
+                "for which no description exists")
         return parser
 
     @classmethod
     def find_sources_with_params(cls, path, **extra_params):
```
zhiltsov-max, line 115: How about passing extra params as named args? This will automatically document the code and add checks for unsupported agruments.

datumaro/plugins/imagenet_txt_format.py
```diff
@@ -87,17 +102,34 @@ def build_cmdline_parser(cls, **kwargs):
...
+            def file_filter(p):
+                return osp.basename(p) != labels_file_name
+        elif labels is None:
+            def file_filter(p):
+                return osp.basename(p) != ImagenetTxtPath.LABELS_FILE
+        else:
+            file_filter = None
+
         labels_file_name = osp.basename(labels) \
             if isinstance(labels, str) else ImagenetTxtPath.LABELS_FILE
```
zhiltsov-max, line 129: The variable is redefined in the `str` case and unused in others.

datumaro/plugins/imagenet_txt_format.py
```diff
@@ -87,17 +102,34 @@ def build_cmdline_parser(cls, **kwargs):
         parser = super().build_cmdline_parser(**kwargs)
         parser.add_argument('--labels-file', dest='labels',
             help="Path to the file with label descriptions (synsets.txt)")
+        parser.add_argument('--no-labels-file', dest='labels',
+            action='store_const', const=(),
+            help="Do not load a file with label descriptions")
```
zhiltsov-max, line 107: Consider changing the option to be `--labels` and accept value `auto` or making a mutually-exclusive group. Otherwise, the behavior can be undefined or obscure, at least.

datumaro/plugins/imagenet_txt_format.py
```diff
@@ -87,17 +102,34 @@ def build_cmdline_parser(cls, **kwargs):
...
+        parser.add_argument('--auto-labels', action='store_true',
+            help="Automatically create generic label descriptions "
+                "if an annotation refers to a label "
+                "for which no description exists")
```
zhiltsov-max, line 111:
```suggestion
            help="Use auto-generated labels for annotations, "
                "when no label information is available in the dataset")
```

IRDonch: That doesn't describe the option correctly. It has an effect even if label information _is_ available.

zhiltsov-max: How it can be used in such cases?

PR #493, opened 2021-10-05
Improve type annotations in annotations, add more docs

zhiltsov-max:
### Summary

- Improved type annotations for annotation and dataset classes
- Added more docs
- Refactored `CompiledMask.from_instance_masks`

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/components/annotation.py
+++ b/datumaro/components/annotation.py
@@ -3,7 +3,10 @@
 # SPDX-License-Identifier: MIT
 
 from enum import Enum, auto
-from typing import Tuple, Union
+from typing import (
+    Any, Callable, Dict, Iterable, Iterator, List, Literal, Optional, Set,
+    Tuple, Union,
+)
 
 from attr import attrib, attrs
 import attr
@@ -22,13 +25,21 @@ class AnnotationType(Enum):
     caption = auto()
     cuboid_3d = auto()
 
-_COORDINATE_ROUNDING_DIGITS = 2
+COORDINATE_ROUNDING_DIGITS = 2
 
 @attrs(kw_only=True)
 class Annotation:
-    id = attrib(default=0, validator=default_if_none(int))
-    attributes = attrib(factory=dict, validator=default_if_none(dict))
-    group = attrib(default=0, validator=default_if_none(int))
+    """
+    A base annotation class.
+
+    Derived classes must define the '_type' class variable with a value
+    from the AnnotationType enum.
+    """
+
+    id: int = attrib(default=0, validator=default_if_none(int))
+    attributes: Dict[str, Any] = attrib(
+        factory=dict, validator=default_if_none(dict))
+    group: int = attrib(default=0, validator=default_if_none(int))
 
     def __attrs_post_init__(self):
         assert isinstance(self.type, AnnotationType)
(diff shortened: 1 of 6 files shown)


datumaro/components/annotation.py
```diff
@@ -178,40 +203,55 @@ def __eq__(self, other):
 
 @attrs(eq=False)
 class Mask(Annotation):
+    """
+    Represents a 2d single-instance binary segmentation mask.
+    """
+
+    BinaryMask = np.ndarray # 2d of type bool
```
IRDonch, line 210: It's a bit unclear from the documentation, but I don't think type aliases are allowed at local scope. Instead, this declaration is treated as a variable of type `Type[np.ndarray]`. It's mentioned in [PEP 613](https://www.python.org/dev/peps/pep-0613/#id6), and in practice, mypy refuses to validate such usage. You can check it yourself by running mypy on this simplified example:

```python
import numpy as np

class C:
    x = np.ndarray

    def foo(self) -> x:
        ...
```

datumaro/components/annotation.py
```diff
@@ -266,46 +328,72 @@ def from_instance_masks(instance_masks,
         else:
             instance_labels = [None] * len(instance_masks)
 
-        instance_masks = sorted(enumerate(instance_masks),
-            key=lambda m: m[1].z_order)
-        instance_masks = ((m.image, 1 + j,
-                instance_ids[i] if instance_ids[i] is not None else 1 + j,
-                instance_labels[i] if instance_labels[i] is not None else m.label
-            ) for j, (i, m) in enumerate(instance_masks))
+        max_index = len(instance_masks) + 1
```
IRDonch, line 331: You can't apply `len` to an `Iterable`. You have to either convert it to a list, or change the type to something more restrictive (e.g. `Sequence`).

datumaro/components/annotation.py
```diff
@@ -316,45 +404,60 @@ def _get_image(image):
...
-            for k in keys if k & ((1 << class_shift) - 1) != 0
+        instance_labels = {
+            k & ((1 << class_shift) - 1): k >> class_shift
+            for k in keys
+            if k & ((1 << class_shift) - 1) != 0
         }
         return instance_labels
 
-    def extract(self, instance_id):
+    def extract(self, instance_id: int):
```
IRDonch, line 436: Please add a return type annotation.

datumaro/components/annotation.py
```diff
@@ -316,45 +404,60 @@ def _get_image(image):
...
+        """
+        Extracts a single-instance mask from the compiled mask.
+        """
+
         return self.instance_mask == instance_id
 
-    def lazy_extract(self, instance_id):
+    def lazy_extract(self, instance_id: int):
```
IRDonch, line 443: Add a return type annotation here too.

datumaro/components/annotation.py
```diff
@@ -266,46 +328,72 @@ def from_instance_masks(instance_masks,
...
         return __class__(class_mask=merged_class_mask,
             instance_mask=merged_instance_mask)
 
-    def __init__(self, class_mask=None, instance_mask=None):
+    CompiledMaskImage = np.ndarray # 2d of integers
+
+    def __init__(self,
+            class_mask: Union[None,
+                CompiledMaskImage,
+                Callable[..., CompiledMaskImage]
```
IRDonch, line 391: Shouldn't it be `Callable[[], CompiledMaskImage]`?

datumaro/components/annotation.py
```diff
@@ -493,7 +596,7 @@ def as_polygon(self):
             x, y + h
         ]
 
-    def iou(self, other):
+    def iou(self, other: _Shape) -> Union[float, Literal[-1]]:
```
IRDonch, line 599: The literal seems unnecessary, because `-1` is type-compatible with `float`.

datumaro/components/annotation.py
```diff
@@ -555,7 +658,7 @@ class Visibility(Enum):
 
     _type = AnnotationType.points
 
-    visibility: list = attrib(default=None)
+    visibility: List[bool] = attrib(default=None)
```
IRDonch, line 661: Should be `Optional` if the default is `None`.

datumaro/components/annotation.py
```diff
@@ -38,42 +49,47 @@ def type(self) -> AnnotationType:
...
+    dataset-wide metainfo like available labels, label colors,
+    label attributes etc.
+    """
+
+    attributes: Set[str] = attrib(
+        factory=set, validator=default_if_none(set), eq=False)
 
 @attrs
 class LabelCategories(Categories):
-    @attrs(repr_ns='LabelCategories')
```
IRDonch, line 49: Please remove the one in `PointsCategories`, as well.

IRDonch (changes requested):

datumaro/components/annotation.py
zhiltsov-max: Apparently, you're right about scoping. It's unfortunate, because I'm not sure I want such aliases in the global scope.

datumaro/components/annotation.py
zhiltsov-max: Done

datumaro/components/annotation.py
zhiltsov-max: Fixed

Issue #129, opened 2021-03-04
Support video import


zhiltsov-max:
Visual datasets are not restricted by images and often come from video frames. The ability to use video data is essential for such project as Datumaro.

Requirements:
- Ability to import a video as a data source
  - A way to split video into frames
  - A way to utilize a video as a single dataset entry (action recognition tasks etc.)
- More natural support of track annotations and other video-specific types of annotations
- Ability to export a dataset as a video

zhiltsov-max:
v1 design:

- For clients, who interested in just video reading - provide video reading options 
  - CLI / API levels: add video frames extractor, should work like other import-only formats
  - API level: add video reader, a class to read video by frames
  - Use `opencv` as the backend (would use `av`, but it has problems with GPL precompiled binaries)
- For clients, who need stable results, which are independent of library versions - provide video splitting options 
  - An instruction in docs and/or a command
  - Formats that use video frames either work with frame directories or provide such option.
- Issue a warning on addition of the video frames extractor in project. Recommend manual splitting for having stable results
- API level: `DatasetItem` still have either a video or a frame linked. No options to use both.
- Videos are not exported by recoding. When supposed by a format, only file copying is done.

PR #521, opened 2021-10-18
Video input support

zhiltsov-max:
### Summary

Related #129 

- Added video reading API
- Added tests using pytest

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/components/media.py
+++ b/datumaro/components/media.py
@@ -296,6 +296,14 @@ class Video(MediaElement, Iterable[VideoFrame]):
         self._iterator: Optional[_VideoFrameIterator] = None
         self._frame_size: Optional[Tuple[int, int]] = None
 
+        # We don't provide frame count unless we have a reliable source of
+        # this information.
+        # - Not all videos provide length / duration metainfo
+        # - We can get an estimation based on metadata, but it
+        #   can be invalid or inaccurate due to variable frame rate
+        #   or fractional values rounded up. Relying on the value will give
+        #   errors during requesting frames.
+        # https://stackoverflow.com/a/47796468
         self._frame_count = None
         self._length = None
 
@@ -332,17 +340,6 @@ class Video(MediaElement, Iterable[VideoFrame]):
         else:
             # Need to decode to iterate over frames
             yield from self._get_iterator()
-
-    @property
-    def est_frame_count(self) -> Optional[int]:
-        """
-        Returns estimated frame count in the video file.
-        """
-
-        est_count = self._frame_count
-        if est_count is None:
-            est_count = self._get_est_frame_count(self._get_reader())
-        return est_count
 
     @property
     def length(self) -> Optional[int]:
@@ -393,23 +390,6 @@ class Video(MediaElement, Iterable[VideoFrame]):
 
         return frame_size
 
-    @staticmethod
-    def _get_est_frame_count(reader) -> Optional[int]:
-        # We don't provide frame count unless we have a reliable source of
-        # this information.
-        # - Not all videos provide length / duration metainfo
-        # - We can get an estimation based on metadata, but it
-        #   can be invalid or inaccurate due to variable frame rate
-        #   or fractional values rounded up. Relying on the value will give
-        #   errors during requesting frames.
-        # https://stackoverflow.com/a/47796468
-
-        frame_count = reader.get(cv2.CAP_PROP_FRAME_COUNT)
-        if frame_count:
-            frame_count = int(frame_count)
-        else:
-            frame_count = None
-        return frame_count
 
     def _get_end_frame(self):
         if self._end_frame is not None and self._frame_count is not None:
--- a/tests/test_video.py
+++ b/tests/test_video.py
@@ -106,13 +106,11 @@ class VideoTest:
         on_exit_do(video.close)
 
         assert None == video.length
-        assert 4 == video.est_frame_count
 
         for idx, frame in enumerate(video):
             assert idx == frame.index
 
         assert 4 == video.length
-        assert 4 == video.est_frame_count
 
     @mark_requirement(Requirements.DATUM_GENERAL_REQ)
     @scoped


datumaro/components/media.py
```diff
@@ -178,3 +179,255 @@ def save(self, path):
...
+
+        return frame_size
+
+    @staticmethod
+    def _get_frame_count(container) -> Optional[int]:
+        # Not all videos provide length / duration metainfo
+        # Note that this information can be invalid or inaccurate
+        # due to variable frame rate
+        # https://stackoverflow.com/a/47796468
+        video_length = container.get(cv2.CAP_PROP_FRAME_COUNT)
```
IRDonch, line 370:
```suggestion
        frame_count = container.get(cv2.CAP_PROP_FRAME_COUNT)
```

No reason to introduce a synonym.

datumaro/components/media.py
```diff
@@ -178,3 +179,255 @@ def save(self, path):
...
+        self._frame_size: Optional[Tuple[int, int]] = None
+
+        self._container = None
+        self._reset()
+
+        self._step = int(step)
+        self._start_frame = int(start_frame)
+        self._end_frame = int(end_frame) if end_frame else None
+
+        self._frame_count = self._get_frame_count(self._container)
```
IRDonch, line 298: Do both `self._end_frame` and `self._frame_count` need to exist? It seems like you don't really need both of them after the initialization is complete. I think the following initialization algorithm would be sufficient:

1. Try to fetch the total frame count for the video (I'll call this `frame_count`).
1. If `end_frame` is `None`, set `end_frame = frame_count`. Otherwise, verify that `frame_count is None or end_frame <= frame_count`.
1. If `end_frame` is not `None`, verify that `end_frame > start_frame`.
1. Save `self._start_frame = start_frame; self._end_frame = end_frame`. Now you can always use `self._end_frame` as the end of the segment.

datumaro/components/media.py
```diff
@@ -178,3 +179,255 @@ def save(self, path):
...
+            self._iterator = _VideoFrameIterator(self)
+        return self._iterator
+
+    def _get_container(self):
+        if self._container is None:
+            self._reset()
+        return self._container
+
+    def _open(self):
+        return cv2.VideoCapture(self._path)
```
IRDonch, line 418: This method seems kind of useless. It's only used once and its body is a single call.

datumaro/components/media.py
```diff
@@ -178,3 +179,255 @@ def save(self, path):
...
+            end_frame = self._get_end_frame()
+
+            length = None
+            if end_frame is not None:
+                length = (end_frame - self._start_frame) // self._step
+                assert 0 < length
+
+            self._length = length
+
+        return self._length
```
IRDonch, line 396: This seems like a good fit for a `functools.cached_property` (or maybe just a regular property - it only requires a few arithmetic operations to compute).

datumaro/util/image.py
```diff
@@ -199,28 +197,8 @@ def decode_image(image_bytes: bytes,
...
-        return False
-
-    for d, _, filenames in walk(dirpath,
-            max_depth=max_depth if recursive else 0):
-        for filename in filenames:
-            if not _check_image_ext(filename):
-                continue
-
-            yield osp.join(d, filename)
+    yield from find_files(dirpath, exts=exts or IMAGE_EXTENSIONS,
```
IRDonch, line 200:
```suggestion
    return find_files(dirpath, exts=exts or IMAGE_EXTENSIONS,
```

This will avoid creating a new generator for the `find_images` function call.

tests/test_video.py
```diff
@@ -0,0 +1,113 @@
...
+    """
+
+    writer = cv2.VideoWriter(path, frameSize=tuple(frame_size[::-1]),
+        fps=float(fps), fourcc=cv2.VideoWriter_fourcc(*'MJPG'))
+
+    for _ in range(frames):
+        # Apparently, only uint8 values are supported, but not floats
+        writer.write(np.ones((*frame_size, 3), dtype=np.uint8) * 255)
+
+    writer.release()
```
IRDonch, line 27: You should probably use `try`/`finally` to call `release`.

tests/test_video.py
```diff
@@ -0,0 +1,113 @@
...
+    @mark_requirement(Requirements.DATUM_GENERAL_REQ)
+    @scoped
+    def test_can_read_frames_sequentially(self, fxt_sample_video):
+        video = Video(fxt_sample_video)
+        on_exit_do(video.close)
+
+        for idx, frame in enumerate(video):
+            assert frame.size == video.frame_size
+            assert frame.index == idx
+            assert id(frame.video) == id(video)
```
IRDonch, line 56:
```suggestion
            assert frame.video is video
```

tests/test_video.py
```diff
@@ -0,0 +1,113 @@
...
+            assert frame == np.ones((*video.frame_size, 3)) * 255
```
IRDonch, line 57: How does this work? Normally, `ndarray` equality returns another `ndarray`, which is not convertible to a boolean.

tests/test_video.py
```diff
@@ -0,0 +1,113 @@
...
+    @mark_requirement(Requirements.DATUM_GENERAL_REQ)
+    @scoped
+    def test_can_read_frames_randomly(self, fxt_sample_video):
+        video = Video(fxt_sample_video)
+        on_exit_do(video.close)
+
+        for idx in {1, 3, 2, 0, 3}:
+            frame = video[idx]
+            assert frame.index == idx
+            assert frame == np.ones((*video.frame_size, 3)) * 255
```
IRDonch, line 68: This is not a great test, because it doesn't ensure that the returned image actually corresponds to the requested frame. IMO, you should change `make_sample_video` so that each frame is different (e.g. use the frame index as the color).

tests/test_video.py
```diff
@@ -0,0 +1,113 @@
...
+
+    @mark_requirement(Requirements.DATUM_GENERAL_REQ)
+    @scoped
+    def test_can_skip_frames_between(self, fxt_sample_video):
+        video = Video(fxt_sample_video, step=2)
+        on_exit_do(video.close)
+
+        assert 2 == video.frame_count
```
IRDonch, line 76: With all of these skipping tests, it would also be useful to check that the returned frames are actually the frames that are expected.

tests/test_video.py
```diff
@@ -0,0 +1,113 @@
...
+        for last_frame in video:
+            pass
+
+        assert 2 == video.frame_count
+        assert 1 == last_frame.index
+
+    @mark_requirement(Requirements.DATUM_GENERAL_REQ)
+    @scoped
+    def test_can_init_frame_count_lazily(self, fxt_sample_video):
+        with mock.patch.object(Video, '_get_frame_count',
```
IRDonch, line 103: IMO, it would be better to create an actual test video file whose frame count can't be determined in advance and use that:

1. It would serve as a proof that such a situation can actually occur in reality.
2. It would avoid the fragility that comes with mocking internal methods.

IRDonch:
FWIW, the refactoring changes that are in this PR could easily be submitted as a separate PR, which would make it easier to review.

tests/test_video.py
zhiltsov-max: A good question. Frame is an `Image`, and it has `__eq__` redefined. The reverse comparison will fail (unlike in C++, BTW). Do you think we shouldn't use it this way?

tests/test_video.py
zhiltsov-max: Good point. I tried this. Unfortunately, this doesn't work exactly this way, because video codecs encode colors, and there are only few colors that can be decoded without loss.

datumaro/components/media.py
zhiltsov-max: Unfortunately, we need to initialize `frame_count` lazily, because not all videos can provide frame count info. Honestly, I'm not even sure we should use it all, because the reported value can be different from the real frame count (because FPS can be variable and frame count is an estimation obtained as `fps * duration`), however it can be a significant performance issue if we try to decode all the frames just to count them. So the idea is that we just save the input values and defer the action you suggested to the `get_end_frame` method, which always checks if we already know the real frame count. The bad thing about this is that `Video` can report different frame count after we get the real value. A dedicated metafile describing frames could be useful here. I'll think what we can do here.

datumaro/components/media.py
```diff
@@ -178,3 +179,255 @@ def save(self, path):
...
+            self._reset()
+        return self._container
+
+    def _open(self):
+        return cv2.VideoCapture(self._path)
+
+    def _reset(self):
+        if self._container is not None:
+            self._container.release()
+        self._container = self._open()
```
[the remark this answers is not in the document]
zhiltsov-max: It was a term from pyav https://pyav.org/docs/develop/api/container.html. I'm ok with `_readed`, renamed. I don't want to depend on OpenCV naming here, because it is not very explanatory in this context.

datumaro/components/media.py
```diff
@@ -178,3 +179,255 @@ def save(self, path):
...
+                self._current_frame_data = frame.astype(float)
+                yield VideoFrame(self, index=self._pos)
+
+            success, frame = cap.read()
+
+        if self._video._frame_count is None:
+            self._video._frame_count = self._pos + 1
+
+    def __iter__(self):
+        return self
```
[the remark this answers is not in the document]
zhiltsov-max: Removed

datumaro/components/media.py
```diff
@@ -178,3 +179,255 @@ def save(self, path):
...
+        """
+
+        self._pos = -1
+
+        success, frame = cap.read()
+        while success:
+            self._pos += 1
+            if self._video._includes_frame(self._pos):
+                self._current_frame_data = frame.astype(float)
+                yield VideoFrame(self, index=self._pos)
```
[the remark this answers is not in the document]
zhiltsov-max: Yes, fixed.

datumaro/components/media.py
```diff
@@ -178,3 +179,255 @@ def save(self, path):
...
+class _VideoFrameIterator(Iterator[VideoFrame]):
+    """
+    Provides sequential access to the video frames.
+    """
+
+    def __init__(self, video: 'Video'):
+        self._video: 'Video' = video
+        self._iterator: Iterator[VideoFrame] = None
+        self._pos: int = -1
+        self._current_frame_data: Optional[np.ndarray] = None
```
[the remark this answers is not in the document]
zhiltsov-max: Done

datumaro/components/media.py
zhiltsov-max: Done

datumaro/components/media.py
zhiltsov-max: Removed

tests/test_video.py
zhiltsov-max: Fixed

tests/test_video.py
zhiltsov-max: I agree, but it's not simple to create such video with opencv.

IRDonch: I don't mean creating it with OpenCV; I mean creating it offline with whatever tool you can use and adding it to the repository.

tests/test_video.py
zhiltsov-max: Done, worked with grayscale colors.

tests/test_video.py
zhiltsov-max: Done

tests/test_video.py
IRDonch: Yeah, I'm not a fan of the `Image`<->`ndarray` comparison operator. Non-symmetric equality is pretty bad on its own, and here it's made worse by behaving inconsistently with `ndarray`'s own equality operator. I think this should be replaced by explicitly getting the frame data (which can be compared to the reference value using `np.testing.assert_array_equal`).

zhiltsov-max: Changed

datumaro/components/media.py
zhiltsov-max: The value is already cached. I moved the code to the existing property for this.

tests/test_video.py
zhiltsov-max: It looks like it [can happen](https://github.com/opencv/opencv/blob/4.x/modules/videoio/src/cap_ffmpeg_impl.hpp#L1633-L1686), but I failed to create a reproducing example. However, it is simple to obtain an incorrect value here, e.g. in case of variable FPS. Videos look to always have `duration` info, which is used to calculate frame count, but it can be rounded up, giving extra frames in computations. Thus, I'm not sure we should provide or use this info until we iterate over frames or have a reliable source of this info.

IRDonch: > It looks like it can happen, but I failed to create a reproducing example. However, it is simple to obtain an incorrect value here, e.g. in case of variable FPS. 

Okay, but can you produce an example of that? It could be a useful testcase too.

datumaro/components/media.py
```diff
@@ -178,3 +179,276 @@ def save(self, path):
...
+            # However, it can be inacurrate
+            end_frame = self._get_end_frame()
+            for index in range(self._start_frame, end_frame, self._step):
+                yield VideoFrame(video=self, index=index)
+        else:
+            # Need to decode to iterate over frames
+            yield from self._get_iterator()
+
+    @property
+    def est_frame_count(self) -> Optional[int]:
```
IRDonch, line 337: IMO, this should either be renamed to `est_total_frame_count` (with appropriate updates to the documentation), or just not provided at all. Otherwise, it's too similar to `length`. Personally, I would argue for not providing it, because the `Video` class essentially represents a _segment_ of a video, and the total length of the video shouldn't be relevant.

datumaro/components/media.py
IRDonch: Okay, fair enough.

datumaro/components/media.py
zhiltsov-max: I removed the function until we get more understanding on it's use.

tests/test_video.py
zhiltsov-max: I have a video, for which we have 2 extra frame in computations, however, I was not able to craft such example. Also, we can't just store a random video from internet in the repo for legal and storage reasons. We definitely know it happens sometimes and there are multiple evidence for it across internet. I think, we can just leave a comment in the code.

IRDonch (approved):

PR #536, opened 2021-11-03
Add more dataset format detectors

IRDonch:
### Summary

Add detectors for a few more dataset formats:

* `ade20k2017`
* `ade20k2020`
* `cvat`
* `datumaro`
* `icdar_text_segmentation`
* `icdar_word_recognition`
* `kitti_raw`
* `label_me`
* `mot_seq`
* `yolo`

To support them, add a mechanism for placing requirements on file contents.

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/components/format_detection.py
+++ b/datumaro/components/format_detection.py
@@ -126,22 +126,22 @@ class FormatDetectionContext:
         unspecified which one of them is returned.
         """
 
-        requirement_str = \
+        requirement_desc = \
             f"dataset must contain a file matching pattern \"{pattern}\""
 
         if not self._is_path_within_root(pattern):
-            self.fail(requirement_str)
+            self.fail(requirement_desc)
 
         pattern_abs = osp.join(glob.escape(self._root_path), pattern)
         for path in glob.iglob(pattern_abs, recursive=True):
             if osp.isfile(path):
                 return osp.relpath(path, self._root_path)
 
-        self.fail(requirement_str)
+        self.fail(requirement_desc)
 
     @contextlib.contextmanager
     def probe_text_file(
-        self, path: str, requirement_str: str,
+        self, path: str, requirement_desc: str,
     ) -> Iterator[TextIO]:
         """
         Returns a context manager that can be used to place a requirement on
@@ -163,14 +163,14 @@ class FormatDetectionContext:
         requirement being unmet, that exception is reraised and the new
         requirement is abandoned.
 
-        `requirement_str` must be a human-readable statement describing the
+        `requirement_desc` must be a human-readable statement describing the
         requirement.
         """
 
-        requirement_str_full = f"{path}: {requirement_str}"
+        requirement_desc_full = f"{path}: {requirement_desc}"
 
         if not self._is_path_within_root(path):
-            self.fail(requirement_str_full)
+            self.fail(requirement_desc_full)
 
         try:
             with open(osp.join(self._root_path, path), encoding='utf-8') as f:
@@ -178,7 +178,7 @@ class FormatDetectionContext:
         except FormatRequirementsUnmet:
             raise
         except Exception:
-            self.fail(requirement_str_full)
+            self.fail(requirement_desc_full)
 
 
 FormatDetector = Callable[
--- a/datumaro/components/format_detection.py
+++ b/datumaro/components/format_detection.py
@@ -3,7 +3,8 @@
 # SPDX-License-Identifier: MIT
 
 from enum import IntEnum
-from typing import Callable, Optional, Sequence
+from typing import Callable, Iterator, Optional, Sequence, TextIO
+import contextlib
 import glob
 import os.path as osp
 
@@ -84,6 +85,27 @@ class FormatDetectionContext:
         """
         return self._root_path
 
+    def _is_path_within_root(self, path: str) -> bool:
+        """
+        Checks that `path` is a relative path and does not attempt to leave
+        the dataset root by using `..` segments.
+
+        Requirement-placing methods that use this to verify their arguments
+        should raise a FormatRequirementsUnmet rather than a "hard" error like
+        AssertionError if False is returned. The reason is that the path passed
+        by the detector might not have been hardcoded, and instead might have
+        been acquired from another file in the dataset. In that case, an invalid
+        pattern signifies a problem with the dataset, not with the detector.
+        """
+        if osp.isabs(path) or osp.splitdrive(path)[0]:
+            return False
+
+        path = osp.normpath(path)
+        if path.startswith('..' + osp.sep):
+            return False
+
+        return True
+
     def fail(self, requirement: str) -> NoReturn:
         """
         Places a requirement that is never met. `requirement` must contain
@@ -107,15 +129,7 @@ class FormatDetectionContext:
         requirement_str = \
             f"dataset must contain a file matching pattern \"{pattern}\""
 
-        # These pattern checks raise a FormatRequirementsUnmet rather than an
-        # AssertionError, because the detector might have gotten the pattern
-        # from another file in the dataset. In that case, an invalid pattern
-        # signifies a problem with the dataset, rather than with the detector.
-        if osp.isabs(pattern) or osp.splitdrive(pattern)[0]:
-            self.fail(requirement_str)
-
-        pattern = osp.normpath(pattern)
-        if pattern.startswith('..' + osp.sep):
+        if not self._is_path_within_root(pattern):
             self.fail(requirement_str)
 
         pattern_abs = osp.join(glob.escape(self._root_path), pattern)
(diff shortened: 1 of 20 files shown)


datumaro/components/format_detection.py
```diff
@@ -84,6 +85,27 @@ def root_path(self) -> str:
...
+        """
+        Checks that `path` is a relative path and does not attempt to leave
+        the dataset root by using `..` segments.
+
+        Requirement-placing methods that use this to verify their arguments
+        should raise a FormatRequirementsUnmet rather than a "hard" error like
+        AssertionError if False is returned. The reason is that the path passed
+        by the detector might not have been hardcoded, and instead might have
+        been acquired from another file in the dataset. In that case, an invalid
+        pattern signifies a problem with the dataset, not with the detector.
```
zhiltsov-max, line 98: Maybe we need to raise invalid requirement error in such cases?

datumaro/components/format_detection.py
```diff
@@ -125,6 +139,48 @@ def require_file(self, pattern: str) -> str:
 
         self.fail(requirement_str)
 
+    @contextlib.contextmanager
+    def probe_text_file(
+        self, path: str, requirement_str: str,
```
zhiltsov-max, line 144:
```suggestion
        self, path: str, requirement_desc: str,
```

datumaro/plugins/ade20k2020_format.py
```diff
@@ -174,6 +175,19 @@ def _load_class_mask(path):
...
+        annot_path = context.require_file('*/**/*.json')
+
+        with context.probe_text_file(
+            annot_path, "must be a JSON object with an \"annotation\" key",
+        ) as f:
+            contents = json.load(f)
+            if not isinstance(contents, dict):
+                raise Exception
+            if 'annotation' not in contents:
+                raise Exception
```
zhiltsov-max, line 189: Looks pretty similar to `datumaro` and `coco`. How the conflict is resolved?

datumaro/plugins/cvat_format/extractor.py
```diff
@@ -315,6 +316,17 @@ def _load_items(self, parsed):
...
 class CvatImporter(Importer):
+    @classmethod
+    def detect(cls, context: FormatDetectionContext) -> None:
+        annot_file = context.require_file('*.xml')
+
+        with context.probe_text_file(
+            annot_file, "must be an XML file with an \"annotations\" root element",
+        ) as f:
+            _, root_elem = next(ElementTree.iterparse(f, events=('start',)))
+            if root_elem.tag != 'annotations':
```
zhiltsov-max, line 327: Looks like the same test would work for labelme.

datumaro/plugins/cvat_format/extractor.py
```diff
@@ -315,6 +316,17 @@ def _load_items(self, parsed):
         return parsed
 
 class CvatImporter(Importer):
+    @classmethod
+    def detect(cls, context: FormatDetectionContext) -> None:
+        annot_file = context.require_file('*.xml')
```
zhiltsov-max, line 321: Maybe, we can also check for the `/annotation/` dir.

datumaro/plugins/datumaro_format/extractor.py
```diff
@@ -190,6 +191,18 @@ def _load_annotations(item):
...
+    @classmethod
+    def detect(cls, context: FormatDetectionContext) -> None:
+        annot_file = context.require_file('annotations/*.json')
+
+        with context.probe_text_file(
+            annot_file, "must be a JSON object with \"categories\" "
+                "and \"items\" keys",
+        ) as f:
+            contents = json.load(f)
+            if not {'categories', 'items'} <= contents.keys():
```
zhiltsov-max, line 203: Maybe we should check there is no `info` field to avoid conflicting with coco.

tests/cli/test_revpath.py
```diff
@@ -111,12 +105,11 @@ def test_can_parse(self):
...
-                parse_full_revpath(dataset_url)
-            self.assertEqual(
-                {ProjectNotFoundError, MultipleFormatsMatchError},
-                set(type(e) for e in cm.exception.problems)
-            )
+            dataset, project = parse_full_revpath(dataset_url)
+            if project:
+                scope_add(project)
+            self.assertTrue(isinstance(dataset, IDataset))
+            self.assertEqual(None, project)
```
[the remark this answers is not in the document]
IRDonch: I changed it, because there are no errors anymore. Since the Datumaro format can now be uniquely identified, the dataset is successfully loaded despite no format being specified.

datumaro/plugins/cvat_format/extractor.py
IRDonch: But there is no such directory in the CVAT format.

datumaro/plugins/cvat_format/extractor.py
IRDonch: Unfortunately, no, because it would be insufficient to distinguish it from VOC.

datumaro/plugins/ade20k2020_format.py
IRDonch: In this format, the key is `annotation`. In `coco`, it's `annotations`.

In Datumaro it's neither, so I don't think it's even similar. 🙂

datumaro/plugins/datumaro_format/extractor.py
IRDonch: > Maybe we should check there is no info field to avoid conflicting with coco.

I don't think there is a conflict, otherwise the COCO detection test would have failed (since there is currently no medium-confidence detector for COCO, it would have gotten detected as Datumaro instead).

> Think if we can use https://github.com/ICRAR/ijson to avoid storing big jsons in memory.

Interesting... I'll investigate it.

datumaro/components/format_detection.py
IRDonch: What do you mean by that? A new exception type? How would it be handled?

datumaro/components/format_detection.py
IRDonch: Yeah, that would be more descriptive. Updated.

datumaro/components/format_detection.py
```diff
@@ -125,6 +139,48 @@ def require_file(self, pattern: str) -> str:
...
+    ) -> Iterator[TextIO]:
+        """
+        Returns a context manager that can be used to place a requirement on
+        the contents of the file referred to by `path`. To do so, you must
+        enter and exit this context manager (typically, by using a `with`
```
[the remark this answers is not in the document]
IRDonch: I don't think my version was incorrect (a `with` statement is just a specific instance of the `with` statement), but "the" works too. Updated.

datumaro/components/format_detection.py
zhiltsov-max: Raising `FormatRequirementsUnmet` doesn't seem correct when incorrect requirements are specified. But I agree that the whole detection process shouldn't be interrupted in such cases. So, probably, just another error can be introduced. I suppose, we can come up with:
`DatasetDetectionError` 
^- `FormatRequirementsUnmet` 
^- `InvalidRequirement`

And just catch and return `DatasetDetectionError`s.

datumaro/plugins/cvat_format/extractor.py
zhiltsov-max: It's not specified anywhere (like in COCO and Datumaro), but I don't see reasons to exclude this option. We decided some time ago we'll support the same layout with `images` and `annotations` dirs in such cases.

tests/cli/test_revpath.py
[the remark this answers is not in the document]
zhiltsov-max: Try to add a test for the previously existing situation.

datumaro/plugins/cvat_format/extractor.py
IRDonch: But I don't think the extractor even supports this layout. It only looks for an `images` directory [in the same directory as the XML file](https://github.com/openvinotoolkit/datumaro/blob/v0.2/datumaro/plugins/cvat_format/extractor.py#L24).

And as far as I can see from CVAT's code, it never creates an `annotations` directory either.

zhiltsov-max: Okay, maybe we need to fix it everywhere then.

datumaro/components/format_detection.py
IRDonch: > So, probably, just another error can be introduced.

That will cause a problem later with the addition of alternatives. If one alternative raises a `FormatRequirementsUnmet`, and another raises an `InvalidRequirement`, then what exception should the detector as a whole raise?

Furthermore, I predict that "incorrect requirements" will, in practice, be caused by the dataset not meeting other format requirements (like in the scenario explained in the comment: when a value is read from a file that's supposed to be a path, but isn't), and therefore it's actually reasonable to report them as unmet requirements.

The only other possible cause that I can see would be an invalid requirement that is hardcoded into the detector, but such a detector will always fail, so an error like that would be caught in testing and corrected.

tests/cli/test_revpath.py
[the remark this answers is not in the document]
IRDonch: Okay, I added a new test using a more ambiguous dataset.

datumaro/components/format_detection.py
zhiltsov-max: > That will cause a problem later with the addition of alternatives. If one alternative raises a `FormatRequirementsUnmet`, and another raises an `InvalidRequirement`, then what exception should the detector as a whole raise?

I don't think I understand the way you want to implement alternatives, but unless they are implemented, the question makes no sense. From my perspective, we will iterate over alternatives and then return a list of errors. If there is a matching one, the confidence is returned. 

Maybe, if there are incorrect requirements, we need to stop checking the format at all. Print a debug log message / a warning, return the NONE confidence. I expect such situations to happen only in custom plugins, or during development. So maybe, we can just fail, actually.

tests/cli/test_revpath.py
```diff
@@ -125,6 +126,36 @@ def test_can_parse(self):
             self.assertTrue(isinstance(dataset, IDataset))
             self.assertEqual(None, project)
 
+    @mark_requirement(Requirements.DATUM_GENERAL_REQ)
+    @scoped
+    def test_ambiguous_format(self):
+        test_dir = scope_add(TestDir())
+
+        # this dataset can be loaded as either imagenet or image_dir
```
zhiltsov-max, line 134: I don't think it is a blocker now, but I'm not sure if this should be true. `ImageDir` should be a fallback format, when no others match. Maybe we can have a json file instead (eg. with `annotation` and `annotations` fields), or add a couple of test formats in this test?

IRDonch: > ImageDir should be a fallback format, used when no others match.

I'm not sure I agree, but we can discuss that later. For now, I've changed the test to use a dataset that is _definitely_ ambiguous.

PR #539, opened 2021-11-06
Generic media support

zhiltsov-max:
### Summary

Depends on #538
Related #675 

- Added `DatasetItem.media` to replace dedicated members for each media type
- Added the `PointCloud` media type
- Added the `media_type()` method to `Extractor`s
- Added merging for all media types, mixed media types for an item or in the dataset produce an error
- Datasets can't have mixed media types in items. If such situation occurs, an error is raised (checked during dataset caching/iteration)
- Datasets can't change media type using transforms
- Extractors must report their media type with the `media_type()` method
- Added a new mandatory `media_type` argument to `Dataset.from_iterable`. It has a default value of `Image` for the transition period (to be tracked in #675).
- Deprecated `DatasetItem.image`, `.related_images`, `.point_cloud`, `save-images` and `require_images`
- Added deprecation messages about annotation classes in `components.extractor`
- Suppressed Datumaro deprecation messages when using Datumaro from CLI

TODO:
- [x] `--save-media` in CLI
- <s> [ ] Media type deduction by path (do we need it at all?) </s>
- [x] Update formats to the new interface
- Questions with Dataset domain 
  - [x] Is it better to differentiate datasets by domain or media type? **By media type**
  - [x] Dataset items must use the same media type? **Yes**
  - [x] How to avoid calling transforms with unsupported media types? Transforms and converters must check input media type
  - [x] How to avoid exporting in formats for datasets with unsupported media types?

### Checklist

### License

```python
#
```


datumaro/components/extractor.py
```diff
@@ -21,65 +24,132 @@
...
+    }:
+        warnings.warn(f"Using {name} from {__package__} is deprecated and "
+            "will be removed in future. The class is moved to "
+            "'datumaro.components.annotation'",
+            DeprecationWarning, stacklevel=2)
+
+        import datumaro.components.annotation as annotation
+        return getattr(annotation, name)
+    raise AttributeError(f"module {__name__} has no attribute {name}")
+
```
zhiltsov-max, line 46:
```suggestion
```

Removed in develop

datumaro/cli/contexts/project/diff.py
```diff
@@ -243,11 +243,11 @@ def save_item_bbox_diff(self, item_a, item_b, diff):
         _, mispred, a_unmatched, b_unmatched = diff
 
         if 0 < len(a_unmatched) + len(b_unmatched) + len(mispred):
-            if not item_a.has_image or not item_a.image.has_data:
+            if not item_a.media:
```
zhiltsov-max, line 246: I think, we need to have a check somewhere in this class for media type and use `media_as()` for type deduction

datumaro/components/dataset.py
```diff
@@ -956,9 +956,11 @@ def import_from(cls, path: str, format: Optional[str] = None, *,
         for src_conf in detected_sources:
             if not isinstance(src_conf, Source):
                 src_conf = Source(src_conf)
-            extractors.append(env.make_extractor(
+            extractor = env.make_extractor(
                 src_conf.format, src_conf.url, **src_conf.options
-            ))
+            )
+            cls.check_media_type(extractor)
```
zhiltsov-max, line 962: Probably, this call is better be moved to `init_cache()`.

datumaro/components/dataset.py
```diff
@@ -1003,6 +1005,12 @@ def detect(path: str, *,
         if 1 < len(matches):
             raise MultipleFormatsMatchError(matches)
 
+    @classmethod
+    def check_media_type(cls, source: IDataset):
```
zhiltsov-max, line 1009:
```suggestion
    def _check_media_type(cls, source: IDataset):
```
No reasons to make it public.

datumaro/components/dataset_filter.py
```diff
@@ -21,8 +22,8 @@ def encode(cls, item, categories=None):
         ET.SubElement(item_elem, 'id').text = str(item.id)
         ET.SubElement(item_elem, 'subset').text = str(item.subset)
 
-        image = item.image
-        if image is not None:
+        image = item.media
+        if image and isinstance(image, Image):
```
zhiltsov-max, line 26:
```suggestion
        if isinstance(image, Image):
```
The check is excessive, `None` is not an `Image`

datumaro/components/extractor.py
```diff
@@ -5,11 +5,14 @@
...
-from typing import Any, Callable, Dict, Iterator, List, Optional
+from typing import (
+    Any, Callable, Dict, Iterator, List, Optional, Type, TypeVar, Union, cast,
+)
 import os
 import os.path as osp
+import warnings
 
-from attr import attrs, field
+from attr import attrib, attrs, field
```
zhiltsov-max, line 15:
```suggestion
from attr import attrs, field
```

datumaro/components/extractor.py
```diff
@@ -170,6 +241,9 @@ def get(self, id, subset=None):
                 return item
         return None
 
+    def media_type(self):
```
zhiltsov-max, line 244: Probably, it would be reasonable to add to `IExtractor` also.

datumaro/plugins/datumaro_format/extractor.py
```diff
@@ -98,30 +98,39 @@ def _load_items(self, parsed):
...
                         image_path = old_image_path
 
-                image = Image(path=image_path, size=image_info.get('size'))
+                media = Image(path=image_path, size=image_info.get('size'))
+                self._media_type = Image
 
-            point_cloud = None
             pcd_info = item_desc.get('point_cloud')
-            if pcd_info:
+            if not media and pcd_info:
```
zhiltsov-max, line 105: I think, multiple media records should be considered an error.

datumaro/plugins/kitti_raw_format/converter.py
```diff
@@ -414,16 +414,17 @@ def _write_item(self, item, index):
         if not self._reindex:
             index = cast(item.attributes.get('frame'), int, index)
 
-        if self._save_images:
-            if item.has_point_cloud:
+        if self._save_media:
```
zhiltsov-max, line 417:
```suggestion
        if self._save_media and item.media:
```

zhiltsov-max, line 417:
```suggestion
        if self._save_media and item.media:
```

datumaro/plugins/kitti_raw_format/converter.py
```diff
@@ -414,16 +414,17 @@ def _write_item(self, item, index):
...
-                        item.id + self._find_image_ext(image)))
+            if item.media:
+                images = sorted(item.media.extra_images, key=lambda img: img.path)
+                for i, image in enumerate(images):
+                    if image.has_data:
+                        image.save(osp.join(self._save_dir,
+                            KittiRawPath.IMG_DIR_PREFIX + ('%02d' % i), 'data',
+                            item.id + self._find_image_ext(image)))
 
         else:
```
zhiltsov-max, line 429:
```suggestion
        elif self._save_media and not item.media:
```

datumaro/plugins/kitti_raw_format/extractor.py
```diff
@@ -245,9 +246,8 @@ def _load_items(self, parsed):
         for frame_id, item_desc in parsed.items():
             name = name_mapping.get(frame_id, '%010d' % int(frame_id))
             items[frame_id] = DatasetItem(id=name, subset=self._subset,
-                point_cloud=osp.join(self._rootdir,
-                    KittiRawPath.PCD_DIR, name + '.pcd'),
-                related_images=sorted(images.get(name, [])),
+                media=PointCloud(osp.join(self._rootdir, KittiRawPath.PCD_DIR,
+                    name + '.pcd'), extra_images=sorted(images.get(name, []))),
```
zhiltsov-max, line 250:
```suggestion
                media=PointCloud(
                    osp.join(self._rootdir, KittiRawPath.PCD_DIR, name + '.pcd'), 
                    extra_images=sorted(images.get(name, []))),
```

datumaro/plugins/kitti_raw_format/extractor.py
```diff
@@ -256,9 +256,8 @@ def _load_items(self, parsed):
                 continue
 
             items[frame_id] = DatasetItem(id=name, subset=self._subset,
-                point_cloud=osp.join(self._rootdir,
-                    KittiRawPath.PCD_DIR, name + '.pcd'),
-                related_images=sorted(images.get(name, [])),
+                media=PointCloud(osp.join(self._rootdir, KittiRawPath.PCD_DIR,
+                    name + '.pcd'), extra_images=sorted(images.get(name, []))),
```
zhiltsov-max, line 260:
```suggestion
                media=PointCloud(
                    osp.join(self._rootdir, KittiRawPath.PCD_DIR, name + '.pcd'), 
                    extra_images=sorted(images.get(name, []))),
```

datumaro/plugins/mots_format.py
```diff
@@ -77,9 +78,12 @@ def _parse_items(self):
 
         for p in sorted(iglob(self._anno_dir + '/**/*.png', recursive=True)):
             item_id = osp.splitext(osp.relpath(p, self._anno_dir))[0]
+            image = None
+            image_path = images.get(item_id)
+            if image_path:
+                image = Image(path=image_path)
```
zhiltsov-max, line 84:
```suggestion
            image = images.get(item_id)
            if image:
                image = Image(path=image)
```
Seems like a pretty common pattern

datumaro/plugins/transforms.py
```diff
@@ -50,9 +50,9 @@ def transform_item(self, item):
         if not segments:
             return item
 
-        if not item.has_image:
+        if not item.media:
```
zhiltsov-max, line 53: As for converters, I think, transforms should check input media type and fail, if it is not supported.

datumaro/util/test_utils.py
```diff
@@ -148,10 +149,11 @@ def compare_datasets(test, expected: IDataset, actual: IDataset,
         elif not ignored_attrs:
             test.assertEqual(item_a.attributes, item_b.attributes, item_a.id)
 
-        if (require_images and item_a.has_image and item_a.image.has_data) or \
-                item_a.has_image and item_a.image.has_data and \
-                item_b.has_image and item_b.image.has_data:
-            test.assertEqual(item_a.image, item_b.image, item_a.id)
+        if require_images and item_a.media and item_b.media:
```
zhiltsov-max, line 152: Consider replacing `require_images` with `require_media`

datumaro/components/extractor.py
yasakova-anastasia: Removed

datumaro/components/converter.py
```diff
@@ -71,13 +73,13 @@ def patch(cls, dataset, patch, save_dir, **options):
...
 
-    def __init__(self, extractor, save_dir, save_images=False,
+    def __init__(self, extractor, save_dir, save_images=False, save_media=False,
             image_ext=None, default_image_ext=None, save_dataset_meta=False):
         default_image_ext = default_image_ext or self.DEFAULT_IMAGE_EXT
         assert default_image_ext
         self._default_image_ext = default_image_ext
 
-        self._save_images = save_images
+        self._save_media = save_media or save_images
```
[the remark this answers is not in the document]
yasakova-anastasia: Added a deprecation warning

datumaro/components/dataset.py
yasakova-anastasia: Moved to `init_cache()`

datumaro/components/dataset.py
yasakova-anastasia: Fixed

datumaro/components/dataset_filter.py
yasakova-anastasia: Fixed

datumaro/components/extractor.py
```diff
@@ -21,65 +24,132 @@
...
-        factory=list, validator=default_if_none(list))
-    subset: str = field(converter=lambda v: v or DEFAULT_SUBSET_NAME,
-        default=None)
 
-    # TODO: introduce "media" field with type info. Replace image and pcd.
-    image: Optional[Image] = field(default=None)
-    # TODO: introduce pcd type like Image
-    point_cloud: Optional[str] = field(
-        converter=lambda x: str(x).replace('\\', '/') if x else None,
+    subset: str = attrib(converter=lambda v: v or DEFAULT_SUBSET_NAME,
```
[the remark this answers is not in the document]
yasakova-anastasia: Fixed

datumaro/components/extractor.py
yasakova-anastasia: Fixed

datumaro/components/extractor.py
yasakova-anastasia: Added this to `IExtractor`

datumaro/components/media.py
```diff
@@ -481,3 +482,19 @@ def __eq__(self, other: object) -> bool:
     def __hash__(self):
         # Required for caching
         return hash((self._path, self._step, self._start_frame, self._end_frame))
+
+class PointCloud(MediaElement):
+    def __init__(self, path: str,
+            extra_images: Optional[List[Union[
+                str, Image, np.ndarray, Callable[[str], np.ndarray]
```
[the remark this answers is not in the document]
yasakova-anastasia: I agree with you. Removed other types.

datumaro/plugins/kitti_raw_format/converter.py
yasakova-anastasia: Fixed

datumaro/plugins/kitti_raw_format/converter.py
yasakova-anastasia: Fixed

datumaro/plugins/kitti_raw_format/extractor.py
yasakova-anastasia: Fixed

datumaro/plugins/kitti_raw_format/extractor.py
yasakova-anastasia: Fixed

datumaro/plugins/mots_format.py
yasakova-anastasia: Fixed it in all formats

tests/cli/test_diff.py
```diff
@@ -30,7 +30,8 @@ def test_can_compare_projects(self): # just a smoke test
             point_categories1.add(index, ['cat1', 'cat2'], joints=[[0, 1]])
 
         dataset1 = Dataset.from_iterable([
-            DatasetItem(id=100, subset='train', image=np.ones((10, 6, 3)),
+            DatasetItem(id=100, subset='train',
+            media=Image(data=np.ones((10, 6, 3))),
```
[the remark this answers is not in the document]
yasakova-anastasia: Fixed

datumaro/cli/contexts/project/diff.py
yasakova-anastasia: Added check for media type. But I didn't understand why we need to use `media_as() `there.

datumaro/components/operations.py
```diff
@@ -116,7 +116,7 @@ def merge(cls, *sources: IDataset) -> DatasetItemStorage:
     def _merge_items(cls, existing_item: DatasetItem,
             current_item: DatasetItem) -> DatasetItem:
         return existing_item.wrap(
-            image=cls._merge_images(existing_item, current_item),
+            media=cls._merge_images(existing_item, current_item),
```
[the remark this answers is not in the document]
yasakova-anastasia: Added merge for all media types

datumaro/plugins/datumaro_format/extractor.py
yasakova-anastasia: Added an error

datumaro/components/dataset.py
zhiltsov-max: The current variant also is not good because it does full dataset traversal, which can be a performance issue. Please try to integrate it into `_iter_init_cache`.

datumaro/plugins/labelme_format.py
```diff
@@ -356,8 +356,8 @@ def _save_item(self, item, subset_dir):
         log.debug("Converting item '%s'", item.id)
 
         image_filename = self._make_image_filename(item)
-        if self._save_images:
-            if item.has_image and item.image.has_data:
+        if self._save_media:
+            if item.media and item.media.has_data:
```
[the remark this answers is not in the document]
yasakova-anastasia: Added checks for media type in converters

datumaro/components/dataset.py
```diff
@@ -956,9 +964,10 @@ def import_from(cls, path: str, format: Optional[str] = None, *,
         for src_conf in detected_sources:
             if not isinstance(src_conf, Source):
                 src_conf = Source(src_conf)
-            extractors.append(env.make_extractor(
+            extractor = env.make_extractor(
                 src_conf.format, src_conf.url, **src_conf.options
-            ))
+            )
+            extractors.append(extractor)
```
zhiltsov-max, line 970: Looks like unnecessary change

datumaro/components/extractor.py
```diff
@@ -21,65 +24,116 @@
...
+from datumaro.components.media import Image, MediaElement, PointCloud
 from datumaro.util import is_method_redefined
 from datumaro.util.attrs_util import default_if_none, not_empty
 
 DEFAULT_SUBSET_NAME = 'default'
 
-@attrs(slots=True, order=False)
+T = TypeVar('T', bound=MediaElement)
+
+@attrs(order=False, init=False)
```
zhiltsov-max, line 35:
```suggestion
@attrs(order=False, init=False, slots=True)
```

datumaro/components/media.py
```diff
@@ -481,3 +482,16 @@ def __eq__(self, other: object) -> bool:
...
+        self._path = path
+
+        self.extra_images: List[Image] = []
+        for image in extra_images or []:
+            assert isinstance(image, (str, Image, np.ndarray)) or callable(image)
+            if isinstance(image, str):
+                image = Image(path=image)
+            elif isinstance(image, np.ndarray) or callable(image):
+                image = Image(data=image)
+            self.extra_images.append(image)
```
zhiltsov-max, line 497:
```suggestion
        self.extra_images: List[Image] = extra_images
```

datumaro/components/operations.py
```diff
@@ -145,14 +149,44 @@ def _merge_attrs(a: Dict[str, Any], b: Dict[str, Any],
...
         return merged
 
-    @staticmethod
-    def _merge_images(item_a: DatasetItem, item_b: DatasetItem) -> Image:
-        image = None
+    @classmethod
+    def _merge_media(cls, item_a: DatasetItem, item_b: DatasetItem) \
+            -> Union[ByteImage, Image, PointCloud, Video, VideoFrame]:
+        if isinstance(item_a.media, (ByteImage, Image, VideoFrame)) or \
+                isinstance(item_b.media, (ByteImage, Image, VideoFrame)):
```
zhiltsov-max, line 156:
```suggestion
        if isinstance(item_a.media, Image) or \
                isinstance(item_b.media, Image):
```
They all are Image subclasses

datumaro/components/operations.py
```diff
@@ -145,14 +149,44 @@ def _merge_attrs(a: Dict[str, Any], b: Dict[str, Any],
...
+            media = cls._merge_images(item_a, item_b)
+        else:
```
zhiltsov-max, line 158: I think, it will be better to have a separate block for each case, even if it produces some code duplication.

datumaro/components/operations.py
```diff
@@ -145,14 +149,44 @@ def _merge_attrs(a: Dict[str, Any], b: Dict[str, Any],
...
+            if isinstance(item_a.media, PointCloud) and isinstance(item_b.media, PointCloud) or \
+                    isinstance(item_a.media, MediaElement) and isinstance(item_b.media, MediaElement) or \
+                    isinstance(item_a.media, Video) and isinstance(item_b.media, Video):
+                if item_a.media.path and item_b.media.path and \
```
zhiltsov-max, line 162: Video needs to be compared for start and stop frames, Point cloud have to be merged on extra images

datumaro/plugins/camvid_format.py
```diff
@@ -293,7 +295,9 @@ def apply(self):
             segm_list = {}
             for item in subset:
                 image_path = self._make_image_filename(item, subdir=subset_name)
-                if self._save_images:
+                if self._save_media and item.media:
+                    if not isinstance(item.media, (ByteImage, Image)):
```
zhiltsov-max, line 299:
```suggestion
                    if not isinstance(item.media, Image):
```

datumaro/components/dataset.py
yasakova-anastasia: Moved to `_iter_init_cache`

datumaro/plugins/labelme_format.py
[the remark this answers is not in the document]
yasakova-anastasia: Fixed

datumaro/util/test_utils.py
yasakova-anastasia: Replaced `require_images` with `require_media`

datumaro/components/dataset.py
yasakova-anastasia: Fixed

datumaro/components/extractor.py
yasakova-anastasia: Fixed

datumaro/components/media.py
yasakova-anastasia: Fixed

datumaro/components/operations.py
yasakova-anastasia: Fixed

datumaro/components/operations.py
yasakova-anastasia: Divided into blocks by type

datumaro/components/operations.py
yasakova-anastasia: Fixed. 

But I'm not sure that I understand correctly how to merge point cloud and video.

datumaro/plugins/camvid_format.py
yasakova-anastasia: Fixed

datumaro/components/operations.py
zhiltsov-max: I don't think we can merge video, because if there are different frame constraints, the videos are different. It will be reasonable to throw an error in this case.

For point clouds we can append extra images, if they are not specified, and we can append pcd path, if it's unknown.

datumaro/components/dataset.py
```diff
@@ -383,6 +381,11 @@ def _update_status(item_id, new_status: ItemStatus):
 
         i = -1
         for i, item in enumerate(source):
+            if self.media_type():
+                if item.media and not isinstance(item.media, self.media_type()):
+                    raise MediaTypeError("Dataset elements must have a '%s' " \
+                        "media type" % self.media_type())
```
zhiltsov-max, line 387: Probably, it is reasonable to deduce and cache this info during processing, if it was not specified on creation.

datumaro/components/operations.py
```diff
@@ -145,14 +149,43 @@ def _merge_attrs(a: Dict[str, Any], b: Dict[str, Any],
...
+
+                if item_a.media.path:
+                    media = item_a.media
+                else:
+                    media = item_b.media
+
+            elif isinstance(item_a.media, MediaElement):
+                media = item_a.media
+            else:
+                media = item_b.media
```
zhiltsov-max, line 177: This block is only valid if one of items doesn't have media info. If both items have media, but the types are different, we definitely need to raise an error.

datumaro/components/operations.py
```diff
@@ -145,14 +149,43 @@ def _merge_attrs(a: Dict[str, Any], b: Dict[str, Any],
 
         return merged
 
+    @classmethod
+    def _merge_media(cls, item_a: DatasetItem, item_b: DatasetItem) \
+            -> Union[ByteImage, Image, PointCloud, Video, VideoFrame]:
```
zhiltsov-max, line 154:
```suggestion
            -> Union[Image, PointCloud, Video]:
```

datumaro/cli/contexts/project/diff.py
```diff
@@ -243,11 +245,11 @@ def save_item_bbox_diff(self, item_a, item_b, diff):
         _, mispred, a_unmatched, b_unmatched = diff
 
         if 0 < len(a_unmatched) + len(b_unmatched) + len(mispred):
-            if not item_a.has_image or not item_a.image.has_data:
+            if not item_a.media or not isinstance(item_a.media, Image):
```
zhiltsov-max, line 248:
```suggestion
            if not isinstance(item_a.media, Image) or not item_a.media.has_data:
```

datumaro/components/extractor.py
```diff
@@ -106,10 +160,14 @@ def categories(self) -> CategoriesInfo:
     def get(self, id, subset=None) -> Optional[DatasetItem]:
         raise NotImplementedError()
 
+    def media_type(self):
+        return None
+
 class _ExtractorBase(IExtractor):
-    def __init__(self, length=None, subsets=None):
+    def __init__(self, length=None, subsets=None, media_type=Image):
```
zhiltsov-max, line 167: It is questionable if such default is reasonable. I can't say it is totally wrong, but it at least needs justification.

datumaro/plugins/labelme_format.py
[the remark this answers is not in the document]
zhiltsov-max: How it was fixed? I don't see input dataset media type checks anywhere.

datumaro/cli/contexts/project/diff.py
yasakova-anastasia: Fixed

datumaro/components/operations.py
yasakova-anastasia: Fixed

datumaro/components/operations.py
yasakova-anastasia: Added an error for this

datumaro/components/operations.py
yasakova-anastasia: I think I fixed it

datumaro/components/extractor.py
zhiltsov-max: I can see it a reasonable solution for the transition period, but then I'd like to see `None` here and even request this info from the extractor. Please think about the transition plan and document it somewhere. Maybe, create an issue to track this.

datumaro/components/converter.py
```diff
@@ -147,7 +151,12 @@ def __init__(self, extractor: IExtractor, save_dir: str, *,
         assert default_image_ext
         self._default_image_ext = default_image_ext
 
-        self._save_images = save_images
+        if save_images:
+            warnings.warn("'save-images' is deprecated and will be "
+                "removed in future. Use 'save-media' instead.",
+                DeprecationWarning, stacklevel=2)
+
+        self._save_media = save_media or save_images
```
zhiltsov-max:
```suggestion
        self._save_media = save_media if save_media is not None else save_images
```

datumaro/cli/contexts/project/diff.py
```diff
@@ -243,11 +245,11 @@ def save_item_bbox_diff(self, item_a, item_b, diff):
         _, mispred, a_unmatched, b_unmatched = diff
 
         if 0 < len(a_unmatched) + len(b_unmatched) + len(mispred):
-            if not item_a.has_image or not item_a.image.has_data:
+            if not isinstance(item_a.media, Image) or not item_a.media.has_data:
```
zhiltsov-max, line 251: Probably, it would be reasonable to add the same check for media type as in converters.

datumaro/components/dataset.py
```diff
@@ -636,7 +643,8 @@ def from_iterable(cls, iterable: Iterable[DatasetItem],
         class _extractor(Extractor):
             def __init__(self):
                 super().__init__(length=len(iterable) \
-                    if hasattr(iterable, '__len__') else None)
+                    if hasattr(iterable, '__len__') else None,
```
zhiltsov-max:
```suggestion
                        if hasattr(iterable, '__len__') else None,
```

datumaro/components/errors.py
```diff
@@ -332,6 +345,14 @@ def __str__(self):
             "for ann %s, votes %s, sources %s" % \
             (self.item_id, self.ann, self.votes, self.sources)
 
+@define(auto_exc=False)
+class VideoMergeError(DatasetMergeError):
+    item_id = field()
+
+    def __str__(self):
+        return "Item %s: video merging is not possible" % \
+            self.item_id
```
zhiltsov-max:
```suggestion
            (self.item_id, )
```

datumaro/components/extractor.py
```diff
@@ -114,10 +166,14 @@ def categories(self) -> CategoriesInfo:
     def get(self, id, subset=None) -> Optional[DatasetItem]:
         raise NotImplementedError()
 
+    def media_type(self):
+        return None
```
zhiltsov-max:
```suggestion
        raise NotImplementedError()
```

datumaro/components/extractor.py
```diff
@@ -114,10 +166,14 @@ def categories(self) -> CategoriesInfo:
...
+
 class _ExtractorBase(IExtractor):
-    def __init__(self, *, length=None, subsets=None):
+    def __init__(self, *, length=None, subsets=None, media_type=Image):
```
zhiltsov-max:
```suggestion
    def __init__(self, *, length=None, subsets=None, media_type=None):
```
I don't think this class should define default for this. For transforms it will be different, because they need to reuse parent value by default. For extractors it will work, but it will be better to define it there.

datumaro/components/extractor.py
```diff
@@ -166,7 +222,7 @@ def __iter__(_):
             def categories(_):
                 return self.categories()
```
zhiltsov-max, line 245:
```suggestion
            def media_type(_):
                return self.media_type()
```

datumaro/components/extractor.py
```diff
@@ -166,7 +222,7 @@ def __iter__(_):
...
 
-        return _DatasetFilter()
+        return _DatasetFilter(media_type=self.media_type())
```
zhiltsov-max:
```suggestion
        return _DatasetFilter()
```

datumaro/components/extractor.py
```diff
@@ -280,6 +343,9 @@ def get(self, id, subset=None):
         assert subset == self._subset, '%s != %s' % (subset, self._subset)
         return super().get(id, subset or self._subset)
 
+    def media_type(self):
+        return self._media_type
+
```
zhiltsov-max:
```suggestion
```
It is the same definition as in the base class

datumaro/components/extractor.py
```diff
@@ -178,6 +234,9 @@ def get(self, id, subset=None):
                 return item
         return None
 
+    def media_type(self):
```
zhiltsov-max:
```suggestion
    def media_type(self) -> Optional[Type[MediaElement]]:
```

datumaro/components/operations.py
```diff
@@ -163,45 +201,91 @@ def _merge_images(item_a: DatasetItem, item_b: DatasetItem) -> Image:
...
+            media = item_b.media
+
+        return media
+
+    @staticmethod
+    def _merge_videos(item_a: DatasetItem, item_b: DatasetItem) -> Video:
+        media = None
+
+        if isinstance(item_a.media, Video) and isinstance(item_b.media, Video):
+            raise VideoMergeError(item_a.id)
```
zhiltsov-max: Does it mean we can never merge video datasets? I shouldn't be so. We can merge videos if they have the same paths and frame options.

datumaro/plugins/transforms.py
```diff
@@ -53,9 +53,11 @@ def transform_item(self, item):
...
```
zhiltsov-max:
```suggestion
        if not isinstance(item.media, Image):
```
Please adjust other cases as well

datumaro/plugins/transforms.py
```diff
@@ -53,9 +53,11 @@ def transform_item(self, item):
...
             raise Exception("Image info is required for this transform")
-        h, w = item.image.size
+        if not isinstance(item.media, Image):
+            raise MediaTypeError("Item %s: media type is not an image")
```
zhiltsov-max:
```suggestion
```

zhiltsov-max:
Will you add any tests?

datumaro/components/converter.py
yasakova-anastasia: In which case `save_media is None`?

datumaro/cli/contexts/project/diff.py
yasakova-anastasia: Should we abort the save with an error if we have something other than an image? I think the check that we have now is enough.

datumaro/components/converter.py
```diff
@@ -201,14 +210,14 @@ def _save_image(self, item, path=None, *,
             self._make_image_filename(item, name=name, subdir=subdir))
         path = osp.abspath(path)
 
-        item.image.save(path)
+        item.media.save(path)
 
     def _save_point_cloud(self, item=None, path=None, *,
             name=None, subdir=None, basedir=None):
```
sizov-kirill, line 230: I think it's better restrict the media type of the input item in this function (or add warning), because now this function can potentially save not only `point_cloud` item but and `Image` as well.

datumaro/components/converter.py
zhiltsov-max: Hm, I didn't notice it was set to False by default. I see a problem here that we can't check if any or both args were used.  Probably, it should just fall back to save_images by default and throw an error if both args are supplied.

datumaro/cli/contexts/project/diff.py
zhiltsov-max: Currently, this class only works for images, which was essential previously. I think, we can just avoid calling image-specific parts, or disallow class use entirely and support it later. It is a good candidate into the list of necessary changes for this feature.

datumaro/components/hl_ops.py
```diff
@@ -99,8 +100,13 @@ def merge(*datasets: IDataset) -> IDataset:
     """
 
     categories = ExactMerge.merge_categories(d.categories() for d in datasets)
+    media_type = datasets[0].media_type()
+    for d in datasets:
+        if d.media_type() is not media_type:
+            raise MediaTypeError("Datasets have different media types")
```
zhiltsov-max: It would be better to put this into `ExactMerge.merge`

datumaro/components/operations.py
```diff
@@ -279,7 +279,12 @@ def _merge_videos(item_a: DatasetItem, item_b: DatasetItem) -> Video:
         media = None
 
         if isinstance(item_a.media, Video) and isinstance(item_b.media, Video):
-            raise VideoMergeError(item_a.id)
+            if item_a.media.path is not item_b.media.path or \
+                    item_a.media.length() is not item_b.media.length() or \
+                    item_a.media.frame_size() is not item_b.media.frame_size():
```
zhiltsov-max: I think, for video we should check path, start, end frame and step instead. Using length and frame size is not reliable and doesn't really give us any information. Probably, we should make `start`, `end` frame and `step` read-only properties in the `Video` class. Please do this in a separate PR.

kirill-sizov (commented):
Now [these](https://github.com/openvinotoolkit/datumaro/blob/develop/site/content/en/docs/user-manual/media_formats.md?plain=1#L51-L61) approaches are not alternative, because if user using `video_frames` format it's not possible to export such dataset in other `Image` media type format.  I think we should to mention it in docs or allow `VideoFrame` media type for `Image` formats.

zhiltsov-max:
@kirill-sizov, VideoFrame is subclass of Image, so it should be working.

datumaro/components/converter.py
yasakova-anastasia: Added a check for media type in this function

datumaro/components/hl_ops.py
yasakova-anastasia: Moved this to `ExactMerge.merge_media_types`

datumaro/cli/contexts/project/diff.py
yasakova-anastasia: Added a check before this function

datumaro/components/converter.py
yasakova-anastasia: Added an error if both args are supplied

sizov-kirill:
> @kirill-sizov, VideoFrame is subclass of Image, so it should be working.

@zhiltsov-max, I meant that for example such code throws an error:

```python

dataset = Dataset.import_from('video.mp4', format='video_frames')
dataset.export('./save_dir', 'voc', save_media=True)
```

```
datumaro.components.errors.MediaTypeError: Media type is not an image
```

datumaro/plugins/cifar_format.py
```diff
@@ -178,6 +183,10 @@ class CifarConverter(Converter):
     DEFAULT_IMAGE_EXT = '.png'
 
     def apply(self):
+        if self._extractor.media_type() and \
+                self._extractor.media_type() is not Image:
```
zhiltsov-max:
```suggestion
                not issubclass(self._extractor.media_type(), Image):
```
This is more correct way of comparison and it should fix the problem Kirill's highlighted.

datumaro/components/hl_ops.py
zhiltsov-max: Why not into `merge`? It would be an essential check for this operation.

datumaro/components/converter.py
zhiltsov-max: As I said - let's make `None` the default value. Otherwise this check can't know what the user has passed to the function. The error is when both are not `None`.

datumaro/components/hl_ops.py
yasakova-anastasia: Because need to give `media_type` in `DatasetItemStorageDatasetView`. Could this be a problem?

datumaro/plugins/cifar_format.py
yasakova-anastasia: Fixed

datumaro/components/converter.py
yasakova-anastasia: Done

PR #550, opened 2021-11-15
Add detection support for 4 more formats

IRDonch:
### Summary
The formats in question are `lfw`, `open_images`, `vgg_face2` and `widerface`. As supporting infrastructure, add alternative requirement support.

Also, add low-level format detection tests, which I should really have done earlier.

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/components/format_detection.py
+++ b/datumaro/components/format_detection.py
@@ -121,14 +121,14 @@ class FormatDetectionContext:
             f"a requirement ({req_type}) can't be placed directly within " \
             "a 'require_any' block"
 
-    def fail(self, requirement: str) -> NoReturn:
+    def fail(self, requirement_desc: str) -> NoReturn:
         """
-        Places a requirement that is never met. `requirement` must contain
+        Places a requirement that is never met. `requirement_desc` must contain
         a human-readable description of the requirement.
         """
         self._start_requirement("fail")
 
-        raise FormatRequirementsUnmet((requirement,))
+        raise FormatRequirementsUnmet((requirement_desc,))
 
     def require_file(self, pattern: str) -> str:
         """
--- a/datumaro/components/format_detection.py
+++ b/datumaro/components/format_detection.py
@@ -70,7 +70,7 @@ class FormatDetectionContext:
     met, the return value depends on the method.
     """
 
-    class _AlternativesContext:
+    class _OneOrMoreContext:
         failed_alternatives: List[str]
         had_successful_alternatives: bool
 
@@ -78,13 +78,13 @@ class FormatDetectionContext:
             self.failed_alternatives = []
             self.had_successful_alternatives = False
 
-    # This points to an _AlternativesContext when and only when the detector
-    # is directly within an `alternatives` block.
-    _alternatives_context: Optional[_AlternativesContext]
+    # This points to a `_OneOrMoreContext` when and only when the detector
+    # is directly within a `require_any` block.
+    _one_or_more_context: Optional[_OneOrMoreContext]
 
     def __init__(self, root_path: str) -> None:
         self._root_path = root_path
-        self._alternatives_context = None
+        self._one_or_more_context = None
 
     @property
     def root_path(self) -> str:
@@ -117,9 +117,9 @@ class FormatDetectionContext:
         return True
 
     def _start_requirement(self, req_type: str) -> None:
-        assert not self._alternatives_context, \
+        assert not self._one_or_more_context, \
             f"a requirement ({req_type}) can't be placed directly within " \
-            "an 'alternatives' block"
+            "a 'require_any' block"
 
     def fail(self, requirement: str) -> NoReturn:
         """
@@ -202,7 +202,7 @@ class FormatDetectionContext:
             self.fail(requirement_desc_full)
 
     @contextlib.contextmanager
-    def alternatives(self) -> Iterator[None]:
+    def require_any(self) -> Iterator[None]:
         """
         Returns a context manager that can be used to place a requirement that
         is considered met if at least one of several alternative sets of
@@ -211,7 +211,7 @@ class FormatDetectionContext:
         requirements represented as nested `with` statements using the context
         manager returned by `alternative`:
 
-            with context.alternatives():
+            with context.require_any():
                 with context.alternative():
                     # place requirements from alternative set 1 here
                 with context.alternative():
@@ -222,58 +222,58 @@ class FormatDetectionContext:
         executed, even if an alternative that is met is found early.
 
         Requirements must not be placed directly within a
-        `with context.alternatives()` block.
+        `with context.require_any()` block.
         """
 
-        self._start_requirement("alternatives")
+        self._start_requirement("require_any")
 
-        self._alternatives_context = self._AlternativesContext()
+        self._one_or_more_context = self._OneOrMoreContext()
 
         try:
             yield
 
             # If at least one `alternative` block succeeded,
-            # then the `alternatives` block succeeds.
-            if self._alternatives_context.had_successful_alternatives:
+            # then the `require_any` block succeeds.
+            if self._one_or_more_context.had_successful_alternatives:
                 return
 
             # If no alternatives succeeded, and none failed, then there were
             # no alternatives at all.
-            assert self._alternatives_context.failed_alternatives, \
-                "an 'alternatives' block must contain " \
+            assert self._one_or_more_context.failed_alternatives, \
+                "a 'require_any' block must contain " \
                 "at least one 'alternative' block"
 
             raise FormatRequirementsUnmet(
-                self._alternatives_context.failed_alternatives)
+                self._one_or_more_context.failed_alternatives)
         finally:
-            self._alternatives_context = None
+            self._one_or_more_context = None
 
     @contextlib.contextmanager
     def alternative(self) -> Iterator[None]:
         """
         Returns a context manager that can be used in combination with
-        `alternatives` to define alternative requirements. See the documentation
-        for `alternatives` for more details.
+        `require_any` to define alternative requirements. See the
+        documentation for `require_any` for more details.
 
         Must only be used directly within a `with context.requirements()` block.
         """
 
-        assert self._alternatives_context, \
+        assert self._one_or_more_context, \
             "An 'alternative' block must be directly within " \
-            "an 'alternatives' block"
+            "a 'require_any' block"
 
-        saved_alternatives_context = self._alternatives_context
-        self._alternatives_context = None
+        saved_one_or_more_context = self._one_or_more_context
+        self._one_or_more_context = None
 
         try:
             yield
         except FormatRequirementsUnmet as e:
-            saved_alternatives_context.failed_alternatives.extend(
+            saved_one_or_more_context.failed_alternatives.extend(
                 e.failed_alternatives)
         else:
-            saved_alternatives_context.had_successful_alternatives = True
+            saved_one_or_more_context.had_successful_alternatives = True
         finally:
-            self._alternatives_context = saved_alternatives_context
+            self._one_or_more_context = saved_one_or_more_context
 
 
 FormatDetector = Callable[
--- a/datumaro/components/format_detection.py
+++ b/datumaro/components/format_detection.py
@@ -118,7 +118,7 @@ class FormatDetectionContext:
 
     def _start_requirement(self) -> None:
         assert not self._alternatives_context, \
-            "checks can't be done directly within `alternatives` blocks"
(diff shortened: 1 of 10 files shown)


datumaro/components/format_detection.py
```diff
@@ -180,6 +200,80 @@ def probe_text_file(
...
+                ...
+
+        The contents of all `with context.alternative()` blocks will be
+        executed, even if an alternative that is met is found early.
+
+        Requirements must not be placed directly within a
+        `with context.alternatives()` block.
+        """
+
+        self._start_requirement()
```
zhiltsov-max, line 227: Probably, this will produce a confusing error, if alternatives are used twice in a row.

datumaro/components/format_detection.py
```diff
@@ -180,6 +200,80 @@ def probe_text_file(
...
+            # If no alternatives succeeded, and none failed, then there were
+            # no alternatives at all.
+            assert self._alternatives_context.failed_alternatives, \
+                "an `alternatives` block must contain " \
+                "at least one `alternative` block"
+
+            raise FormatRequirementsUnmet(
+                self._alternatives_context.failed_alternatives)
+        finally:
+            self._alternatives_context = None
```
zhiltsov-max, line 248: I'm not sure we need to allow having more than 1 alternatives section, and having both alternatives and plain reqs.

tests/test_format_detection.py
```diff
@@ -0,0 +1,169 @@
...
+class FormatDetectionTest(TestCase):
+    def setUp(self) -> None:
+        test_dir_context = TestDir()
+        self._dataset_root = test_dir_context.__enter__()
+        self.addCleanup(test_dir_context.__exit__, None, None, None)
+
+    @mark_requirement(Requirements.DATUM_GENERAL_REQ)
+    def test_empty_detector(self):
+        result = apply_format_detector(self._dataset_root, lambda c: None)
+        self.assertEqual(result, FormatDetectionConfidence.MEDIUM)
```
zhiltsov-max, line 21: Do you think we need to check if context was used at all?

tests/test_format_detection.py
```diff
@@ -0,0 +1,169 @@
...
+        result = apply_format_detector(self._dataset_root,
+            lambda c: FormatDetectionConfidence.LOW)
+        self.assertEqual(result, FormatDetectionConfidence.LOW)
+
+    @mark_requirement(Requirements.DATUM_GENERAL_REQ)
+    def test_root_path(self):
+        provided_root: str
+        def detect(context):
+            nonlocal provided_root
+            provided_root = context.root_path
```
zhiltsov-max, line 34: Consider using a mock.

IRDonch: You mean as the detector? I don't see a great way to check the `context.root_path` attribute if I do that. It would be possible to retrieve the context from the mock _after_ the call to `apply_format_detector` (and then get `root_path` from it), but the format detection API implies that the context should not be accessed outside the detector function, and I would rather not have the tests violate that.

zhiltsov-max: I thought you could use `.assert_called_*` and check input args, but probably you're right about inaccessible object.

datumaro/components/format_detection.py
IRDonch: What do you mean? This?

```python
with context.alternatives():
   ...

with context.alternatives():
   ...
```

That shouldn't result in any error. (And it checks that at least one alternative from the first block succeeds _and_ at least one alternative from the second block succeeds.)

datumaro/components/format_detection.py
IRDonch: Why not? An `alternatives` block is just another type of requirement (just made out of other requirements). This allows detectors to combine requirements in arbitrary ways (both logical conjunctions and disjunctions).

datumaro/components/format_detection.py
zhiltsov-max: No, I meant this:
```python
with context.alternatives():
  with context.alternatives():
    ...
```
This is going to produce `"checks can't be done directly within `alternatives` blocks"`, which is pretty misleading.

tests/test_format_detection.py
IRDonch: Are you suggesting to add this check to `apply_format_detector`?

I don't think that would be useful in practice. A detector that does not use the context will result in a detection on any directory, which is weird, but not really invalid. If such a detector is mistakenly added to one of the built-in plugins, that will be caught by existing tests (since the existing formats will no longer be unambiguous). And if a user writes a custom plugin with an empty detector, well... why not let them? Maybe they don't want to bother writing a real detector because they are only going to work with their custom format anyway.

datumaro/components/format_detection.py
```diff
@@ -106,11 +116,17 @@ def _is_path_within_root(self, path: str) -> bool:
 
         return True
 
+    def _start_requirement(self) -> None:
+        assert not self._alternatives_context, \
+            "checks can't be done directly within `alternatives` blocks"
```
zhiltsov-max, line 121:
```suggestion
            "checks can't be done directly within 'alternatives' blocks"
```

datumaro/plugins/open_images_format.py
```diff
@@ -552,15 +553,25 @@ def _load_and_resize_mask(path, size):
...
+        '*' + OpenImagesPath.MASK_DESCRIPTION_FILE_SUFFIX,
+    )
+
+    @classmethod
+    def detect(cls, context: FormatDetectionContext) -> None:
+        with context.alternatives():
+            for pattern in cls.POSSIBLE_ANNOTATION_PATTERNS:
+                with context.alternative():
+                    context.require_file(
+                        f'{OpenImagesPath.ANNOTATIONS_DIR}/{pattern}')
```
zhiltsov-max, line 570: Is there a way to return confidence for a matched alternative?

tests/test_format_detection.py
IRDonch: > I thought you could use .assert_called_* and check input args

Yeah, that would definitely not work - the argument is a context object, which is not comparable.

I was thinking more along the lines of `self.assertEqual(detector.call_args[0].root_path, self._dataset_root)`. But, to reiterate, I don't like that this smuggles the context out of the detector, since by the semantics it's only supposed to be valid while in the detector.

tests/test_format_detection.py
zhiltsov-max: I not insist on having such check, but I'm trying to guess possible reasons to allow such situations. If a user don't want to implement a detector, he just doesn't define the class method. If has empty implementation, it is unlikely to be working in practice. However, probably, it can be useful in some situations, where the user just don't want to use the context and just returns confidence.

datumaro/components/format_detection.py
zhiltsov-max: It is questionable if we can benefit from denormalized requirements on formats. It looks like it will be hard to implement different confidence values for different alternatives this way, but I can be wrong about this.

datumaro/plugins/open_images_format.py
IRDonch: Do you mean to return different confidences for different alternatives?

I hadn't really considered this before (I have not encountered any cases where this would be necessary), but I think you can arrange the alternatives from most confident to least confident and then add `return <confidence>` to each.

datumaro/components/format_detection.py
IRDonch: Sure; updated.

datumaro/components/format_detection.py
IRDonch: Well... `alternatives` _is_ a check, but I see what you mean. I'll think about how I can improve this.

IRDonch: Oh, I just realized that I've mixed up my terminology in these messages (I changed "checks" to "requirements" when I went from prototype to real implementation). I'll definitely have to fix that.

datumaro/components/format_detection.py
IRDonch: What do you mean by "denormalized"?

tests/test_format_detection.py
IRDonch: > I not insist on having such check, but I'm trying to guess possible reasons to allow such situations.

FWIW, my main reasons are:
1. It doesn't break the contract.
2. It would take extra work to forbid it, and the benefit seems miniscule.

datumaro/components/format_detection.py
zhiltsov-max: > What do you mean by "denormalized"?

The opposite of https://en.wikipedia.org/wiki/Disjunctive_normal_form, i.e. ability to represent conditions like "A and (B or C) ..." and not just "(A and B) or (A and C)". 

Probably, now I understood the model you're trying to implement. I want to say that maybe it is too powerful for our task and thus complicated for the user. It is an open question, which we're unlikely to answer now.

datumaro/plugins/open_images_format.py
zhiltsov-max: Yes, I thought it is what was meant by "returning different confidences for different alternatives". Maybe, the word "alternatives" doesn't fit exactly to the solution implemented, because it implies that the variants are mutually-exclusive.

IRDonch: > Maybe, the word "alternatives" doesn't fit exactly to the solution implemented, because it implies that the variants are mutually-exclusive.

That's a fair point. After thinking about it, I decided to rename `FormatDetectionContext.alternatives` to `require_one_or_more`, which is a better name for several reasons:

* more clear what it does;
* consistent with the "requirement" terminology;
* visually distinct from `alternative`.

I kept `alternative` as it was, though, because I don't know what else to call it.

datumaro/components/format_detection.py
IRDonch: Okay, I tuned the message and added the type of the requirement that was at fault. That should be clear enough.

datumaro/components/format_detection.py
IRDonch: > The opposite of https://en.wikipedia.org/wiki/Disjunctive_normal_form, i.e. ability to represent conditions like "A and (B or C) ..." and not just "(A and B) or (A and C)".

There are at least a couple reasons to allow that:

* it doesn't forcing the user to duplicate the code for A;
* it prevents evaluating A twice.

datumaro/components/format_detection.py
```diff
@@ -106,11 +116,18 @@ def _is_path_within_root(self, path: str) -> bool:
...
+            "a 'require_one_or_more' block"
+
     def fail(self, requirement: str) -> NoReturn:
         """
         Places a requirement that is never met. `requirement` must contain
         a human-readable description of the requirement.
         """
+        self._start_requirement("fail")
+
         raise FormatRequirementsUnmet((requirement,))
```
zhiltsov-max, line 131:
```suggestion
    def fail(self, message: str) -> NoReturn:
        """
        Places a requirement that is never met. `requirement` must contain
        a human-readable description of the requirement.
        """
        self._start_requirement("fail")

        raise FormatRequirementsUnmet((message,))
```

datumaro/components/format_detection.py
```diff
@@ -116,16 +116,17 @@ def _is_path_within_root(self, path: str) -> bool:
...
+            f"a requirement ({req_type}) can't be placed directly within " \
+            "an 'alternatives' block"
 
     def fail(self, requirement: str) -> NoReturn:
         """
         Places a requirement that is never met. `requirement` must contain
         a human-readable description of the requirement.
         """
-        self._start_requirement()
+        self._start_requirement("fail")
```
zhiltsov-max, line 129: "a requirement ('fail') can't be placed directly within an 'alternatives' block"

Still sound strange, IMO. Not a blocker, anyway.

datumaro/components/format_detection.py
zhiltsov-max: Ok, let's keep it that way.

datumaro/plugins/open_images_format.py
zhiltsov-max: How about `require_any`?

IRDonch: > How about require_any?

Works for me. Renamed.

> Consider adding a confidence parameter to alternative().

I'll consider it when I actually need it.

datumaro/components/format_detection.py
IRDonch: This sort of value is called `requirement_desc` elsewhere, so I renamed it to that instead.

zhiltsov-max (approved):

datumaro/components/format_detection.py
IRDonch: 🤷‍♂️ It's a message for developers. I think they can figure it out.

PR #568, opened 2021-11-26
Remove the `Image`<->`np.ndarray` equality comparison support

IRDonch:
### Summary

This support has some questionable semantics:

* It's not symmetric (`Image` can be equal to `ndarray`, but not the other way around).

* It doesn't behave the same as `ndarray` comparison (which returns an `ndarray` rather than `bool`), so it's not quite transitive either (if `image == array1` and `image == array2` are true, then `array1 == array2` is not true (but an array of `True`s).

Moreover, it's seemingly only used in two tests, and those tests can easily use `np.testing.assert_array_equal` instead, which has the added benefit of printing detailed mismatch information.

Inspired by the discussion in #521.

### Checklist

### License

```python
#
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   (<https://github.com/openvinotoolkit/datumaro/pull/538>)
 
 ### Removed
-- TBD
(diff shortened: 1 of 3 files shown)


zhiltsov-max:
I'm not sure I'm totally agree with removing underused functionality, but the semantics point looks more or less significant. But I see it more like a Python design problem.

PR #576, opened 2021-12-08
Add a dedicated command to detect a dataset's format

IRDonch:
### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/components/format_detection.py
+++ b/datumaro/components/format_detection.py
@@ -4,12 +4,13 @@
 
 from enum import IntEnum
 from typing import (
-    Callable, Collection, Iterator, List, NoReturn, Optional, Sequence, TextIO,
-    Union,
+    Callable, Collection, Iterable, Iterator, List, NoReturn, Optional,
+    Sequence, TextIO, Tuple, Union,
 )
 import contextlib
 import fnmatch
 import glob
+import logging as log
 import os.path as osp
 
 
@@ -58,6 +59,22 @@ class FormatRequirementsUnmet(Exception):
     def __init__(self, failed_alternatives: Sequence[str]) -> None:
         assert failed_alternatives
         self.failed_alternatives = tuple(failed_alternatives)
+
+    def generate_message(self) -> Sequence[str]:
+        """
+        Generate a human-readable description of the exception as a sequence
+        of text lines.
+        """
+        lines = []
+
+        if len(self.failed_alternatives) > 1:
+            lines.append("None of the following requirements were met:")
+        else:
+            lines.append("The following requirement was not met:")
+
+        lines.extend('  ' + req for req in self.failed_alternatives)
+
+        return lines
 
 class FormatDetectionContext:
     """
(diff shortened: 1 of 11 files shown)


datumaro/components/format_detection.py
```diff
@@ -331,3 +350,79 @@ def apply_format_detector(
...
+    """
+    Determines which format(s) the dataset at the specified path belongs to.
+
+    `detectors` is a list of (name, detector) tuples. The function applies all
+    of the detectors to the path, and returns all names for which the
+    corresponding detector matched. If a detector returns matches with less
+    confidence than another detector, then the corresponding name is omitted.
+
+    For formats that are rejected, `rejection_callback`, if not `None`,
+    is called to report the reason why.
```
zhiltsov-max, line 379:
```suggestion
    Args:
        `detectors` (a list of (name, detector) tuples) - The list of 
            format detectors to be checked. The function applies 
            all of the detectors to the path given, and returns all names 
            for which the corresponding detector matched. If a detector 
            returns matches with less confidence than another detector, 
            then the corresponding name is omitted

        `rejection_callback` (RejectionCallback) - Unless `None`, 
            called to report the reason why a format was rejected

    Returns: a list of best-matching format names
```

datumaro/components/format_detection.py
```diff
@@ -331,3 +350,79 @@ def apply_format_detector(
...
+
+    max_confidence = 0
+    matches = []
+
+    for format_name, detector in detectors:
+        log.debug("Checking '%s' format...", format_name)
+        try:
+            new_confidence = apply_format_detector(path, detector)
+        except FormatRequirementsUnmet as cf:
+            human_message = cf.generate_message()
```
zhiltsov-max, line 405:
```suggestion
        except FormatRequirementsUnmet as e:
            human_message = e.generate_message()
```
- `cf` is not very clear acronym (check failure?), is suggest to use the regular naming
- If it is a regular exception, we can probably use `human_message = str(e)`

tests/cli/test_detect_format.py
```diff
@@ -0,0 +1,113 @@
...
+                run(self, 'detect-format', test_dir)
+
+            output = output_file.getvalue()
+
+            self.assertIn(Ade20k2017Importer.NAME, output)
+            self.assertIn(Ade20k2020Importer.NAME, output)
+
+    # Ideally, there should be a test for the case where no formats match,
+    # but currently that's impossible, because some low-confidence detectors
+    # always match.
```
zhiltsov-max, line 60: You can add such test on API level. I see, there are no API-level tests though for the new function, so maybe we should introduce them and move some tests there?

zhiltsov-max (approved):

datumaro/components/format_detection.py
IRDonch: > cf is not a very clear acronym (check failure?), I suggest to use the regular naming

Yeah, I don't even remember why I called it that. It's probably a reference to a name I replaced a long time ago. Renamed.

> If it is a regular exception, we can probably use human_message = str(e)

Okay, I did that. I also changed the type of `human_message` to `str`, since I couldn't find a compelling reason for it to be a list.

tests/cli/test_detect_format.py
IRDonch: Yeah, that's a good idea. I'll do that.

IRDonch: Well, as it turns out, API-level tests don't really help with this case, because there's no special logic on the API level for the "zero formats detected" case - that logic only exists in the CLI command. However, the API tests could still be useful, so I added them.

tests/test_format_detection.py
```diff
@@ -199,3 +201,50 @@ def detect(context):
...
+
+        rejected_formats = {}
+
+        def rejection_callback(format, reason, message):
+            rejected_formats[format] = (reason, message)
+
+        detected_datasets = detect_dataset_format(formats, self._dataset_root,
+            rejection_callback=rejection_callback)
+
+        self.assertEqual(detected_datasets, ["bbb", "eee"])
```
zhiltsov-max, line 231:
```suggestion
        self.assertEqual(set(detected_datasets), {"bbb", "eee"})
```

zhiltsov-max (approved):

tests/test_format_detection.py
IRDonch: Applied.

zhiltsov-max (approved):

PR #579, opened 2021-12-09
Enable a few more Pylint checks

IRDonch:
### Checklist

### License

```python
#
```

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -141,6 +141,7 @@ enable=
     W0221, # arguments-differ
     W0222, # signature-differs
     W0233, # non-parent-init-called
+    W0301, # unnecessary-semicolon
     W0311, # bad-indentation
     W0404, # reimported
     W0410, # misplaced-future
--- a/.pylintrc
+++ b/.pylintrc
@@ -54,6 +54,7 @@ disable=all
 # multiple time (only on the command line, not in the configuration file where
 # it should appear only once). See also the "--disable" option for examples.
 enable=
+    C0121, # singleton-comparison
     C0123, # unidiomatic-typecheck
     C0200, # consider-using-enumerate
     C0303, # trailing-whitespace
--- a/.pylintrc
+++ b/.pylintrc
@@ -59,6 +59,7 @@ enable=
     C0200, # consider-using-enumerate
     C0303, # trailing-whitespace
     C0304, # missing-final-newline
+    C0325, # superfluous-parens
 
     E0001, # syntax-error
     E0100, # init-is-generator
--- a/datumaro/plugins/align_celeba_format.py
+++ b/datumaro/plugins/align_celeba_format.py
@@ -55,7 +55,7 @@ class AlignCelebaExtractor(SourceExtractor):
         label_categories = self._categories[AnnotationType.label]
 
         labels_path = osp.join(root_dir, AlignCelebaPath.LABELS_FILE)
-        if (not osp.isfile(labels_path)):
+        if not osp.isfile(labels_path):
             raise DatasetImportError("File '%s': was not found" % labels_path)
 
         with open(labels_path, encoding='utf-8') as f:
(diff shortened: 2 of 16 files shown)


kirill-sizov (dismissed):
Looks good to me

.pylintrc
```diff
@@ -54,10 +54,12 @@ disable=all
 # multiple time (only on the command line, not in the configuration file where
 # it should appear only once). See also the "--disable" option for examples.
 enable=
+    C0121, # singleton-comparison
```
zhiltsov-max, line 57: How this check will help us writing better code?

IRDonch: > How this check will help us writing better code? I can agree about ==, because it can hiddenly use an overloaded operator instead of direct comparison.

Then it sounds like your issue is not with the check itself, but with the way I chose to resolve some of the warnings emitted by that check.

In my experience, it is rarely necessary to compare directly against `True` or `False` (one example where it _is_ appropriate is [here](https://github.com/openvinotoolkit/datumaro/blob/8e1203a3d83e3d6421317d1ebf55c8c6371a1a37/datumaro/util/image.py#L263), where it's needed to distinguish boolean values from other types). Usually, you instead want to check whether your object is a true or a false value, which is what the `bool` constructor does. I think the cases where I used `bool()` are like that.

For example, in `coco_format/converter.py`, the original code will set `use_masks` to `True` if `leader.attributes['is_crowd']` is 1, and to `False` if it's 0 or 2, which doesn't make a lot of sense to me. In my version, it will set `use_masks` to `True` if and only if `leader.attributes['is_crowd']` is a true value (boolean, numeric or any other kind).

In these kinds of scenarios we _want_ to use the overloaded `__bool__` operator, because it affects the truth value of the object.

zhiltsov-max: > For example, in coco_format/converter.py, the original code will set use_masks to True if leader.attributes['is_crowd'] is 1, and to False if it's 0 or 2, which doesn't make a lot of sense to me. In my version, it will set use_masks to True if and only if leader.attributes['is_crowd'] is a true value (boolean, numeric or any other kind).

In fact, we only expect it to be boolean or `None` here (and in other cases you've applied `bool()`), that's why `bool()` should produce the same results as `is` and `==`. From my POV, we have only 3 valid values we can use here and others should be ignored. In practice, the value can be of any type, though, so such values are just discarded now. Note that your variant will work only for boolean an numeric types, but for strings it can give unexpected results (e.g.: `bool('no') is True`). I'm not sure if we need to support casting to the right type for each attribute in all formats, though. We can either fail, skip or try to convert invalid values.

zhiltsov-max: > For example, in coco_format/converter.py, the original code will set use_masks to True if leader.attributes['is_crowd'] is 1, and to False if it's 0 or 2, which doesn't make a lot of sense to me. In my version, it will set use_masks to True if and only if leader.attributes['is_crowd'] is a true value (boolean, numeric or any other kind).

In fact, we only expect it to be boolean or `None` here (and in other cases you've applied `bool()`), that's why `bool()` should produce the same results as `is` and `==`. From my POV, we have only 3 valid values we can use here and others should be ignored. In practice, the value can be of any type, though, so such values are just discarded now. Note that your variant will work only for boolean an numeric types, but for strings it can give unexpected results (e.g.: `bool('no') is True`). I'm not sure if we need to support casting to the right type for each attribute in all formats, though. We can either fail, skip or try to convert invalid values.

IRDonch: > In practice, the value can be of any type, though, so such values are just discarded now.

That's incorrect, because `True` compares as equal to `1` and `1.0`, so those values would be considered true as well. My version of the code keeps that behavior, while also changing the behavior for numbers that are not 0 or 1 to be more sensible (in my opinion, anyway).

I'm not particularly attached to this logic, though. If the _intent_ has always been to treat non-boolean values as false, I can change the code to do an `is True` comparison instead.

In fact, I _will_ do that, and you tell me if that's not what you meant. :P

IRDonch: OK, updated.

yasakova-anastasia (approved):

zhiltsov-max (approved):

PR #581, opened 2021-12-10
Add image resize transform

### LINKED ISSUE #562: Add transform to change dataset image resolution
I annotated the dataset using cvat. But I found that the resolution of my image was too high. Is there any way to reduce the resolution of the marked image?

zhiltsov-max:
### Summary

Closes #562 

- Added rescale transform
- Added tests

### Checklist

### License

```python
#
```


datumaro/plugins/transforms.py
```diff
@@ -734,3 +736,82 @@ def transform_item(self, item):
...
+                    interpolation=cv2.INTER_NEAREST).astype(np.uint8)
+
+                if isinstance(ann, RleMask):
+                    rle = mask_tools.mask_to_rle(rescaled_mask)
+                    resized_annotations.append(ann.wrap(
+                        rle=mask_utils.frPyObjects(rle, *rle['size'])))
+                else:
+                    resized_annotations.append(ann.wrap(image=rescaled_mask))
+            else:
+                resized_annotations.append(ann)
```
IRDonch, line 813: IMO, unhandled annotation types should be rejected; otherwise, we can forget to add support for any future annotation types and they will be passed through unscaled.

datumaro/plugins/transforms.py
```diff
@@ -734,3 +736,82 @@ def transform_item(self, item):
...
+
+        h, w = item.image.size
+        xscale = self._width / float(w)
+        yscale = self._height / float(h)
+
+        if item.image.has_data:
+            method = cv2.INTER_LINEAR if (xscale * yscale) < 1 \
+                else cv2.INTER_LANCZOS4
+            image = item.image.data / 255.0
+            resized_image = cv2.resize(image, (0, 0),
```
IRDonch, line 778: I think it would be slightly better to pass in the destination dimensions and omit `fx` and `fy`:

1. It guarantees that the result will be of the exact desired dimensions (not affected by rounding weirdness).
2. You have to pass in the dimensions anyway, so might as well pass the correct ones.

datumaro/plugins/transforms.py
zhiltsov-max: Added

datumaro/plugins/transforms.py
zhiltsov-max: Added

IRDonch (approved):

PR #582, opened 2021-12-14
Implement public dataset downloading via TFDS

IRDonch:
### Summary

This adds a new `download` command, which downloads public datasets. In terms of syntax & semantics, it's comparable to the `convert` command, except that instead of a source directory and source format, it accepts a "dataset ID".

Currently, the downloading is done through an external library, TensorFlow Datasets. However, I expect that this might not always be the case: we might implement native downloading later, or add other download backends. Therefore, the dataset ID must begin with a namespace (currently, the only such namespace is `tfds:`) that signifies the download backend.

We could probably make use of the Datumaro plugin manager functionality to implement download backends, but I don't want to bother with it until there are at least two of them.

The way TFDS represents dataset items is much more flexible than the way Datumaro does it; an item can have features with arbitrary names and types. To map those features onto Datumaro's fixed annotation types, this code uses adapters (which are basically just sequences of predefined callbacks that each convert a certain feature or a metadata element into the Datumaro representation).

### How to test
Try the `download` command. :-)

### Checklist

### License

```python
#
```


datumaro/cli/commands/download.py
```diff
@@ -0,0 +1,101 @@
...
+
+    log.info("Downloading the dataset")
+    if TFDS_EXTRACTOR_AVAILABLE and args.dataset_id.startswith('tfds:'):
+        tfds_ds_name = args.dataset_id[5:]
+        if tfds_ds_name in AVAILABLE_TFDS_DATASETS:
+            extractor = make_tfds_extractor(tfds_ds_name)
+        else:
+            raise CliException(f"Unsupported TFDS dataset '{tfds_ds_name}'")
+    else:
+        raise CliException(f"Unknown dataset ID '{args.dataset_id}'")
```
zhiltsov-max, line 94: I think, in this case we should display plugin installation instructions.

datumaro/cli/commands/download.py
```diff
@@ -0,0 +1,101 @@
...
+        |n
+        Supported datasets: {}|n
+        |n
+        Supported output formats: {}|n
+        |n
+        Examples: TBD
+        """.format(available_datasets, ', '.join(builtin_writers)),
+        formatter_class=MultilineFormatter)
+
+    parser.add_argument('-i', '--dataset-id', required=True,
```
zhiltsov-max, line 44: What do you think about an option to display dataset info by dataset id?

IRDonch: What kind of info do you have in mind?

Either way, I think that as long as we only get datasets from TFDS, it would be more practical to link to the [TFDS catalog](https://www.tensorflow.org/datasets/catalog/overview) from the help text, which contains all the info one could want.

datumaro/cli/commands/download.py
IRDonch: What plugin?

IRDonch:
Tests are now implemented.

datumaro/cli/commands/download.py
zhiltsov-max: `pip install datumaro[tfds]`

datumaro/cli/commands/download.py
zhiltsov-max: Yeah, I think, it mostly covers what I was thinking about. Info like dataset domain, annotation types, subsets, classes, size etc.

datumaro/cli/commands/download.py
IRDonch: Okay, I added that.

datumaro/cli/commands/download.py
IRDonch: I linked it in the help message.

datumaro/cli/commands/download.py
```diff
@@ -0,0 +1,121 @@
...
+        - Download the VOC 2012 dataset, saving only the annotations in the COCO
+          format into a specific directory:|n
+        |s|s%(prog)s -i tfds:voc/2012 -f coco -o path/I/like/
+        """.format(available_datasets, ', '.join(builtin_writers)),
+        formatter_class=MultilineFormatter)
+
+    parser.add_argument('-i', '--dataset-id', required=True,
+        help="Which dataset to download")
+    parser.add_argument('-f', '--output-format', required=True,
+        help="Output format")
```
zhiltsov-max, line 61: What is the point of having this option is the command? From my point, it should be pretty straightforward that we download a dataset and save it in the same format. Why do we need this separation? I can see it as a shortcut for calling `download` + `convert` (which should not improve performance) and as a way to support loading from custom storages (which is questionable).

datumaro/cli/commands/download.py
```diff
@@ -0,0 +1,121 @@
...
+    parser.add_argument('-o', '--output-dir', dest='dst_dir',
+        help="Directory to save output (default: a subdir in the current one)")
```
zhiltsov-max, line 63: I see there is no option for manipulating intermediate cache directory for TFDS data, but it can be pretty useful in practice. Don't you think it will be better to just download the data as is or to provide an option like `cache-dir`?

datumaro/components/extractor_tfds.py
```diff
@@ -0,0 +1,280 @@
...
+
+@attrs(auto_attribs=True)
+class _SetImageFromImageFeature:
+    feature_name: str
+    filename_feature_name: Optional[str] = attrib(default=None)
+
+    def __call__(self, tfds_example: Any, item: DatasetItem) -> None:
+        if self.filename_feature_name:
+            filename = tfds_example[self.filename_feature_name].numpy(
+            ).decode('UTF-8')
```
zhiltsov-max, line 86:
```suggestion
            filename = tfds_example[self.filename_feature_name].numpy() \
                .decode('UTF-8')
```
The formatting is strange here.

zhiltsov-max:
I failed to download mnist dataset on windows behind proxy, unless I set the env variable. It is actual for `pip` too. Maybe, it should be documented somewhere, because it is not very typical for Windows users to set env vars manually.

Yes, I see new there is a user cache directory for TFDS.

datumaro/cli/commands/download.py
IRDonch: The original format of the dataset is lost once it goes through TFDS. By the time we get the data, it's transformed into a generic form, so it makes as much sense to save it in the original format as any other format.

In addition, TFDS contains a wide variety of datasets, and we might want to implement downloading datasets for which Datumaro itself has no support (or no write support). The only way we'd be able to save those is with conversion.

datumaro/cli/commands/download.py
IRDonch: The location can be customized with the `TFDS_DATA_DIR` environment variable, so I didn't think it was important enough to add an option for it.

datumaro/components/extractor_tfds.py
IRDonch: Yeah, I don't know what I was thinking. Fixed.

IRDonch:
> I failed to download mnist dataset on windows behind proxy, unless I set the env variable. It is actual for pip too. Maybe, it should be documented somewhere, because it is not very typical for Windows users to set env vars manually.

I guess that's fair, but where?

(Interestingly, TFDS itself doesn't seem to document it at all.)

datumaro/cli/commands/download.py
zhiltsov-max: > The original format of the dataset is lost once it goes through TFDS. By the time we get the data, it's transformed into a generic form, so it makes as much sense to save it in the original format as any other format.

I'd suggest to go from the task and not from the implementation. If I write
```bash
datum download mnist
```
I feel it's pretty reasonable for this command to work without specification of the output format (at least, if it is supported in DM). We know what dataset we're downloading, and TFDS datasets probably can be mapped to DM formats without ambiguity. Can we make this parameter optional by maintaining a table of default formats for tfds ids? Output conversion on the fly can be useful in some cases, but I don't understand, why it must be required.

zhiltsov-max:
> I guess that's fair, but where?

I'll vote for just a comment in the online command docs.

IRDonch:
> > I guess that's fair, but where?
> 
> I'll vote for just a comment in the online command docs.

Okay, I added one.

datumaro/cli/commands/download.py
IRDonch: > Can we make this parameter optional by maintaining a table of default formats for tfds ids?

I can do that, but keep in mind that conversion will still happen behind the scenes. It'll just be conversion to that default format.

> datum download mnist

Incidentally, one of the reasons why the command does _not_ look like that (you have to specify `tfds:mnist`) is that the output is not the original MNIST dataset, but MNIST as it is represented in TFDS. There may be some loss of information incurred by TFDS's own parsing/conversion code (for example, the original order of items is lost, because TFDS shuffles the items before saving them).

zhiltsov-max: > I can do that, but keep in mind that conversion will still happen behind the scenes. It'll just be conversion to that default format.

I understand, but in this variant the user is less distracted with an extra mandatory parameter. 

> output is not the original MNIST dataset, but MNIST as it is represented in TFDS

Sure, but it was not the point I wanted to highlight - all the same applies to `datum download tfds:mnist`. The key here is that the (default) output format should be predefined. One can argue if it should be tfrecord or just MNIST, I believe it should be MNIST, because (1) the TFDS dataset will still be available in the cache and (2) it allows to just start using the dataset immediately.

> for example, the original order of items is lost, because TFDS shuffles the items before saving them

Yes, I noticed that, is there an option to disable this?

IRDonch: Okay, I will add default output formats.

> Yes, I noticed that, is there an option to disable this?

Not on the client side. It's controlled by the dataset implementation (see <https://www.tensorflow.org/datasets/add_dataset#maintain_dataset_order>), and last time I checked, none of the built-in datasets disabled it.

IRDonch: Implemented.

datumaro/util/test_utils.py
```diff
@@ -263,3 +265,37 @@ def compare_dirs(test, expected: str, actual: str):
 def run_datum(test, *args, expected_code=0):
     from datumaro.cli.__main__ import main
     test.assertEqual(expected_code, main(args), str(args))
+
+@contextlib.contextmanager
+def mock_tfds_data(example=None):
```
zhiltsov-max, line 270: Probably, we need to place such functionality (mocks, fixtures, pytest-dependents etc.) in the `tests/` directory. `run_datum` is also a  good candidate for moving. This is not a blocker, but we should do it at some point.

datumaro/cli/commands/download.py
```diff
@@ -0,0 +1,125 @@
...
+
+    log.info("Downloading the dataset")
+    extractor = extractor_factory()
+
+    log.info("Exporting the dataset")
+    converter.convert(extractor, dst_dir,
+        default_image_ext='.png', **extra_args)
+
+    log.info("Dataset exported to '%s' as '%s'" % \
+        (dst_dir, args.output_format))
```
zhiltsov-max, line 125:
```suggestion
    output_format = args.output_format or default_converter_name

    try:
        converter = env.converters[output_format]
    except KeyError:
        raise CliException("Converter for format '%s' is not found" %
            output_format)
    extra_args = converter.parse_cmdline(args.extra_args)

    dst_dir = args.dst_dir
    if dst_dir:
        if not args.overwrite and osp.isdir(dst_dir) and os.listdir(dst_dir):
            raise CliException("Directory '%s' already exists "
                "(pass --overwrite to overwrite)" % dst_dir)
    else:
        dst_dir = generate_next_file_name('%s-%s' % (
            make_file_name(args.dataset_id),
            make_file_name(output_format),
        ))
    dst_dir = osp.abspath(dst_dir)

    log.info("Downloading the dataset")
    extractor = extractor_factory()

    log.info("Exporting the dataset")
    converter.convert(extractor, dst_dir,
        default_image_ext='.png', **extra_args)

    log.info("Dataset exported to '%s' as '%s'" % (dst_dir, output_format))
```

`datum download -i tfds:cifar10 -- --save-images` fails because of `make_file_name(args.output_format)`.

zhiltsov-max (approved):

PR #585, opened 2021-12-17
Add MARS (Import only)

kirill-sizov:
### Checklist

### License

```python
#
```


tests/test_mars_format.py
```diff
@@ -0,0 +1,60 @@
...
+
+class MarsImporterTest(TestCase):
+    @mark_requirement(Requirements.DATUM_GENERAL_REQ)
+    def test_can_import(self):
+        expected_dataset = Dataset.from_iterable([
+            DatasetItem(id='0001C1T0001F001', image=np.ones((10, 10, 3)),
+                subset='train', annotations=[Label(label=2, attributes={
+                    'is_distractors': False,
+                    'is_junk': False,
+                    'pedestrian_id': '0001',
```
zhiltsov-max, line 29: These 3 attributes look mutually exclusive. Can we reduce them to only one?

tests/test_mars_format.py
```diff
@@ -0,0 +1,60 @@
...
+                    'camera_id': '1',
+                    'track_id': '0001',
+                    'frame_id': '001'})]
```
zhiltsov-max, line 32: All the label attributes here look like item attributes. Why do you think we should store them in annotations?

tests/test_mars_format.py
```diff
@@ -0,0 +1,60 @@
...
+            DatasetItem(id='00-1C2T0081F201', image=np.ones((10, 10, 3)),
+                subset='test', annotations=[Label(label=0, attributes={
+                    'is_distractors': False,
+                    'is_junk': True,
+                    'pedestrian_id': '00-1',
+                    'camera_id': '2',
+                    'track_id': '0081',
+                    'frame_id': '201'})]
+            ),
+        ], categories=['00-1', '0000', '0001'])
```
zhiltsov-max, line 52: The LFW, WiderFace and MARS are related to the same task, consider aligning their representations.

datumaro/plugins/mars_format.py
```diff
@@ -0,0 +1,121 @@
...
+        for label_cat in self._categories[AnnotationType.label]:
+            label = label_cat.name
+            label_id = self._categories[AnnotationType.label].find(label)[0]
+            for image_path in find_images(osp.join(path, label)):
+                image_name = osp.basename(image_path)
+                pedestrian_id = image_name[0:4]
+
+                if not fnmatch.fnmatch(image_name,
+                        label + MarsPath.IMAGE_NAME_POSTFIX):
+                    log.warning(f'The image {image_path} will be skip because '
```
zhiltsov-max, line 70:
```suggestion
                    log.warning(f'The image {image_path} will be skipped because '
```

tests/test_mars_format.py
```diff
@@ -0,0 +1,60 @@
...
+
+ASSETS_DIR = osp.join(osp.dirname(__file__), 'assets')
+DUMMY_MARS_DATASET = osp.join(ASSETS_DIR, 'mars_dataset')
+
+class MarsImporterTest(TestCase):
+    @mark_requirement(Requirements.DATUM_GENERAL_REQ)
+    def test_can_import(self):
+        expected_dataset = Dataset.from_iterable([
+            DatasetItem(id='0001C1T0001F001', image=np.ones((10, 10, 3)),
+                subset='train', annotations=[Label(label=2, attributes={
```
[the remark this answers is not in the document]
kirill-sizov: Because each image in the dataset represents itself as clipped bounding box, and we don't have original image or information about coordinates of this bounding box in the original image.

site/content/en/docs/formats/mars.md
```diff
@@ -0,0 +1,76 @@
...
+datum create
+datum add -f mars ./dataset
+```
+
+> Note: the directory with dataset should be subdirectory of the
+> project directory.
+
+```
+mars_dataset
+├── <bbox_subset_name1>
```
zhiltsov-max, line 44: Will this format support metafile?

tests/test_mars_format.py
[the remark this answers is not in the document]
zhiltsov-max: Probably, I don't quite understand the problem. We have an image, which is a bbox cut from the original image. But original images are not used in this format, so this bbox should just fit the whole image, shouldn't it?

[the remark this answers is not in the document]
kirill-sizov: If I understand you correctly,  in such implementation we need to know size of image to assign the correct values to the coordinates of the bbox. Would there be an overhead of reading the image to find out the size of the image?

site/content/en/docs/formats/mars.md
```diff
@@ -0,0 +1,76 @@
...
+description: ''
+weight: 14
+---
+
+## Format specification
+
+MARS is a dataset for and motion analysis and person identification task,
+and this dataset it's extension of Market-1501 dataset format.
+MARS dataset is available for downloading
+[here](http://zheng-lab.cecs.anu.edu.au/Project/project_mars.html)
```
IRDonch, line 13:
```suggestion
[here](https://zheng-lab.cecs.anu.edu.au/Project/project_mars.html)
```

site/content/en/docs/formats/mars.md
```diff
@@ -0,0 +1,76 @@
+---
+title: 'MARS'
+linkTitle: 'MARS'
+description: ''
+weight: 14
+---
+
+## Format specification
+
+MARS is a dataset for and motion analysis and person identification task,
```
IRDonch, line 10:
```suggestion
MARS is a dataset for the motion analysis and person identification task,
```

site/content/en/docs/formats/mars.md
```diff
@@ -0,0 +1,76 @@
...
+and this dataset it's extension of Market-1501 dataset format.
```
IRDonch, line 11: There seems to be no verb here.

site/content/en/docs/formats/mars.md
```diff
@@ -0,0 +1,76 @@
...
+MARS dataset is available for downloading
+[here](http://zheng-lab.cecs.anu.edu.au/Project/project_mars.html)
+
+Supported types of annotations:
+- `Label`
+
+Required attributes:
+- `is_distractors` (bool): True when image with distractors,
```
IRDonch, line 19: "When" implies a dependent clause, which needs a verb - but there is no verb here.

site/content/en/docs/formats/mars.md
```diff
@@ -0,0 +1,76 @@
...
+    which negatively affect retrieval accuracy
+- `is_junk`: True for junk image which do not affect retrieval accuracy;
```
IRDonch, line 21:
```suggestion
- `is_junk`: True for junk images which do not affect retrieval accuracy;
```

Also, what's the data type?

site/content/en/docs/formats/mars.md
```diff
@@ -0,0 +1,76 @@
...
+- `pedestrian_id`: four-digit number in format `%04d`;
```
IRDonch, line 22: I'm confused by this. First of all, isn't it redundant to say `%04d` when you already said it's a four-digit number? Second, what data type is this attribute supposed to have? If `int`, then the format doesn't make sense. If `str`, then why not `int`?

In addition, the meaning of the attribute should be explained beyond just "number".

site/content/en/docs/formats/mars.md
```diff
@@ -0,0 +1,76 @@
...
+│   ├── 00-1 # junk images which do not affect retrieval accuracy
+│   │   ├── 00-1C1T0001F001.jpg
+│   │   ├── 00-1C1T0001F001.jpg
+│   │   ├── ...
+├── <bbox_subset_name2>
+│   ├── ...
+├── ...
+```
+
+All images in MARS dataset has strict convention of naming:
```
IRDonch, line 66:
```suggestion
All images in MARS dataset follow a strict convention of naming:
```

site/content/en/docs/formats/mars.md
```diff
@@ -0,0 +1,76 @@
...
+```
+xxxxCxTxxxxFxxx.jpg
+```
+- the first four digits indicate the pedestrian's number;
+- digits after `C` indicate the camera id;
```
IRDonch, line 71:
```suggestion
- digit after `C` indicates the camera id;
```

(Since there's only one digit.)

site/content/en/docs/formats/mars.md
```diff
@@ -0,0 +1,76 @@
...
+- four digits after `T` indicate the track id for this pedestrian;
+- three digits after `F` indicate the frame id with this track
```
IRDonch, line 73:
```suggestion
- three digits after `F` indicate the frame id with this track.
```

site/content/en/docs/formats/mars.md
```diff
@@ -0,0 +1,76 @@
...
+
+> Note: there are two specific pedestrian IDs 0000 and 00-1
+> which indicate distracting images and unwanted images respectively
```
IRDonch, line 76:
```suggestion
> which indicate distracting images and unwanted images respectively.
```

tests/test_mars_format.py
IRDonch, line 29: Does the value of this attribute represent the same information as the label value? If so, why does it exist?

tests/test_mars_format.py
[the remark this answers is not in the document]
zhiltsov-max: It's a fair point. Yes, it can affect loading time, however, if we look at masks in VOC-like formats, it works pretty fine even with more complex processing. I think, we can try to implement this.

site/content/en/docs/formats/mars.md
kirill-sizov: I'm not sure that this format needs a metafile.

tests/test_mars_format.py
kirill-sizov: What do you mean by "aligning"? Do you mean that the formats should have a similar representation in Datumaro or something else?

zhiltsov-max: Yes, exactly. It will look pretty hard to use if these formats are not convertible between each other.

site/content/en/docs/formats/mars.md
zhiltsov-max: Probably, I can agree.

kirill-sizov:
I fixed all comments and updated `market1501` format for compatibility with `mars` format, there is changes that I made for `market1501`:
- **Importer**: made heuristic more unique
- **Extractor**: made `query` items as separate subset, because the original dataset don't suppose that query image relate with images from other subsets
- **Load items**: added attributes according original dataset [description](http://zheng-lab.cecs.anu.edu.au/Project/project_reid.html) (I used `track_id` instead of `sequence_id` because these two attributes have the same meaning)
- **Converter**: implemented this strategy: if item has `person_id` attribute `Converter` try to save this item with original image naming convention, using default values for unknown attributes and incrementing `frame_id` for collisions.

I decided to keep both `person_id` attribute and `Label` for MARS format for compatibility with classification-task formats.
I also used the `str` type for `person_id`, for cases where users would like to mark some items with a special identifier.

tests/test_market1501_format.py
```diff
@@ -120,20 +101,16 @@ def test_can_save_dataset_with_no_save_images(self):
...
-                }),
-            DatasetItem(id='a/b/c/2', image=Image(
-                    path='a/b/c/2.bmp', data=np.zeros((3, 4, 3))),
-                attributes={
-                    'camera_id': 1,
-                    'person_id': 2,
-                    'query': True
-                }),
+            DatasetItem(id='0001_c1s1_0000_00', image=Image(
+                    path='0001_c1s1_0000_00.JPEG', data=np.zeros((4, 3, 3))),
```
zhiltsov-max, line 105: I see the new test doesn't check for relative path support. Are they supported?

datumaro/plugins/market1501_format.py
```diff
@@ -3,137 +3,148 @@
...
-                dirname = Market1501Path.BBOX_DIR + subset_name
-                if 'query' in item.attributes:
-                    query = item.attributes.get('query')
-                    if isinstance(query, str):
-                        query = strtobool(query)
-                    if query:
-                        dirname = Market1501Path.QUERY_DIR
+                pid = item.attributes.get('person_id')
+                match = Market1501Path.PATTERN.fullmatch(item.id)
+                if not match and pid is not None:
```
zhiltsov-max, line 131:
```suggestion
                if not match and pid:
```
I'm not sure we can use empty `pid`

zhiltsov-max (changes requested):
Looks good, please update Market docs.

zhiltsov-max (approved):

PR #592, opened 2021-12-21
Fix line matching in multimerge

### LINKED ISSUE #219: Return format is not uniform
https://github.com/openvinotoolkit/datumaro/blob/b3eaf4a5035ad37c4cf591643b5d0f4b18f64992/datumaro/util/annotation_util.py#L182

is different from 

https://github.com/openvinotoolkit/datumaro/blob/b3eaf4a5035ad37c4cf591643b5d0f4b18f64992/datumaro/util/annotation_util.py#L213

zhiltsov-max:
### Summary

Closes #219 

- Renamed `smooth_line` to `approximate_line`, added deprecation notice
- Added more descriptions and type annotations
- The function now always returns only a single array
- Updated implementation of the line comparison algorithm
- Fixed the problem when multimerge failed because of missing label categories
- Fixed area computation in case of self-intersections

### Checklist

### License

```python
#
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Using `Image`, `ByteImage` from `datumaro.util.image` - these classes
   are moved to `datumaro.components.media`
   (<https://github.com/openvinotoolkit/datumaro/pull/538>)
+- Using `smooth_line` from `datumaro.util.annotation_util` - the function
+  is renamed to `approximate_line` and has updated interface
+  (<https://github.com/openvinotoolkit/datumaro/pull/592>)
 
 ### Removed
 - Equality comparison support between `datumaro.components.media.Image`
@@ -58,6 +61,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   (<https://github.com/openvinotoolkit/datumaro/pull/564>)
 - Empty lines in VOC subset lists are not ignored
   (<https://github.com/openvinotoolkit/datumaro/pull/587>)
+- Fails in multimerge when lines are not approximated and when there are no
+  label categories (<https://github.com/openvinotoolkit/datumaro/pull/592>)
 
 ### Security
 - TBD
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,14 +42,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   (<https://github.com/openvinotoolkit/datumaro/pull/563>)
 - Allowed export options in the `datum merge` command
   (<https://github.com/openvinotoolkit/datumaro/pull/545>)
+- `smooth_line` from `datumaro.util.annotation_util` - the function
+  is renamed to `approximate_line` and has updated interface
+  (<https://github.com/openvinotoolkit/datumaro/pull/592>)
 
 ### Deprecated
 - Using `Image`, `ByteImage` from `datumaro.util.image` - these classes
   are moved to `datumaro.components.media`
   (<https://github.com/openvinotoolkit/datumaro/pull/538>)
-- Using `smooth_line` from `datumaro.util.annotation_util` - the function
-  is renamed to `approximate_line` and has updated interface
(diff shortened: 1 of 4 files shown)


datumaro/components/operations.py
```diff
@@ -768,21 +767,33 @@ def distance(self, a, b):
...
-        # compute inter-line area, normalize by common bbox
-        point_count = max(max(len(a.points) // 2, len(b.points) // 2), 5)
-        a, sa = smooth_line(a.points, point_count)
-        b, sb = smooth_line(b.points, point_count)
+        def _approx(line, segments):
+            if len(line) // 2 == segments + 1:
+                return np.reshape(line, (-1, 2))
+            else:
+                return approximate_line(line, segments=segments)
+        segments = max(max(len(a.points) // 2, len(b.points) // 2), 5) - 1
```
IRDonch, line 787: The two `max` calls look redundant - `max` can take more than 2 arguments.

datumaro/util/annotation_util.py
```diff
@@ -178,11 +188,18 @@ def OKS(a, b, sigma=None, bbox=None, scale=None):
...
+def approximate_line(points: np.ndarray, segments: int) -> np.ndarray:
+    """
+    Approximates a 2d line to the required number of segments. Points are
+    distributed uniformly across the resulting line.
 
-    if len(points) // 2 == segments:
-        return points
+    Returns:
+      new_points (ndarray): a flat array of new line point coordinates.
+        The shape is [segments, 2], the layout is [[x0, y0], [x1, y1], ...].
```
IRDonch, line 198: This description seems contradictory - if the array has two dimensions, it's not flat.

Also, the input array should be described as well.

datumaro/util/annotation_util.py
```diff
@@ -14,6 +14,16 @@
...
 
 
+def __getattr__(name: str):
+    if name == 'smooth_line':
+        import warnings
+        warnings.warn(f"Using {name} from '{__package__}' is deprecated, "
+            "the function is renamed to 'approximate_line'", DeprecationWarning,
+            stacklevel=2)
+
+        return approximate_line
```
IRDonch, line 24: From what I see in the diff, `approximate_line` doesn't return the same thing as `smooth_line` did (`new_points, step` vs `new_points`), so can it really be considered a compatible replacement?

IRDonch (changes requested):

datumaro/util/annotation_util.py
zhiltsov-max: Maybe we should just remove the previous version or leave the previous one and mark it deprecated, while introducing a new one? I feel, it was not widely used, but I can't be sure, of course. I noticed that the step value was totally incorrect. It was supposed to be in the resulting line metric, but it was in the original curve metric and had no relation to the resulting one. And then I understood we should not return a single step at all, because it is not equal in the resulting curve.

IRDonch: I [searched on GitHub](https://github.com/search?p=2&q=annotation_util+smooth_line&type=Code) and didn't find any uses that weren't in copies of Datumaro itself or CVAT. So I don't think it would be a big problem to remove it.

datumaro/components/operations.py

datumaro/util/annotation_util.py
zhiltsov-max: Fixed

datumaro/util/annotation_util.py
zhiltsov-max: Removed

CHANGELOG.md
```diff
@@ -47,6 +47,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Using `Image`, `ByteImage` from `datumaro.util.image` - these classes
   are moved to `datumaro.components.media`
   (<https://github.com/openvinotoolkit/datumaro/pull/538>)
+- Using `smooth_line` from `datumaro.util.annotation_util` - the function
+  is renamed to `approximate_line` and has updated interface
+  (<https://github.com/openvinotoolkit/datumaro/pull/592>)
```
IRDonch, line 52: This can't be listed under "Deprecated" anymore, because the function was removed.

zhiltsov-max: Fixed

IRDonch (approved):

PR #623, opened 2022-01-12
Detection for nested folders

kirill-sizov:
### Summary
Datasets are often contains in archives and users usually extract them in a separate directory. This leads to the fact that the dataset can be stored in subfolders, so this PR resolved problem when Datumaro cannot detect dataset format with such folder structure.

### Checklist

### License

```python
#
```


datumaro/components/dataset.py
```diff
@@ -905,16 +906,25 @@ def import_from(cls, path: str, format: Optional[str] = None,
...
             env = Environment()
 
-        matches = env.detect_dataset(path)
+        for _ in range(depth):
+            matches = env.detect_dataset(path)
+            if matches and len(matches) == 1:
+                return matches[0]
+
+            paths = glob.glob(osp.join(path, '*'))
+            if len(paths) != 1 or not osp.isdir(paths[0]):
```
zhiltsov-max, line 920: I suspect, we should also ignore hidden OS files like [`.DS_Store` on Mac](https://perishablepress.com/remove-macosx-ds-store-zip-files-mac/) - I noticed, they often get into the archives.

datumaro/components/dataset.py
```diff
@@ -905,16 +906,26 @@ def import_from(cls, path: str, format: Optional[str] = None,
         return dataset
 
     @staticmethod
-    def detect(path: str, env: Optional[Environment] = None) -> str:
+    def detect(path: str, env: Optional[Environment] = None,
```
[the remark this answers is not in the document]
kirill-sizov: What kind of error should occur if the directory does not exist? I thought that [these](https://github.com/openvinotoolkit/datumaro/blob/210dddd745cb9b66d48b961e4e359d150b95e4c6/datumaro/components/dataset.py#L925-L928) errors cover this situation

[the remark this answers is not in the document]
zhiltsov-max: FileNotFoundError

tests/test_dataset.py
```diff
@@ -157,6 +157,32 @@ def test_can_detect(self):
 
             self.assertEqual(DEFAULT_FORMAT, detected_format)
 
+    @mark_requirement(Requirements.DATUM_GENERAL_REQ)
+    def test_can_detect_with_nested_folder(self):
+        env = Environment()
+        env.importers.items = {DEFAULT_FORMAT: env.importers[DEFAULT_FORMAT]}
+        env.extractors.items = {DEFAULT_FORMAT: env.extractors[DEFAULT_FORMAT]}
```
zhiltsov-max, line 164: The test is generally fine, but consider extending it with a generic "fits everything" format (like `image dir`) to test for collisions and that we continue search till we find one match. Maybe, it should be in another test.

datumaro/components/dataset.py
```diff
@@ -911,6 +911,9 @@ def detect(path: str, env: Optional[Environment] = None,
         if env is None:
             env = Environment()
 
+        if not osp.exists(path):
+            raise FileNotFoundError(path)
```
zhiltsov-max, line 915: Consider moving it to `Environment.detect_dataset()`

tests/test_dataset.py
kirill-sizov: It's look that such situation when we have more than one match and continue search till we find one match, will occurs never. Because we continue search if our path has only one file system object and this object it's a directory, but it for this path we have more than one match it's mean that this path most likely contains more than one file system object, so that these two situations are contradictory.

zhiltsov-max: For example, we could detect "image_dir" and "imagenet" in any root image directory, but only if we go deeper, we we'll detect a COCO dataset, for example. I know it is implemented already, just asking for a test that ensures this logic works.

Example:
```
a/annotations/x_instances.json
a/images/y.jpg
^
here we detect few general matching formats (currently, imagedir, imagenet, celeba, align_celeba, cifar etc.)
```
And inside `a` we detect COCO.

kirill-sizov: Added this test.

datumaro/components/dataset.py
kirill-sizov: I moved it, but it looks like it will affect the #576 a bit.

zhiltsov-max (approved):

PR #628, opened 2022-01-19
Import for BraTS dataset format

### LINKED ISSUE #616: Import for BraTS dataset format
Support import for the original [Brain Tumor Segmentation dataset](https://www.med.upenn.edu/cbica/brats2020/data.html) and dataset with NUMPY annotation files.

Annotation types of BraTS:

- Masks

The data is presented in `.nii` format (for reading - `nibabel`)

yasakova-anastasia:
### Summary

Depends on #539 
- Resolves #616
- Added `nibabel` to requirements
- Reading pickle files is moved to a separate file (also for CIFAR)
- Added a new media format for multiframe images

### How to test

Unit tests and manually on original dataset

### Checklist

### License

```python
#
```


tests/test_brats_format.py
```diff
@@ -0,0 +1,41 @@
...
+    @mark_requirement(Requirements.DATUM_616)
+    def test_can_detect(self):
+        detected_formats = Environment().detect_dataset(DUMMY_DATASET_DIR)
+        self.assertEqual([BratsImporter.NAME], detected_formats)
+
+    @mark_requirement(Requirements.DATUM_616)
+    def test_can_import(self):
+        expected_dataset = Dataset.from_iterable([
+            DatasetItem(id='BRATS_001',
+                image=np.ones((1, 5, 2, 3)),
```
zhiltsov-max, line 27: What do these dimensions mean? I don't think it's a regular image.

site/content/en/docs/formats/brats_numpy.md
```diff
@@ -0,0 +1,82 @@
...
+├── labels
+├── val_brain_bbox.p
+└── val_ids.p
+```
+
+## Export to other formats
+
+Datumaro can convert a BraTS Numpy dataset into any other format [Datumaro supports](/docs/user-manual/supported_formats/).
+To get the expected result, convert the dataset to a format
+that supports segmentation masks.
```
zhiltsov-max, line 54:
```suggestion
that supports segmentation masks or cuboids.
```

datumaro/util/pickle_util.py
```diff
@@ -0,0 +1,28 @@
...
+    def find_class(self, module, name):
+        if module == "numpy.core.multiarray" and \
+                name in PickleLoader.safe_numpy:
+            return getattr(numpy.core.multiarray, name)
+        elif module == 'numpy' and name in PickleLoader.safe_numpy:
+            return getattr(numpy, name)
+        raise pickle.UnpicklingError("Global '%s.%s' is forbidden"
+            % (module, name))
+
+class PickleLoader():
```
zhiltsov-max, line 20:
```suggestion
class PickleLoader:
```

tests/test_brats_format.py
yasakova-anastasia: 3d images are stored in the dataset. These images can be represented as a set of 2d images.

Maybe we need to store 2d images to easily convert to other formats. The AC has a parameter for this (multi_frame - allows converting annotation of 3D images as sequence of 2D frames (optional, default False))

zhiltsov-max: It is the same as video or PCD then, which also have multiple frames inside. Support for new media types is added in #539. 

Current implementation cannot work with all the other system - transforms, filters, inference etc.

yasakova-anastasia: I replaced `Image` with `PointCloud`. Did you mean this?

zhiltsov-max: No, I meant, you probably need a new media type for this. Because neither 3d images, nor multiframe images do not fit in any existing format. Just build upon that PR and mark this dependent.

yasakova-anastasia: Added a new media format `MultiframeImage`.

datumaro/util/test_utils.py
```diff
@@ -148,10 +148,15 @@ def compare_datasets(test, expected: IDataset, actual: IDataset,
...
-                item_b.has_image and item_b.image.has_data:
-            test.assertEqual(item_a.image, item_b.image, item_a.id)
+        if require_images:
+            if (item_a.has_image and item_a.image.has_data) or \
+                    item_a.has_image and item_a.image.has_data and \
+                    item_b.has_image and item_b.image.has_data:
+                test.assertEqual(item_a.image, item_b.image, item_a.id)
+
+            elif item_a.media and item_b.media:
+                test.assertEqual(item_a.media.data, item_b.media.data, item_a.id)
```
zhiltsov-max, line 158: This check is not correct for array comparison, please use `np.array_equal`.

datumaro/components/media.py
```diff
@@ -465,3 +466,40 @@ def __eq__(self, other: object) -> bool:
...
+        for image in extra_images or []:
+            assert isinstance(image, (str, Image, np.ndarray)) or callable(image)
+            if isinstance(image, str):
+                image = Image(path=image)
+            elif isinstance(image, np.ndarray) or callable(image):
+                image = Image(data=image)
+            self._extra_images.append(image)
+
+class MultiframeImage(MediaElement):
+    def __init__(self, images: Optional[List[Union[
```
zhiltsov-max, line 487:
```suggestion
    def __init__(self, images: Optional[Iterable[Union[
```

datumaro/components/media.py
```diff
@@ -465,3 +466,40 @@ def __eq__(self, other: object) -> bool:
...
+                str, Image, np.ndarray, Callable[[str], np.ndarray]
+            ]]], path: Optional(str) = None):
```
zhiltsov-max, line 489:
```suggestion
            ]]], *,
            path: Optional[str] = None):
```

datumaro/components/media.py
```diff
@@ -465,3 +466,40 @@ def __eq__(self, other: object) -> bool:
...
+        self._path = path
```
zhiltsov-max, line 490:
```suggestion
        assert path or images
        self._path = path
```

datumaro/components/media.py
```diff
@@ -465,3 +466,40 @@ def __eq__(self, other: object) -> bool:
...
+        self._images: List[Image] = []
+        for image in images or []:
+            assert isinstance(image, (str, Image, np.ndarray)) or callable(image)
+
+            if isinstance(image, str):
+                image = Image(path=image)
+            elif isinstance(image, np.ndarray) or callable(image):
+                image = Image(data=image)
+
+            self._images.append(image)
```
zhiltsov-max, line 501:
```suggestion
            self._images.append(image)
        assert self._path or self._images
```

tests/test_brats_format.py
zhiltsov-max: Ok, please check if you can contribute to #539

datumaro/components/media.py
```diff
@@ -500,3 +500,30 @@ def __init__(self, path: str, extra_images: Optional[List[Image]] = None):
...
+                image = Image(path=image)
+            elif isinstance(image, np.ndarray) or callable(image):
+                image = Image(data=image)
+
+            self._images[i] = image
+
+        assert self._path or self._images
+
+    @property
+    def data(self) -> np.ndarray:
```
zhiltsov-max: The type annotation is incorrect.

zhiltsov-max, line 528: The type annotation is incorrect.

yasakova-anastasia: Fixed

tests/test_brats_numpy_format.py
```diff
@@ -0,0 +1,61 @@
...
+        expected_dataset = Dataset.from_iterable(
+            [
+                DatasetItem(
+                    id="HGG_val0",
+                    media=MultiframeImage([np.ones((1, 5, 3))] * 2),
+                    annotations=[
+                        Mask(np.array([[[0, 0], [0, 1], [1, 1], [1, 0], [1, 0]]]), label=0),
+                        Mask(np.array([[[1, 1], [1, 0], [0, 0], [0, 0], [0, 0]]]), label=1),
+                        Mask(np.array([[[0, 0], [0, 0], [0, 0], [0, 1], [0, 1]]]), label=2),
+                        Cuboid3d(position=[1, 1, 1], rotation=[2, 2, 2]),
```
zhiltsov-max: I see a big problem here that we can't specify the image id for these annotations. The same problem is with video frame annotations, and I have some ideas how to resolve it, but it goes far beyond the scope of this PR. We can add an `image_id` attribute as a temporary solution, which should be revisited later.

yasakova-anastasia: I think the `Сuboid` refers to the all multiframe image, not to a specific slice.

zhiltsov-max: Maybe, but it doesn't cancel my statement above. Currently I see that these masks are 3d (which is not expected), while the images are 2d. The Cuboid is also 2d, and it's not clear, which image it annotates.

yasakova-anastasia: I agree about the masks. We need to separate them and add an `image_id` attribute.

But there is only one box on the Multiframe image, and there is no information for which slice this annotation is.
I'm a little confused. I thought the Сuboid was 3d. That is, this box refers to several slices of the image.

zhiltsov-max:
1. Do you plan to support other tasks from the table in the [article](https://arxiv.org/pdf/1811.02629.pdf)?

image

2. Are pre-2017 and post-2017 versions compatible? If not, we need to clarify the format supported.

requirements-core.txt
```diff
@@ -3,6 +3,7 @@ defusedxml>=0.7.0
 lxml>=4.4.1
 matplotlib>=3.3.1
 networkx>=2.5
+nibabel==3.2.1
```
zhiltsov-max: Why the version is fixed?

tests/test_brats_numpy_format.py
zhiltsov-max: > But there is only one box on the Multiframe image, and there is no information for which slice this annotation is.

From my understanding of the article, it's possible that frames are just different image channels, and the masks are related to all of them. Mask channels encode one of 4 (or 3) dataset classes and have no relation to separate frames.

> I'm a little confused. I thought the Сuboid was 3d. That is, this box refers to several slices of the image.

Cuboids are 3d, but it also requires a 3d media object to be annotated. Several slices with no spatial info is not a 3d object.

I don't see any info about cuboids in the article, where do they come from? What they represent?

site/content/en/docs/formats/brats.md
```diff
@@ -0,0 +1,85 @@
...
+---
+
+## Format specification
+
+The original BraTS dataset is available
+[here](https://www.med.upenn.edu/sbia/brats2018/data.html).
+
+Supported annotation types:
+- `Mask`
+
```
zhiltsov-max, line 18: Please reflect what is the frame data layout in the parsed representation and how to interpret annotations read.

yasakova-anastasia: Added an information about it.

requirements-core.txt
yasakova-anastasia: This version was written in the requirements of AC. But perhaps we can put a non-fixed version.

tests/test_brats_numpy_format.py
yasakova-anastasia: > I don't see any info about cuboids in the article, where do they come from? What they represent?

I didn't find any information about these boxes either. This format is supported by AC.

yasakova-anastasia:
> Do you plan to support other tasks from the table in the [article](https://arxiv.org/pdf/1811.02629.pdf)?

It seems to me that segmentation will be enough, since I haven't seen the use of other tasks somewhere.

> Are pre-2017 and post-2017 versions compatible? If not, we need to clarify the format supported.

I supported BraTS'17-20. Added an information about it.

requirements-core.txt
```diff
@@ -3,7 +3,7 @@ defusedxml>=0.7.0
 lxml>=4.4.1
 matplotlib>=3.3.1
 networkx>=2.5
-nibabel==3.2.1
+nibabel
```
zhiltsov-max, line 6: >=3.2.1

requirements-core.txt
```diff
@@ -3,6 +3,7 @@ defusedxml>=0.7.0
 lxml>=4.4.1
 matplotlib>=3.3.1
 networkx>=2.5
+nibabel
```
zhiltsov-max:
```suggestion
nibabel>=3.2.1
```

tests/test_brats_numpy_format.py
zhiltsov-max: > From my understanding of the article, it's possible that frames are just different image channels, and the masks are related to all of them. Mask channels encode one of 4 (or 3) dataset classes and have no relation to separate frames.

Have you found confirmation for this? Do we need to identify the frame for annotations?

yasakova-anastasia: I understood it this way. We divide the multi-frame image into frames (these are slices of the brain). These 2d images are four-channel. This is from the documentation on the [official site](https://www.med.upenn.edu/sbia/brats2017/data.html):

> These multimodal scans describe a) native (T1) and b) post-contrast T1-weighted (T1Gd), c) T2-weighted (T2), and d) T2 Fluid Attenuated Inversion Recovery (FLAIR) volumes.

The mask refers to the entire multiframe image, but there is a one-to-one correspondence between image slices and mask slices.

I think it's more correct for Datumaro to store 2d masks with references to the frame (attribute).

zhiltsov-max: Ok, thanks for the investigation. I think, it's clear now.

zhiltsov-max (approved):

PR #630, opened 2022-01-21
Introduce API v1

zhiltsov-max:
### Summary

Introduces comfortable library API. For external use, the recommended way to call Datumaro API is:
``` python
import datumaro as dm
```

- Added HLops module with some CLI command duplicates
- Added more function docs
- Some `Dataset` method parameters became kw-only
- Updated examples in docs
- Removed compatibility symbols for annotation classes in `components.extractor`

Remaining:
- [ ] Check how docs look like

### Checklist

### License

```python
#
```


datumaro/components/dataset.py
```diff
@@ -713,10 +713,29 @@ def remove(self, id: str, subset: Optional[str] = None) -> None:
...
     def filter(self, expr: str, filter_annotations: bool = False,
             remove_empty: bool = False) -> Dataset:
+        """
+        Filters-out some dataset items or annotations, using a custom filter
+        expression.
+
+        Results are stored in-place. Modifications are applied lazily.
+
+        Args:
+            expr (str) - XPath-formatted filter expression
```
IRDonch, line 723: IMO, we shouldn't duplicate the type information in the documentation. It doesn't communicate anything new to the user and creates extra work for developers. And in the case of complex types, the type annotation syntax is more precise than English.

datumaro/components/dataset.py
```diff
@@ -713,10 +713,29 @@ def remove(self, id: str, subset: Optional[str] = None) -> None:
...
+                (e.g. "/item[subset = 'train']",
+                "/item/annotation[label = 'cat']")
```
IRDonch, line 725:
```suggestion
                (e.g. `/item[subset = 'train']`,
                `/item/annotation[label = 'cat']`)
```

datumaro/components/dataset.py
```diff
@@ -713,10 +713,29 @@ def remove(self, id: str, subset: Optional[str] = None) -> None:
...
+            filter_annotations (bool) - Indicates if the filter should be
+                applied to items or annotations
+            remove_empty (bool) - When filtering annotations, allows to
+                exclude empty items from the resulting dataset
+
+        Returns: a lazy-computible wrapper around the input dataset
```
IRDonch, line 731: This seems to be false, as the function delegates to `self.transform`, which always returns `self`.

datumaro/components/dataset.py
```diff
@@ -732,14 +751,28 @@ def update(self,
...
         Applies some function to dataset items.
+
+        Results are stored in-place. Modifications are applied lazily.
+
+        Args:
+            method (str or Transform) - The transformation function to be
+                applied to the dataset. If a string is passed, it is treated
+                as a plugin name, which is searched for in
+                the dataset environment
+            **kwargs - Optional parameters for the transformation
```
IRDonch, line 773:
```suggestion
            **kwargs - Parameters for the transformation
```

I assume that transformations can have both optional and mandatory parameters, which will both be specified using this argument.

datumaro/components/dataset.py
```diff
@@ -751,13 +784,27 @@ def transform(self, method: Union[str, Type[Transform]],
...
 
         return self
 
     def run_model(self, model, batch_size=1) -> Dataset:
+        """
+        Applies a model to dataset items' media and produces a dataset with
+        media and annotations.
+
+        Args:
+            model (Launcher or Modeltransform) - The model to be
```
IRDonch, line 799: These types ought to be in the function signature (also, `ModelTransform` is misspelled).

datumaro/components/dataset.py
```diff
@@ -751,13 +784,27 @@ def transform(self, method: Union[str, Type[Transform]],
...
+                applied to the dataset.
+            batch_size (int) - The number of dataset items, processed
```
IRDonch, line 801:
```suggestion
            batch_size (int) - The number of dataset items processed
```

datumaro/components/dataset.py
```diff
@@ -834,6 +881,16 @@ def flush_changes(self):
     @scoped
     def export(self, save_dir: str, format: Union[str, Type[Extractor]],
             **kwargs) -> None:
+        """
+        Saves the input dataset in some format.
```
IRDonch, line 885:
```suggestion
        Saves the dataset in some format.
```

IMO, "input dataset" doesn't really make sense here, since you don't pass a dataset explicitly into this function.

datumaro/components/dataset.py
```diff
@@ -834,6 +881,16 @@ def flush_changes(self):
...
+
+        Args:
```
IRDonch, line 887: Documentation for `save_dir` is missing.

datumaro/components/dataset.py
```diff
@@ -834,6 +881,16 @@ def flush_changes(self):
...
+            format (str or Converter) - The desired output format.
```
IRDonch, line 888: The type here contradicts the function signature:

* the signature has `Extractor` instead of `Converter` (the signature is wrong here).
* the signature asks for a _subtype_ of `Converter` instead of an instance (the docstring is wrong here).

datumaro/components/dataset.py
```diff
@@ -906,8 +963,18 @@ def import_from(cls, path: str, format: Optional[str] = None,
...
+    def detect(path: str, *,
+            env: Optional[Environment] = None, depth: int = 1) -> str:
+        """
+        Attempts to detect dataset format of a given directory.
+
+        Args:
+            path (str) - The directory to check
+            depth (int) - The maximum depth for recursive search
+            env (Environment) - A plugin collection. If not set, the default one
+                is used
```
IRDonch, line 975: The last sentence is unsatisfactory, IMO. What is "the default one"? You could describe it as "only built-in plugins are used", or, more indirectly, as "a new `Environment` instance is created".

datumaro/components/operations.py
```diff
@@ -86,8 +87,18 @@ def __call__(self, sources):
         raise NotImplementedError()
 
 class ExactMerge:
+    """
+    Merges few datasets using the "simple" algorithm:
```
IRDonch, line 91:
```suggestion
    Merges several datasets using the "simple" algorithm:
```

datumaro/components/operations.py
```diff
@@ -86,8 +87,18 @@ def __call__(self, sources):
...
+
+    - items are matched by (id, subset) pairs
+    - matching items share the data available (nothing + nothing = nothing,
```
IRDonch, line 94: What is meant by "data" here? There's no such attribute in `DatasetItem`.

datumaro/__init__.py
```diff
@@ -1,3 +1,21 @@
-# Copyright (C) 2019-2020 Intel Corporation
+# Copyright (C) 2019-2022 Intel Corporation
 #
 # SPDX-License-Identifier: MIT
+
+import datumaro.components.errors as errors
+import datumaro.components.operations as ops
+import datumaro.components.project as project
+
+from .components.annotation import *
```
IRDonch, line 9: IMO, we shouldn't use star imports. We might want to have names in the internal modules that are exported to other internal modules, but not to external users.

IRDonch (changes requested):

datumaro/components/hl_ops.py
```diff
@@ -0,0 +1,192 @@
...
+        dataset (IDataset) - The dataset to be transformed
+        method (str or Transform) - The transformation function to be
+            applied to the dataset. If a string is passed, it is treated
+            as a plugin name, which is searched for in the environment
+            set by the 'env' argument
+        env (Environment) - A plugin collection. If not set, the default one
+            is used
+        **kwargs - Optional parameters for the transformation
+
+    Returns: a lazy-computible wrapper around the input dataset
```
IRDonch, line 43: "Computible" is not a word, but even if you replace it with "computable" I don't think this sentence would work well. The wrapper itself is computed eagerly, it's the items that aren't. IMO, it should be rephrased to something like "a wrapper around the input dataset whose items are determined lazily".

datumaro/components/hl_ops.py
```diff
@@ -0,0 +1,192 @@
...
+from datumaro.components.environment import Environment
+from datumaro.components.extractor import Transform
+from datumaro.components.launcher import Launcher, ModelTransform
+from datumaro.components.operations import ExactMerge
+from datumaro.components.validator import TaskType, Validator
+from datumaro.util import parse_str_enum_value
+from datumaro.util.scope import on_error_do, scoped
+
+
+def transform(dataset: IDataset, method: Union[str, Transform], *,
```
IRDonch, line 26:
```suggestion
def transform(dataset: IDataset, method: Union[str, Type[Transform]], *,
```

datumaro/components/hl_ops.py
```diff
@@ -0,0 +1,192 @@
...
+        if env is None:
+            env = Environment()
+        method = env.transforms[method]
+
+    produced = method(dataset, **kwargs)
+
+    return Dataset(source=produced, env=env)
+
+def filter(dataset: IDataset, expr: str, *, #pylint: disable=redefined-builtin
+        filter_annotations: bool = False,
```
IRDonch, line 56: IMO, the fact that different values of this function select completely different implementations suggests that this function should be split into two, `filter_items` and `filter_annotations`.

datumaro/components/hl_ops.py
```diff
@@ -0,0 +1,192 @@
...
+        **kwargs - Optional parameters for the model
+
+    Returns: a lazy-computible wrapper around the input dataset
+    """
+
+    if isinstance(model, Launcher):
+        return transform(dataset, ModelTransform, launcher=model,
+            batch_size=batch_size, **kwargs)
+    elif isinstance(model, ModelTransform):
+        return transform(model, batch_size=batch_size, **kwargs)
```
IRDonch, line 120: This branch is suspicious - it doesn't involve the input dataset at all.

datumaro/components/hl_ops.py
```diff
@@ -0,0 +1,192 @@
...
+    else:
+        raise TypeError('Unexpected model argument type: %s' % type(model))
+
+# other operations
+
+@scoped
+def export(dataset: IDataset, path: str, format: Union[str, Converter], *,
```
IRDonch, line 127:
```suggestion
def export(dataset: IDataset, path: str, format: Union[str, Type[Converter]], *,
```

datumaro/components/hl_ops.py
```diff
@@ -0,0 +1,192 @@
...
+    Attempts to detect dataset format of a given directory.
+
+    Args:
+        path (str) - The directory to check
+        depth (int) - The maximum depth for recursive search
+        env (Environment) - A plugin collection. If not set, the default one
+            is used
+    """
+
+    return Dataset.detect(path, env=env, depth=depth)
```
IRDonch, line 168: Is this function even necessary if it's just a wrapper around `Dataset.detect`?

datumaro/components/hl_ops.py
```diff
@@ -0,0 +1,192 @@
...
+        task (str or TaskType) - Target task type - classification, detection
+            etc.
+        env (Environment) - A plugin collection. If not set, the default one
+            is used
+        **kwargs - Optional arguments for the validator
+
+    Returns: a dictionary with validation results
+    """
+
+    task = parse_str_enum_value(task, TaskType).name.lower()
```
IRDonch, line 186: Why is `lower` needed here? The members of `TaskType` are already lowercase.

IRDonch, line 186: Why is `lower` needed here? The members of `TaskType` are already lowercase.

tests/test_dataset.py
```diff
@@ -26,6 +26,7 @@
 from datumaro.components.launcher import Launcher
 from datumaro.components.media import Image
 from datumaro.util.test_utils import TestDir, compare_datasets
+import datumaro.components.hl_ops as hlops
```
IRDonch, line 29: Why change the spelling of the imported name? I think you should keep it as `hl_ops`.

datumaro/components/dataset.py
zhiltsov-max: Agree, don't know, why I came to this.

datumaro/components/dataset.py
zhiltsov-max: > the signature asks for a subtype of Converter instead of an instance (the docstring is wrong here).

It can be both, actually. Fixed

datumaro/components/dataset.py
zhiltsov-max: Fixed

datumaro/components/hl_ops.py
zhiltsov-max: It was done so some time ago, but it wasn't very comfortable to use. I'll reconsider such option.

datumaro/components/hl_ops.py
zhiltsov-max: Probably, you're right, and it brings no value yet. Maybe, we can add it later.

datumaro/components/hl_ops.py
zhiltsov-max: Fixed

datumaro/components/hl_ops.py
zhiltsov-max: Fixed

datumaro/components/hl_ops.py
zhiltsov-max: Fixed

datumaro/components/hl_ops.py
zhiltsov-max: Rephrased

datumaro/components/operations.py

datumaro/components/operations.py

datumaro/components/dataset.py
zhiltsov-max: Added

datumaro/components/dataset.py
zhiltsov-max: Fixed

datumaro/components/dataset.py
zhiltsov-max: Fixed

datumaro/components/dataset.py
zhiltsov-max: Fixed

datumaro/components/dataset.py
zhiltsov-max: Fixed such cases

datumaro/components/dataset.py
zhiltsov-max: Fixed description

datumaro/components/dataset.py
```diff
@@ -713,10 +713,29 @@ def remove(self, id: str, subset: Optional[str] = None) -> None:
 
     def filter(self, expr: str, filter_annotations: bool = False,
             remove_empty: bool = False) -> Dataset:
+        """
+        Filters-out some dataset items or annotations, using a custom filter
```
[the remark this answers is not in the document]
zhiltsov-max: Fixed

datumaro/components/dataset.py
zhiltsov-max: Applied

tests/test_dataset.py
zhiltsov-max: There is no specific reason. Ok, changed.

datumaro/components/hl_ops.py
zhiltsov-max: Actually, not to think about it. Probably, it's better to be checked in tests. Added a test for this.

datumaro/__init__.py
zhiltsov-max: There is a trade-off between having risks of not including something just added and showing internal symbols. In this particular case, I think, there are too many declarations for enumeration.

datumaro/components/dataset.py
IRDonch: I think you missed `validate`.

datumaro/components/hl_ops.py
```diff
@@ -0,0 +1,192 @@
...
+    if isinstance(method, str):
+        if env is None:
+            env = Environment()
+        method = env.transforms[method]
+
+    produced = method(dataset, **kwargs)
+
+    return Dataset(source=produced, env=env)
+
+def filter(dataset: IDataset, expr: str, *, #pylint: disable=redefined-builtin
```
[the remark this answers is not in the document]
IRDonch: Oh, I see. Never mind, then.

tests/test_validator.py
```diff
@@ -115,13 +116,18 @@ def setUpClass(cls):
...
+class TestBaseValidator(_TestValidatorBase):
     @classmethod
     def setUpClass(cls):
         cls.validator = _TaskValidator(task_type=TaskType.classification,
             few_samples_thr=1, imbalance_ratio_thr=50, far_from_mean_thr=5.0,
             dominance_ratio_thr=0.8, topk_bins=0.1)
 
+    @mark_requirement(Requirements.DATUM_GENERAL_REQ)
+    def test_has_enum_entries_in_environment(self):
+        for key in TaskType._member_names_:
```
IRDonch, line 128:
```suggestion
        for key in TaskType.__members__:
```

`_member_names_` is undocumented.

tests/test_validator.py
```diff
@@ -115,13 +116,18 @@ def setUpClass(cls):
...
+            self.assertTrue(key in Environment().validators.items, key)
```
IRDonch, line 129: I would suggest creating the environment once instead of on every iteration, and using `assertIn`.

datumaro/__init__.py
IRDonch: I think explicit imports are more future-proof, though, because if you forget to export something, you can just fix that (by adding it to the list), whereas if you export something you didn't want to, you can't remove it (you can't add exceptions to a star import).

Also, the problem of forgetting to export something can be mitigated by updating tests to import names from `datumaro` whenever possible.

There's also another benefit to explicit imports: they make it easier to see at a glance whether a PR adds or removes public API elements.

tests/test_validator.py
zhiltsov-max: Done

tests/test_validator.py
zhiltsov-max: Changed

IRDonch (approved):

PR #634, opened 2022-01-27
Improve simple merge

zhiltsov-max:
### Summary

This PR implements missing item attribute merging functionality in the simple merging algorithm. Also, it improves image merging in some cases, including data-supplied and placeholder (metainfo-only) images.

- Allowed `Image` creation from just a size value. Now images can be built from array, or a function returning it, a path, or size info, including any mixed variants. The old restriction for size seems unnecessary.
- Added `Image` class docs
- Added image search in VOC XML-based formats
- Added image path equality checks in simple merge, when applicable
- Added item attribute merging in simple merge

### Checklist

### License

```python
#
```


datumaro/components/media.py
```diff
@@ -44,14 +44,37 @@ def __init__(self,
...
+        The `ext` parameter cannot be used as a single argument for
+        construction.
+
+        Args:
+            data - Image pixels or a function to retrieve them. The expected
+                image shape is (H, W [, C]). If a function is provided,
+                it must accept image path as the first argument.
+            path - Image path
+            ext - Image extension. Cannot be used together with `path`. It is
+                useful for saving with a custom extension.
```
IRDonch, line 62: The last sentence isn't saying much. _How_ is it useful?

datumaro/components/media.py
```diff
@@ -72,8 +95,8 @@ def __init__(self,
         self._ext = ext
 
         if not isinstance(data, np.ndarray):
-            assert path or callable(data), "Image can not be empty"
-            assert data is None or callable(data)
+            assert path or callable(data) or size, "Image can not be empty"
+            assert data is None or callable(data) or size
```
IRDonch, line 99: What's the use case for `Image` objects with _only_ the size filled in? Isn't the fundamental property of an image that it can be resolved to a bitmap?

datumaro/components/operations.py
```diff
@@ -106,40 +108,95 @@ def merge(cls, *sources):
...
             annotations=cls.merge_anno(
                 existing_item.annotations, current_item.annotations))
 
     @staticmethod
-    def merge_images(existing_item, current_item):
-        image = None
-        if existing_item.has_image and current_item.has_image:
-            if existing_item.image.has_data:
-                image = existing_item.image
+    def merge_attrs(a: Dict, b: Dict,
```
IRDonch, line 118: Attribute keys are strings, so `Dict[str, Any]` would be a more specific type here.

datumaro/components/operations.py
```diff
@@ -106,40 +108,95 @@ def merge(cls, *sources):
...
+            item_id: Optional[Tuple[str, str]] = None) -> Dict:
+        merged = {}
+
+        for name in set(a) | set(b):
```
IRDonch, line 122:
```suggestion
        for name in a.keys() | b.keys():
```

This avoids creation of new sets.

datumaro/components/operations.py
```diff
@@ -106,40 +108,95 @@ def merge(cls, *sources):
...
+                image = item_b.image
+            elif item_a.image.has_size:
+                image = item_a.image
+            elif item_b.image.has_size:
+                image = item_b.image
             else:
-                image._size = current_item.image._size
-        elif existing_item.has_image:
-            image = existing_item.image
+                image = item_a.image
```
IRDonch, line 183: Shouldn't this branch be impossible? If we got here, it means that neither image has a path, data, _or_ size, which should not be possible.

tests/test_images.py
```diff
@@ -65,8 +65,8 @@ class ImageTest(TestCase):
     def test_lazy_image_shape(self):
         data = np.ones((5, 6, 3))
 
-        image_lazy = Image(data=data, size=(2, 4))
-        image_eager = Image(data=data)
+        image_lazy = Image(data=lambda _: data, size=(2, 4))
+        image_eager = Image(data=lambda _: data)
```
IRDonch, line 69: This seems wrong. It's not eager if it's initialized with a lambda.

tests/test_images.py
```diff
@@ -135,6 +139,7 @@ def test_ctors(self):
                 { 'data': image_bytes, 'path': path, 'size': (2, 4) },
                 { 'path': path },
                 { 'path': path, 'size': (2, 4) },
+                { 'path': path, 'size': (2, 4) },
```
IRDonch, line 142: Duplicate line?

IRDonch (changes requested):

datumaro/components/operations.py
zhiltsov-max: Yes, currently it should be unreachable, but it will be a fallback if `Image` class is changed.

datumaro/components/media.py
zhiltsov-max: > Isn't the fundamental property of an image that it can be resolved to a bitmap?

Actually, it is not true even without this PR, because `path` can point to an inexistent file, so it can only be used as metainfo. And if the file is missing, but there is image size info in the format, we often try to construct a generic path, which such item would have. I think, that it is not totally correct, and we could just provide size metainfo in such cases. But I'm not sure. Maybe, it's worth to leave unchanged, until we have more strong reasons.

datumaro/components/errors.py
```diff
@@ -223,14 +225,35 @@ def _my__init__(self, msg=None, *, sources=None):
...
+@attrs
+class MismatchingAttributesError(DatasetMergeError):
+    item_id: Optional[Tuple[str, str]] = attrib()
+    key: str = attrib()
+    a: Any = attrib()
+    b: Any = attrib()
+
+    def __str__(self):
+        return "Item %s: mismatching image attribute %s: %s vs %s" % \
+            (self.item_id or '', self.key, self.a, self.b)
```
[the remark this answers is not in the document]
zhiltsov-max: Maybe, but I don't want to be very specific at this point about exceptions. Probably, it can be designed more thoroughly in a separate PR. For this PR I'd prefer to make the `ExactMerge` methods private to avoid `item_id=None`.

datumaro/components/errors.py
```diff
@@ -223,14 +225,35 @@ def _my__init__(self, msg=None, *, sources=None):
 
 @attrs
 class MismatchingImageInfoError(DatasetMergeError):
-    item_id = attrib()
-    a = attrib()
-    b = attrib()
+    item_id: Optional[Tuple[str, str]] = attrib()
+    a: int = attrib()
```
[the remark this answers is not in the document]
zhiltsov-max: Fixed

datumaro/components/operations.py

datumaro/components/operations.py
zhiltsov-max: I didn't think dict views are documented to be set-like, but it seems, they are.

IRDonch: It is documented here: https://docs.python.org/3/library/stdtypes.html#dictionary-view-objects.

datumaro/components/operations.py
IRDonch: Shouldn't it throw an exception, then? If this branch is reached due to `Image` changing, then you can't know which image should be selected.

datumaro/components/errors.py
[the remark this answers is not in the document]
IRDonch: Works for me.

tests/test_images.py
zhiltsov-max: Removed

tests/test_images.py
zhiltsov-max: Refactored test

datumaro/components/media.py
zhiltsov-max: Moreover, guessing the path, we can use another image (check the VOC import).

datumaro/components/operations.py
zhiltsov-max: Yes, probably, it will be better.

datumaro/components/media.py
IRDonch: I was going to write a rebuttal, but I think I actually found a use case where having the image dimensions is better than not having anything: you can use image dimensions when saving annotations in a format that requires normalized box coordinates.

So I guess there is some use for it, although I'm not a fan of how it complicates the data model...

datumaro/components/operations.py
zhiltsov-max: Done

IRDonch: I think you forgot to push it.

IRDonch (approved):

PR #641, opened 2022-01-30
Refactor COCO parsing

zhiltsov-max:
### Summary

Experiments have shown that COCO import is quite slow:
```

dataset = Dataset.import_from('coco/', 'coco_instances') # train + val 2014
dataset.init_cache()
```
Works ~57s on my machine. This patch does some refactoring to make loading faster.

- Dont use pycocotools for parsing (no benefits at all except format logic encapsulation) (dropping gives ~6s / 10% speedup)
- Allow lazy rle and make masks lazy

Parsing can be further improved by using more performant json library (`orjson` gives ~3s speedup), it can be done in another PR separately.

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/plugins/coco_format/extractor.py
+++ b/datumaro/plugins/coco_format/extractor.py
@@ -205,7 +205,7 @@ class _CocoExtractor(SourceExtractor):
         mask = bgr2index(mask)
         return mask
 
-    @define(on_setattr=False)
+    @define
     class _lazy_merged_mask:
         segmentation: Any
         h: int
--- a/datumaro/plugins/coco_format/extractor.py
+++ b/datumaro/plugins/coco_format/extractor.py
@@ -1,13 +1,13 @@
-# Copyright (C) 2019-2021 Intel Corporation
+# Copyright (C) 2019-2022 Intel Corporation
 #
 # SPDX-License-Identifier: MIT
 
-from collections import OrderedDict
+from typing import Any
 import json
 import logging as log
 import os.path as osp
 
-from pycocotools.coco import COCO
+from attrs import define
 import pycocotools.mask as mask_utils
 
 from datumaro.components.annotation import (
@@ -18,15 +18,20 @@ from datumaro.components.extractor import (
     DEFAULT_SUBSET_NAME, DatasetItem, SourceExtractor,
 )
 from datumaro.components.media import Image
+from datumaro.util import take_by
 from datumaro.util.image import lazy_image, load_image
 from datumaro.util.mask_tools import bgr2index
 from datumaro.util.meta_file_util import has_meta_file, parse_meta_file
-from datumaro.util.os_util import suppress_output
 
 from .format import CocoPath, CocoTask
 
 
 class _CocoExtractor(SourceExtractor):
+    """
+    Parses COCO annotations written in the following format:
+    https://cocodataset.org/#format-data
+    """
+
     def __init__(self, path, task, *,
         merge_instance_polygons=False,
         subset=None,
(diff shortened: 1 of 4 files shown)


datumaro/plugins/coco_format/extractor.py
```diff
@@ -215,47 +205,54 @@ def _load_pan_mask(path):
         mask = bgr2index(mask)
         return mask
 
+    @define(on_setattr=False)
```
IRDonch, line 208: What does `on_setattr=False` do? According to the reference, it should be a callable.

datumaro/plugins/coco_format/extractor.py
```diff
@@ -54,145 +59,134 @@ def __init__(self, path, task, *,
...
-            self._load_categories(
-                loader,
-                keep_original_ids=keep_original_category_ids,
-            )
-            self._items = list(self._load_items(loader).values())
-
-    @staticmethod
-    def _make_subset_loader(path):
-        # TODO: remove pycocotools dependency?
+            self._items = self._load_items(json_data)
```
IRDonch, line 73: Is there a reason to no longer convert `self._items` to a list? I don't see its keys being used anywhere after `_load_items` completes.

datumaro/plugins/coco_format/extractor.py
IRDonch: And judging by the usage of the class, `attrs.frozen` might be a better fit anyway.

datumaro/plugins/coco_format/extractor.py
```diff
@@ -1,13 +1,13 @@
...
 # SPDX-License-Identifier: MIT
 
-from collections import OrderedDict
+from typing import Any
 import json
 import logging as log
 import os.path as osp
 
-from pycocotools.coco import COCO
+from attrs import define
```
IRDonch, line 10: Using the `attrs` package requires bumping the lower bound for the `attrs` dependency to 21.3.

datumaro/plugins/coco_format/extractor.py
zhiltsov-max: Strangely, I remember I've read it somewhere, but docs are silent about it. Maybe, we don't need this at all in this class.

It disables `setattr` checks for new-style declaration (via `define`):

``` python
def myconv(v):
    print('conv')
    return v

def myvalid(a, b, v):
    print('valid')

@define(on_setattr=False)
class A:
    x = field(default=5, converter=myconv, validator=myvalid)

    def __init__(self):
        self.x = 10

    def f(self):
        print(self.x)

@define(on_setattr=None)
class B:
    x = field(default=5, converter=myconv, validator=myvalid)

    def __init__(self):
        self.x = 10

    def f(self):
        print(self.x)

@attrs(init=False)
class C:
    x = field(default=5, converter=myconv, validator=myvalid)

    def __init__(self):
        self.x = 10

    def f(self):
        print(self.x)

print('A1')
print(A())

print('A2')
A().x = 7

print('B1')
print(B())

print('B2')
B().x = 7

print('C1')
print(C())

print('C2')
C().x = 7

>>>
A1
A(x=10)
A2

B1
conv
valid
B(x=10)
B2

conv
valid
conv
valid
C1
C(x=10)
C2
```

datumaro/plugins/coco_format/extractor.py
zhiltsov-max: I'm not sure I got the question, but there is no point in converting it to list.

IRDonch: The point is to free up the memory used for the keys and for the dictionary internals.

I would also say that changing `_items` to a `dict` breaks the contract established by `SourceExtractor`. It happens to work now, but it could blow up if `SourceExtractor` is modified.

datumaro/plugins/coco_format/extractor.py
```diff
@@ -54,145 +59,134 @@ def __init__(self, path, task, *,
 
         self._merge_instance_polygons = merge_instance_polygons
 
+        json_data = self._load_json(path)
+        self._label_map = {} # coco_id -> dm_id
```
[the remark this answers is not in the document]
zhiltsov-max: This is mostly for comment. I agree with your points, but as a developer, it's not something I would like to worry about. Also, it will produce wrong field access error messages, if there are annotations, but no categories.

datumaro/plugins/coco_format/extractor.py
```diff
@@ -54,145 +59,134 @@ def __init__(self, path, task, *,
...
+        self._load_categories(json_data,
+            keep_original_ids=keep_original_category_ids,
+        )
+
         if self._task == CocoTask.panoptic:
             #panoptic is not added to pycocotools
```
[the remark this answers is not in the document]
zhiltsov-max: Removed

datumaro/plugins/coco_format/extractor.py
zhiltsov-max: Removed

datumaro/plugins/coco_format/extractor.py
zhiltsov-max: Done

datumaro/plugins/coco_format/extractor.py
zhiltsov-max: > The point is to free up the memory used for the keys and for the dictionary internals.

This optimization is not effective.

> I would also say that changing _items to a dict breaks the contract established by SourceExtractor. It happens to work > now, but it could blow up if SourceExtractor is modified.

It's possible, but it's unlikely to happen occasionally and unnoticed.

datumaro/plugins/coco_format/extractor.py
IRDonch: FWIW, I think the documented way of achieving this effect is `on_setattr=attrs.setters.NO_OP`, although I don't think it's needed here.

IRDonch (approved):

PR #648, opened 2022-02-03
Change json lib

zhiltsov-max:
### Summary

For original COCO dataset reading (350 MB json) the speed-up is ~30%, for writing - 10x (which gives ~2.5x speed-up in export). There are some features that can be useful, like builtin `numpy` support.
https://github.com/ijl/orjson

- Replace standard `json` library with `orjson`
- Updated most places (except OpenImages format, where it's not a drop-in replacement)

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/cli/commands/detect_format.py
+++ b/datumaro/cli/commands/detect_format.py
@@ -4,8 +4,6 @@
 
 import argparse
 
-import orjson
-
 from datumaro.cli.util import MultilineFormatter
 from datumaro.cli.util.project import load_project
 from datumaro.components.environment import Environment
@@ -13,6 +11,7 @@ from datumaro.components.errors import ProjectNotFoundError
 from datumaro.components.format_detection import (
     RejectionReason, detect_dataset_format,
 )
+from datumaro.util import dump_json_file
 from datumaro.util.scope import scope_add, scoped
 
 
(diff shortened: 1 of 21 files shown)


datumaro/cli/commands/detect_format.py
```diff
@@ -3,7 +3,8 @@
 # SPDX-License-Identifier: MIT
 
 import argparse
-import json
+
+import orjson
```
nmanovic, line 7: Why not `import orjson as json`?

nmanovic: OK. I see  that it has slightly different interface.

datumaro/cli/commands/diff.py
```diff
@@ -202,8 +203,9 @@ def diff_command(args):
         output_file = osp.join(dst_dir,
             generate_next_file_name('diff', ext='.json', basedir=dst_dir))
         log.info("Saving diff to '%s'" % output_file)
-        with open(output_file, 'w') as f:
-            json.dump(output, f, indent=4, sort_keys=True)
+        with open(output_file, 'wb') as f:
+            f.write(orjson.dumps(output,
+                option=orjson.OPT_SORT_KEYS | orjson.OPT_INDENT_2))
```
IRDonch, line 208: IMO, we need to add utility functions for common operations (i.e. "save JSON to file" and "load JSON from file") and replace all these chunks of code with calls to those functions. That way:

1. We eliminate repetitive code.
2. We'll be able to easily swap out the JSON implementation later if we need to.

zhiltsov-max: Done

PR #650, opened 2022-02-03
Add progress and error reporting API

zhiltsov-max:
### Summary

Closes https://github.com/openvinotoolkit/datumaro/issues/142

- Added progress and error reporting in `Dataset.import_from` and `Dataset.export` (backward compatible)
- Added an ability to skip or fail on import and export problems with specific images or annotations
- Updated COCO, VOC and YOLO formats
- Improved error messages in formats
- Now, `Extractor` ctor has only kw args. Subclasses must define their own c-tor, which accepts 1 positional and any kwargs. Unused kwargs must be passed to the base class.
- <s> Added `Scope.add_many()` to enter multiple CMs in 1 call. </s>

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/components/converter.py
+++ b/datumaro/components/converter.py
@@ -10,6 +10,7 @@ import os.path as osp
 import shutil
 
 from attrs import define, field
+import attr
 
 from datumaro.components.cli_plugin import CliPlugin
 from datumaro.components.errors import (
@@ -63,8 +64,10 @@ class FailingExportErrorPolicy(ExportErrorPolicy):
 
 @define(eq=False)
 class ExportContext:
-    progress_reporter: ProgressReporter = field(factory=NullProgressReporter)
-    error_policy: ExportErrorPolicy = field(factory=FailingExportErrorPolicy)
+    progress_reporter: ProgressReporter = field(default=None,
+        converter=attr.converters.default_if_none(factory=NullProgressReporter))
+    error_policy: ExportErrorPolicy = field(default=None,
+        converter=attr.converters.default_if_none(factory=FailingExportErrorPolicy))
 
 class NullExportContext(ExportContext):
     pass
--- a/datumaro/components/converter.py
+++ b/datumaro/components/converter.py
@@ -32,7 +32,7 @@ class _ExportFail(DatumaroError):
 
 class ExportErrorPolicy:
     def report_item_error(self, error: Exception, *,
-            item_id: Tuple[str, str]):
+            item_id: Tuple[str, str]) -> None:
         """
         Allows to report a problem with a dataset item.
         If this function returns, the converter must skip the item.
@@ -46,7 +46,7 @@ class ExportErrorPolicy:
             raise error
 
     def report_annotation_error(self, error: Exception, *,
-            item_id: Tuple[str, str]):
+            item_id: Tuple[str, str]) -> None:
         """
         Allows to report a problem with a dataset item annotation.
         If this function returns, the converter must skip the annotation.
@@ -59,11 +59,11 @@ class ExportErrorPolicy:
         else:
             raise error
 
-    def _handle_item_error(self, error: ItemExportError):
+    def _handle_item_error(self, error: ItemExportError) -> None:
         """This function must either call fail() or return."""
         self.fail(error)
 
-    def _handle_annotation_error(self, error: AnnotationExportError):
+    def _handle_annotation_error(self, error: AnnotationExportError) -> None:
         """This function must either call fail() or return."""
         self.fail(error)
 
--- a/datumaro/components/converter.py
+++ b/datumaro/components/converter.py
@@ -1,21 +1,59 @@
-# Copyright (C) 2019-2021 Intel Corporation
+# Copyright (C) 2019-2022 Intel Corporation
 #
 # SPDX-License-Identifier: MIT
 
 from tempfile import mkdtemp
-from typing import Union
+from typing import Iterable, NoReturn, Optional, Tuple, TypeVar, Union
 import logging as log
 import os
 import os.path as osp
 import shutil
 
+from attrs import define
+
 from datumaro.components.cli_plugin import CliPlugin
+from datumaro.components.errors import (
+    AnnotationExportError, DatumaroError, ItemExportError,
+)
 from datumaro.components.extractor import DatasetItem
 from datumaro.components.media import Image
+from datumaro.components.progress_reporting import ProgressReporter
 from datumaro.util.meta_file_util import save_meta_file
 from datumaro.util.os_util import rmtree
 from datumaro.util.scope import on_error_do, scoped
 
+T = TypeVar('T')
+
+class _ExportFail(DatumaroError):
+    pass
+
+class ExportErrorPolicy:
+    def report_item_error(self, error: ItemExportError) -> Optional[NoReturn]:
+        """
+        Allows to report a problem with a dataset item.
+
+        This function must either call fail() or return. If this function
+        returns, the converter must skip the item.
+        """
+        raise NotImplementedError
+
+    def report_annotation_error(self,
+            error: AnnotationExportError) -> Optional[NoReturn]:
+        """
+        Allows to report a problem with a dataset item annotation.
+
+        This function must either call fail() or return. If this function
+        returns, the converter must skip the annotation.
+        """
+        raise NotImplementedError
+
+    def fail(self, error: Exception) -> NoReturn:
+        raise _ExportFail from error
+
+@define(eq=False)
+class ExportContext:
+    progress_reporter: Optional[ProgressReporter] = None
+    error_policy: Optional[ExportErrorPolicy] = None
 
 class Converter(CliPlugin):
     DEFAULT_IMAGE_EXT = None
@@ -72,7 +110,8 @@ class Converter(CliPlugin):
         raise NotImplementedError("Should be implemented in a subclass")
 
     def __init__(self, extractor, save_dir, save_images=False,
-            image_ext=None, default_image_ext=None, save_dataset_meta=False):
+            image_ext=None, default_image_ext=None, save_dataset_meta=False,
+            ctx: Optional[ExportContext] = None):
         default_image_ext = default_image_ext or self.DEFAULT_IMAGE_EXT
         assert default_image_ext
         self._default_image_ext = default_image_ext
@@ -92,6 +131,8 @@ class Converter(CliPlugin):
             self._patch = extractor.patch
         else:
             self._patch = None
+
+        self._ctx = ctx
 
     def _find_image_ext(self, item: Union[DatasetItem, Image]):
         src_ext = None
@@ -152,3 +193,29 @@ class Converter(CliPlugin):
 
     def _save_meta_file(self, path):
         save_meta_file(path, self._extractor.categories())
+
+    def _with_progress(self, iterable: Iterable[T], *,
+            total: Optional[int] = None,
+            desc: Optional[str] = None
+    ) -> Iterable[T]:
+        if self._ctx and self._ctx.progress_reporter:
+            yield from self._ctx.progress_reporter.iter(iterable,
+                total=total, desc=desc)
+        else:
+            yield from iterable
+
+    def _report_item_error(self, error: Exception, *,
+            item_id: Tuple[str, str]) -> Optional[NoReturn]:
+        if self._ctx and self._ctx.error_policy:
+            ie = ItemExportError(item_id)
+            ie.__cause__ = error
+            return self._ctx.error_policy.report_item_error(ie)
+        raise _ExportFail from error
+
+    def _report_annotation_error(self, error: Exception, *,
+            item_id: Tuple[str, str]) -> Optional[NoReturn]:
+        if self._ctx and self._ctx.error_policy:
+            ie = AnnotationExportError(item_id)
+            ie.__cause__ = error
+            return self._ctx.error_policy.report_annotation_error(ie)
+        raise _ExportFail from error
--- a/datumaro/components/converter.py
+++ b/datumaro/components/converter.py
@@ -3,13 +3,13 @@
 # SPDX-License-Identifier: MIT
 
 from tempfile import mkdtemp
-from typing import Iterable, NoReturn, Optional, Tuple, TypeVar, Union
+from typing import NoReturn, Optional, Tuple, TypeVar, Union
 import logging as log
 import os
 import os.path as osp
 import shutil
 
-from attrs import define
+from attrs import define, field
 
 from datumaro.components.cli_plugin import CliPlugin
 from datumaro.components.errors import (
@@ -17,7 +17,9 @@ from datumaro.components.errors import (
 )
 from datumaro.components.extractor import DatasetItem, IExtractor
 from datumaro.components.media import Image
-from datumaro.components.progress_reporting import ProgressReporter
+from datumaro.components.progress_reporting import (
+    NullProgressReporter, ProgressReporter,
+)
 from datumaro.util.meta_file_util import save_meta_file
 from datumaro.util.os_util import rmtree
 from datumaro.util.scope import on_error_do, scoped
@@ -49,10 +51,23 @@ class ExportErrorPolicy:
     def fail(self, error: Exception) -> NoReturn:
         raise _ExportFail from error
 
+class FailingExportErrorPolicy(ExportErrorPolicy):
+    def report_item_error(self, error: ItemExportError):
+        self.fail(error)
+
+    def report_annotation_error(self, error: AnnotationExportError):
+        self.fail(error)
+
+    def fail(self, error: Exception) -> NoReturn:
+        raise _ExportFail from error
+
 @define(eq=False)
 class ExportContext:
-    progress_reporter: Optional[ProgressReporter] = None
-    error_policy: Optional[ExportErrorPolicy] = None
+    progress_reporter: ProgressReporter = field(factory=NullProgressReporter)
+    error_policy: ExportErrorPolicy = field(factory=FailingExportErrorPolicy)
+
+class NullExportContext(ExportContext):
+    pass
 
 class Converter(CliPlugin):
     DEFAULT_IMAGE_EXT = None
@@ -134,7 +149,9 @@ class Converter(CliPlugin):
         else:
             self._patch = None
 
-        self._ctx = ctx
+        if ctx is None:
+            ctx = NullExportContext()
+        self._ctx: ExportContext = ctx
 
     def _find_image_ext(self, item: Union[DatasetItem, Image]):
         src_ext = None
@@ -195,16 +212,6 @@ class Converter(CliPlugin):
 
     def _save_meta_file(self, path):
         save_meta_file(path, self._extractor.categories())
-
-    def _with_progress(self, iterable: Iterable[T], *,
-            total: Optional[int] = None,
-            desc: Optional[str] = None
-    ) -> Iterable[T]:
-        if self._ctx and self._ctx.progress_reporter:
-            yield from self._ctx.progress_reporter.iter(iterable,
-                total=total, desc=desc)
-        else:
-            yield from iterable
 
     def _report_item_error(self, error: Exception, *,
             item_id: Tuple[str, str]):
--- a/datumaro/components/converter.py
+++ b/datumaro/components/converter.py
@@ -28,7 +28,7 @@ class _ExportFail(DatumaroError):
     pass
 
 class ExportErrorPolicy:
-    def report_item_error(self, error: ItemExportError) -> Optional[NoReturn]:
+    def report_item_error(self, error: ItemExportError):
         """
         Allows to report a problem with a dataset item.
 
@@ -37,8 +37,7 @@ class ExportErrorPolicy:
         """
         raise NotImplementedError
 
-    def report_annotation_error(self,
-            error: AnnotationExportError) -> Optional[NoReturn]:
+    def report_annotation_error(self, error: AnnotationExportError):
         """
         Allows to report a problem with a dataset item annotation.
 
@@ -208,7 +207,13 @@ class Converter(CliPlugin):
             yield from iterable
 
     def _report_item_error(self, error: Exception, *,
-            item_id: Tuple[str, str]) -> Optional[NoReturn]:
+            item_id: Tuple[str, str]):
+        """
+        Allows to report a problem with a dataset item.
+
+        This function must either fail or return. If this function
+        returns, the converter must skip the item.
+        """
         if self._ctx and self._ctx.error_policy and \
                 not isinstance(error, _ExportFail):
             ie = ItemExportError(item_id)
@@ -217,7 +222,13 @@ class Converter(CliPlugin):
         raise _ExportFail from error
 
     def _report_annotation_error(self, error: Exception, *,
-            item_id: Tuple[str, str]) -> Optional[NoReturn]:
+            item_id: Tuple[str, str]):
+        """
+        Allows to report a problem with a dataset item annotation.
+
+        This function must either fail or return. If this function
+        returns, the converter must skip the item.
+        """
         if self._ctx and self._ctx.error_policy and \
                 not isinstance(error, _ExportFail):
             ie = AnnotationExportError(item_id)
--- a/datumaro/components/converter.py
+++ b/datumaro/components/converter.py
@@ -15,7 +15,7 @@ from datumaro.components.cli_plugin import CliPlugin
 from datumaro.components.errors import (
     AnnotationExportError, DatumaroError, ItemExportError,
 )
-from datumaro.components.extractor import DatasetItem
+from datumaro.components.extractor import DatasetItem, IExtractor
 from datumaro.components.media import Image
 from datumaro.components.progress_reporting import ProgressReporter
 from datumaro.util.meta_file_util import save_meta_file
@@ -109,8 +109,11 @@ class Converter(CliPlugin):
     def apply(self):
         raise NotImplementedError("Should be implemented in a subclass")
 
-    def __init__(self, extractor, save_dir, save_images=False,
-            image_ext=None, default_image_ext=None, save_dataset_meta=False,
+    def __init__(self, extractor: IExtractor, save_dir: str, *,
+            save_images: bool = False,
+            image_ext: Optional[str] = None,
+            default_image_ext: Optional[str] = None,
+            save_dataset_meta: bool = False,
             ctx: Optional[ExportContext] = None):
         default_image_ext = default_image_ext or self.DEFAULT_IMAGE_EXT
         assert default_image_ext
--- a/datumaro/components/converter.py
+++ b/datumaro/components/converter.py
@@ -31,36 +31,51 @@ class _ExportFail(DatumaroError):
     pass
 
 class ExportErrorPolicy:
-    def report_item_error(self, error: ItemExportError):
+    def report_item_error(self, error: Exception, *,
+            item_id: Tuple[str, str]):
         """
         Allows to report a problem with a dataset item.
-
-        This function must either call fail() or return. If this function
-        returns, the converter must skip the item.
+        If this function returns, the converter must skip the item.
         """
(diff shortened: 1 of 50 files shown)


datumaro/components/extractor.py
```diff
@@ -170,22 +176,91 @@ def get(self, id, subset=None):
...
+        """
+        raise NotImplementedError
+
+    def fail(self, error: Exception) -> NoReturn:
+        raise _ImportFail from error
+
+@define(eq=False)
+class ImportContext:
+    progress_reporter: Optional[ProgressReporter] = None
+    error_policy: Optional[ImportErrorPolicy] = None
```
IRDonch, line 210: IMO, it's a bit weird that one of these is called "reporter" and the other is called "policy" when both are similar objects - abstract interfaces for reporting various things. Maybe renaming the latter to `error_reporter` would be better?

IRDonch (commented):
Very limited review so far - I will come back to this.

datumaro/components/extractor.py
```diff
@@ -170,22 +176,91 @@ def get(self, id, subset=None):
                 return item
         return None
 
+T = TypeVar('T')
+
+class _ImportFail(DatumaroError):
+    pass
+
+class ImportErrorPolicy:
+    def report_item_error(self, error: ItemImportError) -> Optional[NoReturn]:
```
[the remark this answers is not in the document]
zhiltsov-max: The idea is to highlight that it should be expected behavior and not something extraordinary.

datumaro/components/extractor.py
zhiltsov-max: I can't say that error policy is just for reporting, because it is supposed to define extractor reaction. Progress reported has a direct function of progress reporting.

datumaro/components/progress_reporting.py
```diff
@@ -0,0 +1,45 @@
...
+    def finish(self):
+        raise NotImplementedError
+
+    def iter(self, iterable: Iterable[T], *,
+            total: Optional[int] = None,
+            desc: Optional[str]
+    ) -> Iterable[T]:
+        if total is None:
+            if hasattr(iterable, '__len__'):
+                total = len(iterable)
```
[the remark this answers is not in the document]
zhiltsov-max: I'm not sure, because it can have unexpected side effects. For extractors it could start caching. I consider `total` as an override value.

datumaro/components/progress_reporting.py
```diff
@@ -0,0 +1,45 @@
...
+
+        self.start(total, desc=desc)
+
+        if total:
+            display_step = math.ceil(total * self.get_frequency())
+        else:
+            display_step = None
```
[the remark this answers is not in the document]
zhiltsov-max: Agree, removed

datumaro/components/progress_reporting.py
```diff
@@ -0,0 +1,45 @@
...
+# SPDX-License-Identifier: MIT
+
+from typing import Iterable, Optional, TypeVar
+import math
+
+T = TypeVar('T')
+
+class ProgressReporter:
+    def get_frequency(self) -> float:
+        raise NotImplementedError
```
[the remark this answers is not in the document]
zhiltsov-max: I don't see reasons to provide such option currently.

IRDonch (changes requested):

datumaro/components/progress_reporting.py
```diff
@@ -0,0 +1,69 @@
...
+class ProgressReporter:
+    def get_frequency(self) -> float:
+        """
+        Returns reporting frequency.
+
+        For example, 0.1 would mean every 10%.
+        """
+        raise NotImplementedError
+
+    def start(self, total: int, *, desc: Optional[str] = None):
```
IRDonch, line 19: IMO, the way that multiple progress bars are handled needs to be reworked. The current approach has two problems:

1. There is no way for the client to gauge how much progress has been made in total, because the caller doesn't know how many times the service (by which I mean the component reporting the progress, such as an extractor) will call `start`. This will be a problem if Datumaro is integrated into a bigger pipeline which only allows reporting progress as a single fraction, such as Accuracy Checker.

2. It's not possible to have multiple concurrently-updating progress bars, because the `report_status` and `finish` functions don't have enough information to determine which bar is being updated.

I propose the following redesign that fixes these problems:

* Remove the `desc` argument from `start` and `iter`.
* Add the following method:

  ```python
  def split(self, subtasks: Sequence[str]) -> Tuple[ProgressReporter, ...]: ...
  ```

  This method tells the reporter that the current task consists of several subtasks (each element of `subtasks` is the description of a subtask) and returns one progress reporter for each subtask.

* Stipulate that the service must call one and only one of the following methods on its reporter:
  * `start` - if it can subdivide its work into N homogenous units. If it calls `start`, then it must use `report_status` to report on the completion of these units.
  * `split` - if it can subdivide its work into heterogenous units (subtasks). If it calls `split`, then no more methods on the original reporter should be called. Instead, the returned reporters must be used by following the rules recursively.

datumaro/plugins/coco_format/converter.py
```diff
@@ -658,17 +658,27 @@ def apply(self):
...
+                    for task_conv in task_converters.values():
+                        try:
+                            task_conv.save_image_info(item,
+                                self._make_image_filename(item))
+                            task_conv.save_annotations(item)
+                        except Exception as e:
+                            self._report_annotation_error(e,
+                                item_id=(item.id, item.subset))
+                except Exception as e:
+                    self._report_item_error(e, item_id=(item.id, item.subset))
```
IRDonch, line 681: This pattern for exception handling seems quite common; perhaps it would be useful to factor it out as a context manager? E.g.:

```python
with self._item_error_handler(item_id=(item.id, item.subset)):
    if self._save_images: ...
        ...
```

zhiltsov-max: Unfortunately it has 2 major disadvantages:

- It will pessimize annotation import performance for no reason (we expect 1M calls):

```
def f1():
    try:
        x = 1
    except Exception as e:
        raise

def f2():
    x = 1

@contextlib.contextmanager
def cm():
    try:
        yield
    finally:
        pass

def f3():
    with cm():
        x = 1

print(timeit.timeit('f1()', globals=globals(), number=1000000))
print(timeit.timeit('f2()', globals=globals(), number=1000000))
print(timeit.timeit('f3()', globals=globals(), number=1000000))

>>>
0.06721735099563375
0.05293095999513753
1.3045302460086532
```

- It will only cover simplest cases like this. In fact, it is expected that with this functionality we'll be able to take different actions (more than just skipping the problem) depending on eg. user choice or predefined config. Looks quite optimistic IMO, but anyway. If we'll support multiple actions, the `except` clause looks more appropriate choice. I think such change would be a premature optimization.

datumaro/components/extractor.py
```diff
@@ -170,22 +176,91 @@ def get(self, id, subset=None):
...
 class Extractor(_ExtractorBase, CliPlugin):
     """
     A base class for user-defined and built-in extractors.
     Should be used in cases, where SourceExtractor is not enough,
     or its use makes problems with performance, implementation etc.
     """
 
+    def __init__(self, *,
+            length: Optional[int] = None,
+            subsets: Optional[Container[str]] = None,
```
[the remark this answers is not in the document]
zhiltsov-max: Changed to sequence

datumaro/components/extractor.py
```diff
@@ -170,22 +176,93 @@ def get(self, id, subset=None):
...
+        """
+        raise NotImplementedError
+
+    def fail(self, error: Exception) -> NoReturn:
+        raise _ImportFail from error
+
+@define(eq=False)
+class ImportContext:
+    progress_reporter: Optional[ProgressReporter] = None
+    error_policy: Optional[ImportErrorPolicy] = None
```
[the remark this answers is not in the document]
zhiltsov-max: From my point, the extra methods hide unnecessary details of underlying implementation for extractor code. Yes, with the current solution we have some extra functions in the base class, but they encapsulate logic for client code in extractors and converters. Introduction of null objects does not seem to change situation significantly in this case.

datumaro/components/progress_reporting.py
zhiltsov-max: Yeah, it makes sense. It still can be hard to solve the first problem with the proposed design. The second problem looks mostly theoretical at this point.

datumaro/components/extractor.py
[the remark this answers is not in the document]
zhiltsov-max: I decided to remove it and add docs

datumaro/components/dataset.py
```diff
@@ -952,13 +1006,36 @@ def import_from(cls, path: str, format: Optional[str] = None, *,
...
+            if ctx:
+                assert 'ctx' not in extractor_kwargs
+                extractor_kwargs['ctx'] = ctx
+
+            try:
+                extractors.append(env.make_extractor(
+                    src_conf.format, src_conf.url, **extractor_kwargs
+                ))
+            except TypeError as e:
+                if "unexpected keyword argument 'ctx'" in str(e):
```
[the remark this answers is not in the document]
IRDonch: That sounds good to me.

datumaro/plugins/coco_format/converter.py
IRDonch: Okay, fair enough. Although FWIW, a 1 second slowdown for a _million_ items doesn't seem bad at all.

datumaro/components/progress_reporting.py
IRDonch: > It still can be hard to solve the first problem with the proposed design.

What do you mean? It seems fairly easy to calculate:

* If the service has called `start`, then the progress is the current number of completed units divided by the total number of units (if the service didn't specify the total number, then the progress can only be approximated as 0 before the call to `finish` and 1 after).
* If the service has called `split`, then the progress is the sum of the progresses of the child reporters, divided by the number of child reporters.
* If the service has called neither, the progress is 0.

datumaro/components/progress_reporting.py
```diff
@@ -0,0 +1,69 @@
...
+
+            i = 0
+            for i, elem in enumerate(iterable):
+                if not total or i % display_step == 0:
+                    self.report_status(i)
+
+                yield elem
+
+            if i:
+                self.report_status(i)
```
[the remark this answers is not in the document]
IRDonch: Then again, maybe it would be even better to just remove this. The `finish` call that happens afterwards already signifies to the reporter that all units were completed.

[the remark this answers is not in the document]
zhiltsov-max: It is a topic for the next PR, but when I experimented with CLI and tqdm, I found out that if progress bars are not cleaned, they may hang with 99% progress, and it looks clumsy.

datumaro/components/extractor.py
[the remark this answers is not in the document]
IRDonch: IMO, even if the extra methods stay, null objects will still be beneficial. They will make the code simpler due to removal of `None` checks, and they will make the default policy clearer by having a dedicated object implementing it.

datumaro/components/progress_reporting.py
zhiltsov-max: Yes, it is basically correct, but the problem is that if we only have a single pbar, children pbars do one of the following:
- Commit to the single pbar. Depending if we know total or not, there are mixed variants possible. For instance, after starting a new pbar, we can get a new total and resize the pbar, or we can get the situation that we overfill the previous pbar (if we knew total, but now we don't). If we know all totals, we can sum them, of course, but it won't mean that units are equal.
- Split pbar into segments and fill them independently
- Rewrite pbar

I don't consider parallel processing here, but it can be a problem, in theory. Generally, I like the idea, but I'm just saying it's not a silver bullet.

datumaro/components/progress_reporting.py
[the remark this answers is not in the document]
IRDonch: How about at least changing it to an abstract property? Seems like it has property semantics.

datumaro/components/progress_reporting.py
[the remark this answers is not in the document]
IRDonch: This just means that it should be the responsibility of `finish` to clean them.

datumaro/components/progress_reporting.py
zhiltsov-max: I've implemented this idea.

datumaro/components/dataset.py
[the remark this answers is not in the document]
zhiltsov-max: Done

datumaro/plugins/coco_format/converter.py
zhiltsov-max: It can be useful it general, because we may need to do some cleanup on skipped export errors, and we also may want to always close the reporter.

datumaro/components/extractor.py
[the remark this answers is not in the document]
zhiltsov-max: Added

datumaro/components/progress_reporting.py
```diff
@@ -0,0 +1,93 @@
...
+            if total:
+                display_step = math.ceil(total * self.get_frequency())
+
+            for i, elem in enumerate(iterable):
+                if not total or i % display_step == 0:
+                    self.report_status(i)
+
+                yield elem
+        finally:
+            self.finish()
```
IRDonch, line 67: I think the call to `finish` should just follow the loop and not be in a `finally` block, for a couple reasons:

1. If the loop terminates with an exception, then the task did _not_ finish. The client and/or the reporter can use the lack of a `finish` call to show that the task was not completed.
2. Even ignoring the above, this is unreliable. If an exception is raised in the caller of `iter` during iteration, then the `finally` block will not be executed until the generator is garbage-collected, and it's undefined when that happens.

zhiltsov-max: Fixed. I've added `close` and CM semantics.

zhiltsov-max:
@IRDonch, please check the PR.

datumaro/components/progress_reporting.py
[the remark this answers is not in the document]
zhiltsov-max: Done

datumaro/components/extractor.py
```diff
@@ -170,22 +176,93 @@ def get(self, id, subset=None):
...
+            desc: Optional[str] = None
+    ) -> Iterable[T]:
+        if self._ctx and self._ctx.progress_reporter:
+            yield from self._ctx.progress_reporter.iter(iterable,
+                total=total, desc=desc)
+        else:
+            yield from iterable
+
+    def _report_item_error(self, error: Exception, *,
+            item_id: Tuple[str, str]) -> Optional[NoReturn]:
```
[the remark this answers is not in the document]
zhiltsov-max: It might be a good topic for another PR, because this pattern occurs frequently, but it's not it the scope of this PR.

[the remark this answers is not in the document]
IRDonch: Okay, fair enough.

datumaro/components/extractor.py
```diff
@@ -170,22 +178,97 @@ def get(self, id, subset=None):
...
+        self.fail(error)
+
+@define(eq=False)
+class ImportContext:
+    progress_reporter: ProgressReporter = field(default=None,
+        converter=attr.converters.default_if_none(factory=NullProgressReporter))
+    error_policy: ImportErrorPolicy = field(default=None,
+        converter=attr.converters.default_if_none(factory=FailingImportErrorPolicy))
+
+class NullImportContext(ImportContext):
```
IRDonch, line 240: This class seems useless. It's only used once and in that instance it could just be replaced by `ImportContext`.

datumaro/components/progress_reporting.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def __exit__(self, exc_type=None, exc_value=None, traceback=None):
+        self.close()
+
+    @property
+    def frequency(self) -> float:
+        """
+        Returns reporting frequency.
+
+        For example, 0.1 would mean every 10%.
```
IRDonch, line 38: Doesn't this actually make it the _period_, not the frequency?

datumaro/components/dataset.py
```diff
@@ -952,17 +1031,59 @@ def import_from(cls, path: str, format: Optional[str] = None, *,
...
             raise UnknownFormatError(format)
 
-        extractors = []
-        for src_conf in detected_sources:
-            if not isinstance(src_conf, Source):
-                src_conf = Source(src_conf)
-            extractors.append(env.make_extractor(
-                src_conf.format, src_conf.url, **src_conf.options
-            ))
+        # TODO: probably, should not be available in lazy mode, because it
```
IRDonch, line 1034: Why is this a TODO? It seems to already be done.

datumaro/components/extractor.py
```diff
@@ -170,22 +178,97 @@ def get(self, id, subset=None):
...
+    def _handle_item_error(self, error: ItemImportError):
+        """This function must either call fail() or return."""
+        raise NotImplementedError("Must be defined in the subclass")
+
+    def _handle_annotation_error(self, error: AnnotationImportError):
+        """This function must either call fail() or return."""
+        raise NotImplementedError("Must be defined in the subclass")
+
+    def fail(self, error: Exception) -> NoReturn:
+        raise _ImportFail from error
```
IRDonch, line 224: Why is `_ImportFail` necessary? What if this function just raised `error`?

datumaro/components/extractor.py
```diff
@@ -170,22 +178,97 @@ def get(self, id, subset=None):
...
 
+    def __init__(self, *,
+            length: Optional[int] = None,
+            subsets: Optional[Sequence[str]] = None,
+            ctx: Optional[ImportContext] = None):
+        super().__init__(length=length, subsets=subsets)
+
+        if ctx is None:
+            ctx = NullImportContext()
+        self._ctx: ImportContext = ctx
```
IRDonch, line 258:
```suggestion
        self._ctx: ImportContext = ctx or ImportContext()
```

datumaro/components/progress_reporting.py
```diff
@@ -0,0 +1,111 @@
...
+        for i, elem in enumerate(iterable):
+            if not total or i % display_step == 0:
+                self.report_status(i)
+
+            yield elem
+
+    def split(self, count: int) -> Tuple[ProgressReporter, ...]:
+        """
+        Splits the progress bar into few independent parts
+        In case of 0 must return an empty tuple.
```
IRDonch, line 86: Why does 0 need to be supported? It doesn't really make sense to split something into 0 parts.

datumaro/plugins/coco_format/extractor.py
```diff
@@ -125,72 +125,77 @@ def _load_person_kp_categories(self, json_cat):
 
         self._categories[AnnotationType.points] = categories
 
+    @scoped
     def _load_items(self, json_data):
+        pbars = scope_add_many(*self._ctx.progress_reporter.split(2))
```
IRDonch, line 130: It would be more readable to assign each bar to its own variable.

datumaro/plugins/yolo_format/extractor.py
```diff
@@ -186,9 +188,12 @@ def _load_categories(names_path):
 
         return label_categories
 
-    def __iter__(self):
-        for subset in self._subsets.values():
-            for item in subset:
+    @scoped(arg_name='scope') # pylint: disable=no-value-for-parameter
+    def __iter__(self, *, scope: Scope = None):
```
IRDonch, line 192: Useless default argument?

datumaro/util/scope.py
```diff
@@ -88,6 +88,15 @@ def add(self, cm: ContextManager[T]) -> T:
 
         return self._stack.enter_context(cm)
 
+    def add_many(self, *cms: ContextManager[T]) -> Tuple[T, ...]:
+        """
+        Enters few context managers and adds them to the exit stack.
```
IRDonch, line 93:
```suggestion
        Enters several context managers and adds them to the exit stack.
```

datumaro/util/scope.py
```diff
@@ -88,6 +88,15 @@ def add(self, cm: ContextManager[T]) -> T:
 
         return self._stack.enter_context(cm)
 
+    def add_many(self, *cms: ContextManager[T]) -> Tuple[T, ...]:
```
IRDonch, line 91: It's worth noting that in every non-test call of this, the argument is an unpacked sequence. So maybe it would be better to change the argument to accept a sequence to start with?

datumaro/components/dataset.py
zhiltsov-max: The idea is that laziness is not an input parameter in the current implementation, and the extractors are not aligned on this topic. Ideally, it should be a function parameter and should also be aligned with the `eager_mode` CM.

datumaro/components/extractor.py
zhiltsov-max: It is used to indicate exit from importing process. Other exceptions can be captured by other methods of this class.

datumaro/components/progress_reporting.py
zhiltsov-max: Basically, it really helps writing clean client code and does not contradict the function logic.

datumaro/components/progress_reporting.py
IRDonch: Okay, but what about about my proposal about moving the description argument to `split`? I think that makes more sense, because if the service doesn't call `split`, then it doesn't need to supply a description - the client already knows what it's going to be doing. For example, if the client calls a converter, then the client can always label the progress bar "Exporting dataset", because that's what converters do.

On the other hand, descriptions are necessary when you're splitting the task into subtasks, since the client doesn't know what each subtask represents.

datumaro/plugins/yolo_format/extractor.py
zhiltsov-max: It just suppresses the linter message, because iter cannot have arguments. But we need an argument here (unless we go for code generation), because otherwise it is not being captured by a generator (which appears from yield) (see #661).

datumaro/util/scope.py
zhiltsov-max: Currently it is only used in a single context, so I can't say whether one variant is better than another. Maybe, we should just put `*` into `add()` to make it a "drop-in" replacement for `with`.

datumaro/components/progress_reporting.py
IRDonch: Why does it need CM semantics, though? If any progress bars are not completed, then the client can deal with that. Requiring services to close their own progress bars just serves to complicate their code.

datumaro/components/converter.py
```diff
@@ -1,21 +1,91 @@
...
+        raise NotImplementedError("Must be defined in the subclass")
+
+    def _handle_annotation_error(self, error: AnnotationExportError):
+        """This function must either call fail() or return."""
+        raise NotImplementedError("Must be defined in the subclass")
+
+    def fail(self, error: Exception) -> NoReturn:
+        raise _ExportFail from error
+
+class FailingExportErrorPolicy(ExportErrorPolicy):
```
IRDonch, line 73: Perhaps the failing behavior should just be the default? That way, a custom policy that wants to only ignore a certain kind of error will only need to override one method.

It'll also allow other kinds of errors to be added in the future without breaking compatibility.

datumaro/components/extractor.py
IRDonch: > Other exceptions can be captured by other methods of this class.

Isn't that a good thing, though? Imagine that you want to ignore items that failed to be imported for any reason (including bad annotations). You could write:

```
    def _handle_item_error(self, error: ItemImportError):
        pass

    def _handle_annotation_error(self, error: AnnotationImportError):
        ???
```

But if `_handle_annotation_error` calls `fail`, then the whole import fails; while if it does nothing, the _annotation_ is ignored, and the item is still used. It seems like it would be better if `_handle_item_error` got a shot at handling the error raised by `_handle_annotation_error`.

datumaro/components/progress_reporting.py
IRDonch: Can you show me an example?

datumaro/components/progress_reporting.py
zhiltsov-max: Yes, probably

datumaro/components/extractor.py
zhiltsov-max: From my point, it clearly designates the caller intention, while provision of defaults in the base class is more of a forward compatibility.

datumaro/components/extractor.py
zhiltsov-max: This is a good question. The behavior is going to be quite inconsistent, because we can have any of the following cases in different formats:
```
[try] load item()
    [try] load annotations()

[try] load item()
[try] load annotations()

[try] load item and annotations()
```
And if we have eg. handle_ann -> fail, handle_item -> skip, then in the second case we'll fail, then in the first and the third we will skip. If we don't capture errors, we can also do any of them. If we assume the format is "right" and allows this, then yes. Probably, we can just throw `error`, but then also allow to capture the error in the `report_item_error`. We still need to wrap errors in dedicated classes to indicate their kinds.

datumaro/components/progress_reporting.py
zhiltsov-max: Probably, it's reasonable to always split the input pbar, so the caller title is not going to be used at all. I don't have enough arguments for or against requiring titles in `split`, but it looks more flexible to just ask the number.

datumaro/components/converter.py
zhiltsov-max: Yes, it makes sense. Done

IRDonch: I don't think you pushed it.

datumaro/components/extractor.py
IRDonch: I don't really understand the nuance here. Is there any circumstance where `ImportContext()` would act any differently from `NullImportContext()`? You're just giving the user a pointless choice.

In addition, this enables users to write nonsense like `NullImportContext(error_policy=...)`, which seems undesirable.

datumaro/components/progress_reporting.py
IRDonch: You have to destroy threads, because otherwise they'll be leaked. But nothing has to get leaked here. The client has a reference to the progress reporter, so it can clean it up, and the progress reporter can keep track of its children and clean them up too.

There are a lot more services (extractors, converters) than there are clients, so the burden of dealing with resources should fall on the latter. That way, the added complexity of services can be minimized.

datumaro/components/extractor.py
zhiltsov-max: > I don't really understand the nuance here. Is there any circumstance where ImportContext() would act any differently from NullImportContext()? You're just giving the user a pointless choice.

I see it that way that ImportContext is just a base class. It doesn't have to be usable, though it can provide convenient defaults. 

> In addition, this enables users to write nonsense like NullImportContext(error_policy=...), which seems undesirable.

Why they would need to do this?

datumaro/util/scope.py
zhiltsov-max: They still have literally the same text.

datumaro/components/converter.py
zhiltsov-max: No, it's in 
[Make failing the default action](https://github.com/openvinotoolkit/datumaro/pull/650/commits/8736ea23284ef738767b31c6650b1283513c8d05)
[8736ea2](https://github.com/openvinotoolkit/datumaro/pull/650/commits/8736ea23284ef738767b31c6650b1283513c8d05)

datumaro/util/scope.py
zhiltsov-max: Removed this funcitonality

datumaro/plugins/yolo_format/extractor.py
zhiltsov-max: After experimenting and thinking over, I decided to remove closing from plugins, but restore `finish`. In the end of the day, it feels natural for pbar (and ui, possible) to manage its children objects. Finishing is still important for signaling the finish of the loading in a pbar. In case of error, we aren't going to have any new pbars. There is still a question if we need a dedicated close (to close an unfinished pbar), but it is not that important now.

PR #663, opened 2022-02-11
voc_format: don't create XML files with no objects on export

### LINKED ISSUE #658: Format detection fails to disambiguate labelme and voc on tfds voc2012
When I try to detect format of a tfds/voc2012, Datumaro detects both LabelMe and VOC. Probably, the problem is in XMLs with only image info inside, like this:

<details>

```
<annotation>
  <folder>2008</folder>
  <filename>2008_000001.jpg</filename>
  <source>
    <database>Unknown</database>
    <annotation>Unknown</annotation>
    <image>Unknown</image>
  </source>
  <segmented>0</segmented>
</annotation>
```

</details>

IRDonch:
### Summary

The original VOC datasets (at least the 2007 and 2012 versions that I have checked) do not contain XML files with no defined objects. Where an image is not annotated with objects, the corresponding XML file is simply omitted.

This change makes the Datumaro VOC converter consistent with that convention by avoiding dumping the XML files when there are no objects associated with a dataset item.

Beside consistency, the other benefit of this is that written datasets can be disambiguated from the LabelMe format by the format detection machinery. The VOC and LabelMe formats seem to only differ by how annotations are represented inside the `object` element, so if there are no objects, the resulting XML file is ambiguous. By avoiding writing such XML files, we
avoid the ambiguity.

Fixes #658.

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/plugins/voc_format/converter.py
+++ b/datumaro/plugins/voc_format/converter.py
@@ -358,13 +358,11 @@ class VocConverter(Converter):
                 if len(attrs_elem):
                     obj_elem.append(attrs_elem)
 
-        if self._tasks & {VocTask.detection, VocTask.person_layout,
-                            VocTask.action_classification}:
-            ann_path = osp.join(self._ann_dir, item.id + '.xml')
-            os.makedirs(osp.dirname(ann_path), exist_ok=True)
-            with open(ann_path, 'w', encoding='utf-8') as f:
-                f.write(ET.tostring(root_elem,
-                    encoding='unicode', pretty_print=True))
+        ann_path = osp.join(self._ann_dir, item.id + '.xml')
+        os.makedirs(osp.dirname(ann_path), exist_ok=True)
+        with open(ann_path, 'w', encoding='utf-8') as f:
+            f.write(ET.tostring(root_elem,
+                encoding='unicode', pretty_print=True))
 
         return objects_with_parts, objects_with_actions
 
--- a/datumaro/plugins/voc_format/converter.py
+++ b/datumaro/plugins/voc_format/converter.py
@@ -265,6 +265,21 @@ class VocConverter(Converter):
                 self.save_segm_lists(subset_name, segm_list)
 
     def _write_xml_objects(self, item, image_filename, bboxes, has_masks):
+        objects_with_parts = []
+        objects_with_actions = defaultdict(dict)
+
+        main_bboxes = []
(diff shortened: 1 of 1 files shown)


IRDonch:
@zhiltsov-max I haven't finished this (need to write tests still), but could you let me know if the solution seems reasonable to you? Maybe I missed some reason why empty XML files are required.

Note that the actual solution is in the last commit; the rest are just refactoring.

zhiltsov-max:
As far as I remember, the idea was to support the test subset and to export empty image metainfo from CVAT. There should even be an option or a check in the converter.

IRDonch:
> As far as I remember, the idea was to support the test subset

I don't think it's necessary to write object-less XML files to support the test subset - after all, the test subset in the original dataset doesn't include them.

> and to export empty image metainfo from CVAT.

What does this mean?

> There should even be an option or a check in the converter.

I see an option to write empty subset lists, but that seems unrelated.

zhiltsov-max:
> What does this mean?

It means that users often asked to export unannotated images from CVAT along with annotated ones. In case when we don't export images, it allows to keep image size info.

IRDonch:
> For VOC, in case when we don't export images, it allows to keep image size info.

I suppose so, but what use is the image size without the image itself (especially if there are no annotations either)?

zhiltsov-max:
> I suppose so, but what use is the image size without the image itself (especially if there are no annotations either)?

The reasons can be different, but I don't think it's relevant to the format detection question. Speaking about the solution, maybe we could check another file or return a lower confidence value, if we see there are VOC directories in the dataset? Or maybe, we could return higher confidence for VOC.

zhiltsov-max:
I suggest to look at confidence levels and define them the following way:
- High (exactly the dataset): 
  - there are only expected files (secondary can be missing, labelmap and meta should be considered normal)
  - they have right names and placement
  - their formats match the expected format (mandatory fields present, no extra fields, but it is negotiable. Just for upper-level fields)
- Medium (looks like a custom dataset with extra info): 
  - there are extra files in the directory (subformats must use the detection of the whole dataset)
  - required file formats look like expected (mandatory fields present)
- Low (can be parsed): 
  - there are extra files or directories (non-system ones)
  - required files present
  - their formats look like expected (mandatory fields present)

This way we're going to have High for VOC and Medium (because there are extra files) or even Low (because of extra directories) for LabelMe.

PR #664, opened 2022-02-11
labelme_format: try harder to disambiguate from VOC

### LINKED ISSUE #658: Format detection fails to disambiguate labelme and voc on tfds voc2012

IRDonch:
### Summary
A dataset in the VOC format may contain XML files with no objects, which will be indistinguishable from LabelMe files. Instead of matching immediately in this case, keep checking files until either one of them can be definitively determined to be in the LabelMe format or not, or all files are exhausted.

Resolves #658.

### Checklist

### License

```python
#
```


tests/assets/voc_dataset/voc_dataset2/ImageSets/Action/trainval.txt
```diff
@@ -1,3 +1,4 @@
+@
```
zhiltsov-max, line 1: What is the purpose in calling this file this way?

IRDonch: It's so that it would come before A in the sorted list. I should probably add a comment about this.

tests/test_format_detection.py
```diff
@@ -109,6 +109,34 @@ def detect(context):
         self.assertIn('*.txt', result.exception.failed_alternatives[0])
         self.assertIn('*.lst', result.exception.failed_alternatives[0])
 
+    @mark_requirement(Requirements.DATUM_GENERAL_REQ)
+    def test_require_files_success(self):
```
zhiltsov-max, line 113: I can't say it's a blocker, but consider splitting the detector tests into more specific ones. For example, this test currently covers at least 4 statements:
- sorting order of files
- exclusion of files
- confidence level on success 
- success itself

tests/assets/voc_dataset/voc_dataset2/ImageSets/Action/trainval.txt
zhiltsov-max: Can we have a dedicated test for this?

tests/test_format_detection.py
IRDonch: I can only agree with regards to the 3rd bullet point - `test_empty_detector` already tests the default confidence level, so it's not necessary to check it in most other tests. I just added the asserts on the result value because I could. As for the rest:

* Success itself is impossible _not_ to test, since it's tested implicitly by virtue of having the call to `apply_format_detector` be unprotected by a `try` block.
* It's true that exclusion could be tested by a separate test. however, I didn't want to create a bunch of new tests, because the exclusion functionality is shared between `require_file` and `require_files` and so it's already tested by `require_file` tests. The reason I added an `exclude_fnames` parameter to the test at all is to make sure that the function forwards the value to the shared code.

IRDonch: I removed the redundant confidence level checks.

tests/assets/voc_dataset/voc_dataset2/ImageSets/Action/trainval.txt
IRDonch: I expanded the comment inside `@.xml`.

zhiltsov-max: It's quite easy to overlook breaking changes, which will lead to regression in this question. There is no specific test for sorting, and if we remove sorting everything can work well depending on platform logic for returned values from file list fetching. VOC auto detection is a selling feature for some customers, so I think it's better to be protected with a bunch of tests around, stating the expected behavior and making sure we covered corner cases. I'll leave it up to you to decide on this, maybe it will be better to add such tests when we cover this case by some kind of logic described in my proposal.

IRDonch: > There is no specific test for sorting, and if we remove sorting everything can work well depending on platform logic for returned values from file list fetching.

We unfortunately cannot decisively verify that sorting works, since we can't be sure that the filesystem on the testing machine doesn't give us the directory entries in sorted order to begin with. `ApplyFormatDetectorTest.test_require_files_success` is the best we can do WRT verifying this.

However, even if the sorting is accidentally broken and we don't catch it, that won't cause a regression with regards to #658. The sorting only exists to make the detector more deterministic (and thus make the tests more deterministic as well). For a regression to happen, the LabelMe detector would _also_ have to be broken (but in such a way that both `VocImportTest.test_can_detect_voc` and `LabelMeImporterTest.test_can_detect` still pass). So I think we're protected against regression as well as we can be.

Also, this argument doesn't really seem to support your original suggestion - a dedicated test would just have the same problem.

> VOC auto detection is a selling feature for some customers, so I think it's better to be protected with a bunch of tests around, stating the expected behavior and making sure we covered corner cases.

I don't necessarily disagree, but this isn't a general "improve VOC detection testing" PR - I'm just trying to fix a specific scenario (#658). And since I can piggyback on an existing test dataset, I don't really see why I shouldn't do that. I think the new test annotation file is reasonably documented, so there should be no confusion as to what it's for.

> The example is that we have only XML files with image metadata.

That would be a different scenario, so I think it's off-topic.

PR #665, opened 2022-02-14
format_detection: add a dedicated function to report unsupported detection

IRDonch:
### Summary
Currently, instances where detection for a format is unsupported are reported via `context.fail`. However, `context.fail` is intended primarily for situations where the dataset fails some requirement, and using it for this leads to unnatural-looking messages:

    The following requirement was not met:
      this format cannot be autodetected

Fix this by adding a dedicated rejection reason for this case and a new method, `context.raise_unsupported` to abort detection for this reason.

### Checklist

### License

```python
#
```

### DIFF
--- a/tests/test_format_detection.py
+++ b/tests/test_format_detection.py
@@ -2,8 +2,9 @@ from unittest import TestCase
 import os.path as osp
 
 from datumaro.components.format_detection import (
-    FormatDetectionConfidence, FormatRequirementsUnmet, RejectionReason,
-    apply_format_detector, detect_dataset_format,
+    FormatDetectionConfidence, FormatDetectionUnsupported,
+    FormatRequirementsUnmet, RejectionReason, apply_format_detector,
+    detect_dataset_format,
 )
 from datumaro.util.test_utils import TestDir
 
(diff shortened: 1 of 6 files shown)


tests/test_format_detection.py
```diff
@@ -202,6 +203,14 @@ def detect(context):
         self.assertEqual(result.exception.failed_alternatives,
             ('bad alternative 1', 'bad alternative 2'))
 
+    @mark_requirement(Requirements.DATUM_GENERAL_REQ)
+    def test_raise_unsupported(self):
+        def detect(context):
+            context.raise_unsupported()
```
zhiltsov-max, line 209: It looks like it should be the default implementation, shouldn't it?

IRDonch: I considered this, but I think the current default implementation (falling back to `find_sources`) is more useful.

zhiltsov-max: When it's supposed to be called then?

IRDonch: I have added calls to it in this PR.

zhiltsov-max (approved):

tests/test_format_detection.py
zhiltsov-max: Ok, I expect it to be redesigned quite soon though.

PR #668, opened 2022-02-16
Save box attributes when downloading the TFDS version of VOC

IRDonch:
### Summary
The main nuisance with this is that Datumaro represents the pose as a string, while TFDS represents it as an index, and to convert the latter into the former, we need access to the list of poses.

To implement this, allow adapters access to a persistent state object that can be used to store and retrieve arbitrary data during conversion. We can then save the list of names during category creation and retrieve it when we convert box annotations.

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/components/extractor_tfds.py
+++ b/datumaro/components/extractor_tfds.py
@@ -99,23 +99,6 @@ class _SetLabelCategoriesFromClassLabelFeature:
         categories[AnnotationType.label] = LabelCategories.from_iterable(
             feature_connector.names)
 
-
-@frozen
-class _SaveFeatureClassList:
-    feature_path: _FeaturePath
-    state_key: str
-
-    def __call__(self,
-        tfds_builder: tfds.core.DatasetBuilder, categories: CategoriesInfo,
-        state: namespace,
-    ) -> None:
-        feature_connector = _resolve_feature_path(
-            self.feature_path, tfds_builder.info.features)
-
-        assert isinstance(feature_connector, tfds.features.ClassLabel)
-        setattr(state, self.state_key, feature_connector.names)
-
-
 @frozen
 class _SetImageFromImageFeature:
     feature_name: str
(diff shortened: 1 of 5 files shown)


datumaro/components/extractor_tfds.py
```diff
@@ -102,20 +137,38 @@ def __call__(self, tfds_example: Any, item: DatasetItem) -> None:
...
+        self, tfds_example: Any, item: DatasetItem, state: namespace,
+    ) -> None:
         item.annotations.append(
             Label(tfds_example[self.feature_name].numpy()),
         )
 
+@frozen
+class _AttributeMemberMapping:
+    member_name: str
+    attribute_name: str = field()
```
zhiltsov-max, line 150:
```suggestion
    attribute_name: str
```

zhiltsov-max (dismissed):

datumaro/components/extractor_tfds.py
```diff
@@ -217,12 +276,22 @@ def __call__(self, tfds_example: Any) -> str:
...
     data_transformers=[
         _SetImageFromImageFeature('image',
             filename_feature_name='image/filename'),
-        _AddObjectsFromFeature('objects', 'bbox', label_member='label')
+        _AddObjectsFromFeature('objects', 'bbox', label_member='label',
+            attribute_members=(
+                _AttributeMemberMapping('is_difficult', 'difficult'),
+                _AttributeMemberMapping('is_truncated', 'truncated'),
+                _AttributeMemberMapping('pose',
+                    value_converter=lambda idx, state: state.pose_names[idx]),
```
zhiltsov-max, line 292: Probably, it needs to be aligned with https://github.com/openvinotoolkit/datumaro/blob/develop/datumaro/plugins/voc_format/format.py#L50

datumaro/components/extractor_tfds.py
IRDonch: I need the field object to be able to call `attribute_name.default`.

datumaro/components/extractor_tfds.py
IRDonch: The only difference was the case of the string. I changed the adapter to convert the pose names to title case to better match the original dataset.

zhiltsov-max (approved):

PR #669, opened 2022-02-16
Remove usages of distutils from the library

IRDonch:
### Summary
`distutils` has been deprecated in Python 3.10. Fortunately, the only name we're using from there is `strtobool`, and conveniently, `attrs` has recently added a function with the same semantics and the same sets of true and false values. So just use that instead.

Note that `strtobool` is still used in `setup.py`, and we can't use `attrs` there. I'm hoping that setuptools will eventually provide a replacement.

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/plugins/market1501_format.py
+++ b/datumaro/plugins/market1501_format.py
@@ -1,14 +1,14 @@
-# Copyright (C) 2020-2021 Intel Corporation
+# Copyright (C) 2020-2022 Intel Corporation
 #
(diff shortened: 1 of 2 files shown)


zhiltsov-max (approved):
Quite fragile, but ok

zhiltsov-max:
I'd say the fact converters have a single argument is accidental.

IRDonch:
Maybe, but I don't think it can change now without breaking attrs's backwards compatibility policy.

PR #684, opened 2022-03-18
Improve COCO parsing error messages

zhiltsov-max:
### Summary

- Extended dataset parsing error classification
- Added default string conversion for parsing errors
- Added COCO json parsing errors
- Added tests

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/components/errors.py
+++ b/datumaro/components/errors.py
@@ -209,7 +209,7 @@ class InvalidFieldError(InvalidAnnotationError):
     name: str
 
     def __str__(self) -> str:
-        return f"Invalid annotation field {self.name} value"
+        return f"Invalid annotation field '{self.name}' value"
 
 
 @define(auto_exc=False)
@@ -222,7 +222,7 @@ class InvalidFieldTypeError(InvalidFieldError):
             expected = self.expected[0]
         else:
             expected = "one of " + ", ".join(self.expected)
-        return f"Invalid annotation field {self.name} type {self.actual}. Expected {expected}"
+        return f"Invalid annotation field '{self.name}' type '{self.actual}'. Expected '{expected}'"
 
 
 @define(auto_exc=False)
@@ -230,7 +230,7 @@ class MissingFieldError(InvalidAnnotationError):
     name: str
 
     def __str__(self) -> str:
-        return f"Missing annotation field {self.name}"
+        return f"Missing annotation field '{self.name}'"
 
 
 @define(auto_exc=False)
@@ -238,7 +238,7 @@ class InvalidLabelError(InvalidAnnotationError):
     id: str  # index or name
 
     def __str__(self) -> str:
-        return f"Invalid label value {self.id}"
+        return f"Invalid label value '{self.id}'"
 
 
 @define(auto_exc=False)
--- a/datumaro/components/errors.py
+++ b/datumaro/components/errors.py
@@ -234,11 +234,12 @@ class MissingFieldError(InvalidAnnotationError):
 
 
 @define(auto_exc=False)
-class InvalidLabelError(InvalidAnnotationError):
-    id: str  # index or name
+class UndeclaredLabelError(InvalidAnnotationError):
+    id: str = field(validator=not_empty)
+    """index or name"""
 
     def __str__(self) -> str:
-        return f"Invalid label value '{self.id}'"
+        return f"Undeclared label '{self.id}'"
 
 
 @define(auto_exc=False)
--- a/datumaro/components/errors.py
+++ b/datumaro/components/errors.py
@@ -197,21 +197,66 @@ class DatasetImportError(DatumaroError):
     pass
 
 
+class InvalidAnnotationError(DatasetImportError):
+    """
+    A basic dataset parsing error. Should include the problem description in
+    the message.
+    """
+
+
+@define(auto_exc=False)
+class InvalidFieldError(InvalidAnnotationError):
+    name: str
+
+    def __str__(self) -> str:
+        return f"Invalid anotation field {self.name} value"
+
+
+@define(auto_exc=False)
+class InvalidFieldTypeError(InvalidFieldError):
+    actual: str
+    expected: Tuple[str]
+
+    def __str__(self) -> str:
+        if len(self.expected) == 1:
+            expected = self.expected[0]
+        else:
+            expected = "one of " + ", ".join(self.expected)
+        return f"Invalid anotation field {self.name} type {self.actual}. Expected {expected}"
+
+
+@define(auto_exc=False)
+class MissingFieldError(InvalidAnnotationError):
+    name: str
+
+    def __str__(self) -> str:
+        return f"Missing anotation field {self.name}"
+
+
+@define(auto_exc=False)
+class InvalidLabelError(InvalidAnnotationError):
+    id: str  # index or name
+
+    def __str__(self) -> str:
+        return f"Invalid label value {self.id}"
+
+
 @define(auto_exc=False)
 class ItemImportError(DatasetImportError):
     """
-    Represents additional item error info. The error itself is supposed to be
-    in the `__cause__` member.
+    Wraps a dataset parsing error and provides additional error context info.
+    The error itself is supposed to be in the `__cause__` member.
     """
 
     item_id: Tuple[str, str]
 
     def __str__(self):
-        return "Failed to import item %s" % (self.item_id,)
+        return f"Failed to import item {self.item_id}"
 
 
 class AnnotationImportError(ItemImportError):
-    pass
+    def __str__(self):
+        return f"Failed to import item {self.item_id} annotation"
 
 
 @define(auto_exc=False)
--- a/datumaro/components/errors.py
+++ b/datumaro/components/errors.py
@@ -2,9 +2,11 @@
 #
 # SPDX-License-Identifier: MIT
 
-from typing import Any, Tuple
+from typing import Any, Optional, Tuple
 
-from attrs import define, field
+from attrs import define, field, validators
+
+from datumaro.util.attrs_util import has_length, not_empty
 
 
 class ImmutableObjectError(Exception):
@@ -206,7 +208,8 @@ class InvalidAnnotationError(DatasetImportError):
 
 @define(auto_exc=False)
 class InvalidFieldError(InvalidAnnotationError):
-    name: str
+    name: str = field(validator=[validators.instance_of(str), not_empty])
+    """Field name"""
 
     def __str__(self) -> str:
         return f"Invalid annotation field '{self.name}' value"
@@ -214,8 +217,11 @@ class InvalidFieldError(InvalidAnnotationError):
 
 @define(auto_exc=False)
 class InvalidFieldTypeError(InvalidFieldError):
-    actual: str
-    expected: Tuple[str]
+    actual: str = field(validator=[validators.instance_of(str), not_empty])
+    """Actual type of the field"""
+
+    expected: Tuple[str] = field(validator=[validators.instance_of(tuple), not_empty])
+    """The list of expected types of the field"""
 
     def __str__(self) -> str:
         if len(self.expected) == 1:
@@ -226,17 +232,15 @@ class InvalidFieldTypeError(InvalidFieldError):
 
 
 @define(auto_exc=False)
-class MissingFieldError(InvalidAnnotationError):
-    name: str
-
+class MissingFieldError(InvalidFieldError):
     def __str__(self) -> str:
         return f"Missing annotation field '{self.name}'"
 
 
 @define(auto_exc=False)
 class UndeclaredLabelError(InvalidAnnotationError):
-    id: str = field(validator=not_empty)
-    """index or name"""
+    id: str = field(validator=validators.instance_of(str))
+    """Index or name"""
 
     def __str__(self) -> str:
         return f"Undeclared label '{self.id}'"
@@ -249,7 +253,13 @@ class ItemImportError(DatasetImportError):
     The error itself is supposed to be in the `__cause__` member.
     """
 
-    item_id: Tuple[str, str]
+    item_id: Tuple[Optional[str], Optional[str]] = field(
+        validator=[validators.instance_of(tuple), has_length(2)]
+    )
+    """
+    (id, subset) of the item with problem.
+    If id or subset cannot be reported, such field is set to None.
+    """
 
     def __str__(self):
         return f"Failed to import item {self.item_id}"
--- a/datumaro/components/errors.py
+++ b/datumaro/components/errors.py
@@ -209,7 +209,7 @@ class InvalidFieldError(InvalidAnnotationError):
     name: str
 
     def __str__(self) -> str:
-        return f"Invalid anotation field {self.name} value"
+        return f"Invalid annotation field {self.name} value"
 
 
 @define(auto_exc=False)
@@ -222,7 +222,7 @@ class InvalidFieldTypeError(InvalidFieldError):
             expected = self.expected[0]
         else:
             expected = "one of " + ", ".join(self.expected)
-        return f"Invalid anotation field {self.name} type {self.actual}. Expected {expected}"
+        return f"Invalid annotation field {self.name} type {self.actual}. Expected {expected}"
 
 
 @define(auto_exc=False)
@@ -230,7 +230,7 @@ class MissingFieldError(InvalidAnnotationError):
     name: str
 
     def __str__(self) -> str:
-        return f"Missing anotation field {self.name}"
+        return f"Missing annotation field {self.name}"
 
 
 @define(auto_exc=False)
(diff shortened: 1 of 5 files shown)


datumaro/components/errors.py
```diff
@@ -197,21 +197,66 @@ class DatasetImportError(DatumaroError):
...
+    name: str
+
+    def __str__(self) -> str:
+        return f"Invalid anotation field {self.name} value"
+
+
+@define(auto_exc=False)
+class InvalidFieldTypeError(InvalidFieldError):
+    actual: str
+    expected: Tuple[str]
```
IRDonch: You should add a `not_empty` validator here.

datumaro/components/errors.py
```diff
@@ -197,21 +197,66 @@ class DatasetImportError(DatumaroError):
...
+    the message.
+    """
+
+
+@define(auto_exc=False)
+class InvalidFieldError(InvalidAnnotationError):
+    name: str
+
+    def __str__(self) -> str:
+        return f"Invalid anotation field {self.name} value"
```
IRDonch:
```suggestion
        return f"Invalid annotation field {self.name} value"
```

There's a few other instances of this mistake, as well.

datumaro/plugins/coco_format/extractor.py
```diff
@@ -48,7 +57,9 @@ def __init__(
         keep_original_category_ids=False,
         **kwargs,
     ):
-        assert osp.isfile(path), path
+        if not osp.isfile(path):
+            raise DatasetImportError(f"Can't find JSON file path at '{path}'")
```
IRDonch:
```suggestion
            raise DatasetImportError(f"Can't find JSON file at '{path}'")
```

datumaro/plugins/coco_format/extractor.py
```diff
@@ -236,19 +274,37 @@ def __call__(self):
...
-        return self._label_map[ann["category_id"]]
+
+        label_id = self._label_map.get(cat_id)
+        if label_id is None:
+            raise InvalidLabelError(str(cat_id))
+        return label_id
+
+    def _parse_field(
+        self, ann: Dict[str, Any], key: str, cls: Union[None, Type[T], Tuple[Type, ...]]
+    ) -> T:
```
IRDonch: I don't think that's a well-defined type annotation. If the user passes `None` or a tuple of types, then there's no way for a type checker to derive `T`.

Also, `None` seems unnecessary here. I don't think it's used anywhere, and if it ever needs to be used, the same effect can be accomplished by passing `object`.

datumaro/plugins/coco_format/extractor.py
```diff
@@ -236,19 +274,37 @@ def __call__(self):
...
+        value = ann.get(key, NOTSET)
+        if value is NOTSET:
+            raise MissingFieldError(key)
+        elif cls and not isinstance(value, cls):
+            cls = (cls,) if isclass(cls) else cls
```
IRDonch, line 301: FWIW, `isclass(cls)` is just `isinstance(cls, type)`. I don't think there's much reason to use `isclass` in Python 3 (in Python 2 it made more sense, since there were several kinds of classes).

datumaro/plugins/coco_format/extractor.py
```diff
@@ -259,10 +315,16 @@ def _load_annotations(self, ann, image_info=None, parsed_annotations=None):
...
-            attributes["is_crowd"] = bool(ann["iscrowd"])
+            attributes["is_crowd"] = bool(self._parse_field(ann, "iscrowd", int))
 
             if self._task is CocoTask.person_keypoints:
-                keypoints = ann["keypoints"]
+                keypoints = self._parse_field(ann, "keypoints", list)
+                if len(keypoints) % 3 != 0:
+                    raise InvalidAnnotationError(
+                        "Keypoints have invalid value count %s, which is not divisible by 3. "
+                        "Expected (x, y, visibility) triplets." % (len(keypoints),)
```
IRDonch: You seem to be using f-strings everywhere, so why not here?

datumaro/plugins/coco_format/extractor.py
```diff
@@ -281,14 +343,21 @@ def _load_annotations(self, ann, image_info=None, parsed_annotations=None):
...
                 rle = None
 
                 if isinstance(segmentation, list):
                     if not self._merge_instance_polygons:
                         # polygon - a single object can consist of multiple parts
                         for polygon_points in segmentation:
+                            if len(polygon_points) % 2 != 0 or len(polygon_points) < 6:
+                                raise InvalidAnnotationError(
+                                    "Polygon has invalid value count %s, "
+                                    "which is not divisible by 2. "
```
IRDonch: This might not be true, since you have two conditions in the `if`. IMO, better to just have two separate `if`s.

datumaro/plugins/coco_format/extractor.py
```diff
@@ -300,26 +369,20 @@ def _load_annotations(self, ann, image_info=None, parsed_annotations=None):
                             )
                     else:
                         # merge all parts into a single mask RLE
-                        rle = self._lazy_merged_mask(
-                            segmentation, image_info["height"], image_info["width"]
-                        )
+                        img_h = self._parse_field(image_info, "height", (int, float))
+                        img_w = self._parse_field(image_info, "width", (int, float))
```
IRDonch: I don't think `height` and `width` should be allowed to be `float`s.

datumaro/plugins/coco_format/extractor.py
```diff
@@ -300,26 +369,20 @@ def _load_annotations(self, ann, image_info=None, parsed_annotations=None):
...
-                            "item #%s: mask #%s "
-                            "does not match image size: %s vs. %s. "
-                            "Skipping this annotation.",
-                            image_info["id"],
-                            ann_id,
-                            (mask_h, mask_w),
-                            (img_h, img_w),
+                    img_h = self._parse_field(image_info, "height", (int, float))
+                    img_w = self._parse_field(image_info, "width", (int, float))
+                    mask_h, mask_w = self._parse_field(segmentation, "size", list)
```
IRDonch: Perhaps there should be a check for the list size before the unpacking?

datumaro/components/errors.py
```diff
@@ -197,21 +197,66 @@ class DatasetImportError(DatumaroError):
...
+@define(auto_exc=False)
+class MissingFieldError(InvalidAnnotationError):
+    name: str
+
+    def __str__(self) -> str:
+        return f"Missing anotation field {self.name}"
+
+
+@define(auto_exc=False)
+class InvalidLabelError(InvalidAnnotationError):
```
IRDonch: I don't think this needs to be a separate exception:

* I don't see a use case for handling it differently from `InvalidAnnotationError`;
* The default message is too generic to be useful (it doesn't explain why the value is invalid).
* Its semantics overlap with other exceptions (e.g. if a field represents a label, and it has the wrong type, is that an invalid label or an invalid field type?).

It's also only used once. I think that use can just be replaced with an `InvalidAnnotationError` with a custom message.

IRDonch (changes requested):

tests/test_coco_format.py
```diff
@@ -836,6 +847,320 @@ def test_can_pickle(self):
...
+                            {
+                                "id": 5,
+                                "width": 10,
+                                "height": 5,
+                                "file_name": "a.jpg",
+                            }
+                        ],
+                        "annotations": [],
+                        "categories": [],
+                    }
```
IRDonch: Most of these tests are very similar; I think it would be more readable to factor out the common parts into a single function that takes one parameter (a function that mangles the dataset) and returns the resulting exception. Then the individual tests could just look like this:

```python
def mangle(anns):
    del anns["images"][0][field]

exc = self._test_mangled_annotations(mangle)
self.assertIsInstance(capture.exception.__cause__, MissingFieldError)
self.assertEqual(capture.exception.__cause__.name, field)
```

tests/test_coco_format.py
```diff
@@ -836,6 +847,320 @@ def test_can_pickle(self):
...
+                    anns["images"][0].pop(field)
```
IRDonch, line 893: FWIW, `pop` is used to remove a key and return the corresponding value. When you just need to remove a key, the idiomatic way to do that is `del`.

datumaro/plugins/coco_format/extractor.py
zhiltsov-max: Maybe, but I don't see this function being deprecated even in 3.10, so I think, it's totally legitimate. Also, I see it more expressive.

datumaro/plugins/coco_format/extractor.py
zhiltsov-max: I don't have preference to any of these styles.

datumaro/plugins/coco_format/extractor.py
zhiltsov-max: Ok, splitted.

datumaro/plugins/coco_format/extractor.py
zhiltsov-max: Ok, agree. I thought, it can be useful to allow values like 100.0, but it seems like not a good idea.

datumaro/plugins/coco_format/extractor.py
zhiltsov-max: Added

datumaro/components/errors.py
zhiltsov-max: > I don't see a use case for handling it differently from InvalidAnnotationError;

Basically, the idea is to group similar errors and display collective error info in the UI. In general, we'd like to have specialized error types instead of a single generic one.

> The default message is too generic to be useful (it doesn't explain why the value is invalid).
> Its semantics overlap with other exceptions (e.g. if a field represents a label, and it has the wrong type, is that an invalid label or an invalid field type?).

It represents exactly the invalid label value, of the right type. In most cases I expect it to indicate a value, that is not declared in the label list of the dataset. Invalid type should be checked before. Maybe, it's better to rename it to "UndeclaredLabel"?

> It's also only used once. I think that use can just be replaced with an InvalidAnnotationError with a custom message.

I expect it to be applicable to almost all formats.

tests/test_coco_format.py
zhiltsov-max: Agree about the common part extraction, I'll check if it's possible.

tests/test_coco_format.py
zhiltsov-max: Maybe, but I find `del anns["images"][0][field]` really confusing. Are we going to delete a key or call the d-tor on the value?

datumaro/plugins/coco_format/extractor.py
zhiltsov-max: Fixed

datumaro/plugins/coco_format/extractor.py
zhiltsov-max: Fixed

datumaro/components/errors.py
zhiltsov-max: Fixed

datumaro/components/errors.py
zhiltsov-max: Updated errors

datumaro/plugins/coco_format/extractor.py
IRDonch: Yeah, I'm not saying it's illegitimate, just unnecessary. Anyway, it's up to you.

tests/test_coco_format.py
IRDonch: I don't know why you would think it's the latter; Python doesn't have any specific syntax for calling destructors. But it's just a nitpick, anyway, so I'll leave it up to you.

datumaro/components/errors.py
IRDonch: > Basically, the idea is to group similar errors and display collective error info in the UI. In general, we'd like to have specialized error types instead of a single generic one.

Okay, that's fair.

> It represents exactly the invalid label value, of the right type. In most cases I expect it to indicate a value, that is not declared in the label list of the dataset.

The problem is that just "invalid" doesn't really say much.

> Maybe, it's better to rename it to "UndeclaredLabel"?

Okay, that's a more specific error, so I think there's a better case for it.

datumaro/plugins/coco_format/extractor.py
```diff
@@ -236,19 +274,37 @@ def __call__(self):
...
+            raise UndeclaredLabelError(str(cat_id))
+        return label_id
+
+    def _parse_field(
+        self, ann: Dict[str, Any], key: str, cls: Union[Type[T], Tuple[Type, ...]]
+    ) -> T:
+        value = ann.get(key, NOTSET)
+        if value is NOTSET:
+            raise MissingFieldError(key)
+        elif cls and not isinstance(value, cls):
```
IRDonch: The `cls` test is no longer necessary, since `cls` can't be `None` anymore.

datumaro/plugins/coco_format/extractor.py
IRDonch: You fixed the second issue, but not the first.

I think typing this correctly requires `@overload`:

```python
@overload
def _parse_field(self, ann: Dict[str, Any], key: str, cls: Type[T]) -> T: ...

@overload
def _parse_field(self, ann: Dict[str, Any], key: str, cls: Tuple[Type, ...]) -> Any: ...

def _parse_field(
        self, ann: Dict[str, Any], key: str, cls: Union[Type[T], Tuple[Type, ...]]
    ) -> Any:
    # implementation here
```

tests/test_coco_format.py
zhiltsov-max: Done

datumaro/plugins/coco_format/extractor.py
zhiltsov-max: Removed

datumaro/plugins/coco_format/extractor.py
zhiltsov-max: Added overloads

tests/test_coco_format.py
IRDonch: You extracted a shared annotation template (which is certainly a big improvement), but there's still a lot of duplication in the structure of the tests. How about factoring that out too?

tests/test_coco_format.py
```diff
@@ -836,6 +848,205 @@ def test_can_pickle(self):
                 compare_datasets_strict(self, source, parsed)
 
 
+class CocoExtractorTests(TestCase):
+    ANNOTATION_JSON_TEMPLATE = {
```
IRDonch, line 852: IMO, it would be useful to have a test that loads the unmodified template, to ensure that any errors in the other tests are due to the modifications made in those tests and not due to the template itself.

tests/test_coco_format.py
zhiltsov-max: How do you see that?

tests/test_coco_format.py
zhiltsov-max: The fact that the common piece was possible to extract is occasional, it is not something designed or expected. There are lots of tests that cover successful loading.

tests/test_coco_format.py
IRDonch: Like in my code snippet above. In full it would look something like this:

```
    def _load_mangled_annotations(self, mangle):
        with TestDir() as test_dir:
            ann_path = osp.join(test_dir, "ann.json")
            anns = mangle(deepcopy(self.ANNOTATION_JSON_TEMPLATE))
            dump_json_file(ann_path, anns)

            with self.assertRaises(ItemImportError) as capture:
                Dataset.import_from(ann_path, "coco_instances")
                    
            return capture

    def test_can_report_missing_item_field(self):
        for field in ["id", "file_name"]:
            with self.subTest(field=field):
                def mangle(anns):
                    anns["images"][0].pop(field)

                exc = self._load_mangled_annotations(mangle)
                self.assertIsInstance(capture.exception.__cause__, MissingFieldError)
                self.assertEqual(capture.exception.__cause__.name, field)
```

tests/test_coco_format.py
IRDonch: > The fact that the common piece was possible to extract is occasional, it is not something designed or expected.

How is that not expected? The point of the new tests is to make sure that specific errors in the dataset cause the extraction to fail. To be certain that it's _those_ errors that cause the failure, we need to ensure that the original template contains no errors.

tests/test_coco_format.py
zhiltsov-max: Maybe, but it seem to save 1 line of code per test. It also requires to catch `DatasetImportError` and check the type and message in the test.

tests/test_coco_format.py
zhiltsov-max: It is checked by the error details checks. However, I can agree, that such a test could be useful on its own, because there is no such test for just a json parsing.

IRDonch (approved):

PR #685, opened 2022-03-21
CSS dataset format (import only)

yasakova-anastasia:
### Summary

- Added Common Semantic Segmentation dataset format
- Updated `mask_tools.unpaint_mask()` error messages in case of unknown color in the mask

### Checklist

### License

```python
#
```


site/content/en/docs/formats/css.md
```diff
@@ -0,0 +1,81 @@
...
+To add classes, you should use [`dataset_meta.json`](/docs/user-manual/supported_formats/#dataset-meta-file).
+
+To make sure that the selected dataset has been added to the project, you can
+run `datum project info`, which will display the project information.
+
+## Export to other formats
+
+Datumaro can convert a CSS dataset into any other format [Datumaro supports](/docs/user-manual/supported_formats/).
+To get the expected result, convert the dataset to formats
+that support the segmentation task (e.g. PascalVOC, CamVID, CItyscapes, etc.)
```
IRDonch:
```suggestion
that support the segmentation task (e.g. PASCAL VOC, CamVid, CItyscapes, etc.)
```

tests/assets/common_segmentation_dataset/dataset/dataset_meta.json
```diff
@@ -0,0 +1,51 @@
+{
+  "labels": [
+    "Void",
+    "Animal",
+    "Archway",
+    "Bicyclist",
+    "Child",
+    "Road"
+  ],
+  "label_map": {
```
IRDonch: Do we really need both `labels` and `label_map`?

site/content/en/docs/formats/css.md
```diff
@@ -0,0 +1,81 @@
+---
+title: 'Common Semantic Segmentation'
+linkTitle: 'Common Semantic Segmentation'
+description: ''
+weight: 2
+---
+
```
IRDonch: This seems to be lacking background information. Where did this format come from?

datumaro/plugins/common_segmentation_format.py
```diff
@@ -0,0 +1,139 @@
...
+    else:  # only copy defined colors
+        label_id = lambda label: label_categories.find(label)[0]
+        colormap = {label_id(name): (desc[0], desc[1], desc[2]) for name, desc in label_map.items()}
+    mask_categories = MaskCategories(colormap)
+    mask_categories.inverse_colormap  # pylint: disable=pointless-statement
+    categories[AnnotationType.mask] = mask_categories
+    return categories
+
+
+class CommonSegmentationExtractor(SourceExtractor):
```
IRDonch: I'm a bit concerned that the word "semantic" has disappeared from the name of the format. Why didn't you include it?

datumaro/plugins/common_segmentation_format.py
```diff
@@ -0,0 +1,139 @@
...
+        super().__init__(subset=subset)
+
+        self._image_prefix = image_prefix
+        self._mask_prefix = mask_prefix
+
+        if has_meta_file(path):
+            label_map = parse_meta_file(path)
+            self._categories = make_categories(label_map)
+        else:
+            meta_file = glob.glob(osp.join(path, "**", DATASET_META_FILE), recursive=True)
```
IRDonch: This globbing doesn't match the description of the format, according to which the meta file must be in the root. Why is it needed?

datumaro/plugins/common_segmentation_format.py
```diff
@@ -0,0 +1,139 @@
...
+                self._categories = make_categories(label_map)
+
+        self._items = list(self._load_items(path).values())
+
+    def _load_items(self, path):
+        items = {}
+
+        image_dir = osp.join(path, "**", CommonSegmentationPath.IMAGES_DIR)
+
+        if osp.isdir(image_dir):
```
IRDonch: This will almost certainly be false, since `image_dir` has `/**/` in it. Seems like a bug.

datumaro/plugins/common_segmentation_format.py
```diff
@@ -0,0 +1,139 @@
...
+            images = {
+                osp.splitext(osp.relpath(p, image_dir))[0].replace("\\", "/")[
+                    len(self._image_prefix) :
```
IRDonch: Shouldn't this also filter out images that _don't_ have the given prefix?

datumaro/plugins/common_segmentation_format.py
```diff
@@ -0,0 +1,139 @@
...
+                ]: p
+                for p in find_images(image_dir, recursive=True)
+            }
+        else:
+            images = {}
+
+        for mask_path in glob.glob(
+            osp.join(path, "**", CommonSegmentationPath.MASKS_DIR, f"{self._mask_prefix}*.*"),
```
IRDonch: 1. Shouldn't masks always be PNG? You even have the `IMAGE_EXT` constant defined for it.
2. `_mask_prefix` should be escaped with `glob.escape`.

IRDonch (changes requested):

site/content/en/docs/formats/css.md
yasakova-anastasia: I took the description of this format in AC. We can add a link to [it](https://github.com/openvinotoolkit/open_model_zoo/blob/master/tools/accuracy_checker/openvino/tools/accuracy_checker/annotation_converters/common_semantic_segmentation.py).

datumaro/plugins/common_segmentation_format.py
yasakova-anastasia: I decided that the name `CommonSemanticSegmentation` is too long. 

We can use the CSS name, but since the format is not very well known, it will be difficult for most users to understand what it is for by this name.

datumaro/plugins/common_segmentation_format.py
yasakova-anastasia: This is necessary to support use cases with folder nesting. This is present in WB datasets.

datumaro/plugins/common_segmentation_format.py
yasakova-anastasia: You are right, thanks. I fixed it.

datumaro/plugins/common_segmentation_format.py
yasakova-anastasia: Added a condition for this.

datumaro/plugins/common_segmentation_format.py
yasakova-anastasia: 1. Masks can be not only PNG. This information is in the description of the format in AC.
2. Added `glob.escape`

tests/assets/common_segmentation_dataset/dataset/dataset_meta.json
yasakova-anastasia: No. Removed `labels`.

site/content/en/docs/formats/css.md
yasakova-anastasia: Fixed

IRDonch: You only fixed it half-way.

site/content/en/docs/formats/css.md
IRDonch: Was the format created by AC? If so, it would make sense to document that and add a link to the AC documentation.

datumaro/plugins/common_segmentation_format.py
IRDonch: Given that in AC this format is referred to as `common_semantic_segmentation`, I think we should do the same.

datumaro/plugins/common_segmentation_format.py
```diff
@@ -0,0 +1,148 @@
...
+        self._items = list(self._load_items(path).values())
+
+    def _load_items(self, path):
+        items = {}
+
+        image_dir = osp.join(path, CommonSegmentationPath.IMAGES_DIR)
+
+        if not osp.isdir(image_dir):
+            image_dir = glob.glob(
+                osp.join(path, "**", CommonSegmentationPath.IMAGES_DIR), recursive=True
```
IRDonch: Do we really need to search for the image and annotation directories like that? I understand the nested folder support, but shouldn't the image and annotation directories at least be in the same directory as the `dataset_meta.json` file? Since we have already located the meta file at this point, we should be able to just locate the images and annotations relative to it.

datumaro/plugins/common_segmentation_format.py
```diff
@@ -0,0 +1,148 @@
...
+            if image_dir:
+                image_dir = image_dir[0]
+
+        if osp.isdir(image_dir):
+            images = {
+                osp.splitext(osp.relpath(p, image_dir))[0].replace("\\", "/")[
+                    len(self._image_prefix) :
+                ]: p
+                for p in find_images(image_dir, recursive=True)
+                if osp.basename(p)[: len(self._image_prefix)] == self._image_prefix
```
IRDonch: You can use `.startswith()`.

datumaro/plugins/common_segmentation_format.py
IRDonch: > Masks can be not only PNG. This information is in the description of the format in AC.

OK, but in that case, shouldn't you at least use `find_images` in order to filter out non-image files?

datumaro/util/mask_tools.py
```diff
@@ -64,7 +64,7 @@ def unpaint_mask(painted_mask, inverse_colormap=None):
     if callable(inverse_colormap):
         map_fn = lambda a: inverse_colormap((a >> 16) & 255, (a >> 8) & 255, a & 255)
     else:
-        map_fn = lambda a: inverse_colormap[((a >> 16) & 255, (a >> 8) & 255, a & 255)]
+        map_fn = lambda a: inverse_colormap.get(((a >> 16) & 255, (a >> 8) & 255, a & 255), 0)
```
IRDonch: Can you explain this change?

tests/test_common_segmentation_format.py
```diff
@@ -0,0 +1,124 @@
...
+from datumaro.plugins.common_segmentation_format import CommonSegmentationImporter, make_categories
+from datumaro.util.test_utils import compare_datasets
+
+from .requirements import Requirements, mark_requirement
+
+DUMMY_DATASET_DIR = osp.join(
+    osp.dirname(__file__), "assets", "common_segmentation_dataset", "dataset"
+)
+
+DUMMY_DATASET_WITH_NON_STANDARD_STRUCTURE = osp.join(
```
IRDonch:
```suggestion
DUMMY_DATASET_WITH_NON_STANDARD_STRUCTURE_DIR = osp.join(
```

datumaro/plugins/common_segmentation_format.py
```diff
@@ -0,0 +1,148 @@
...
+    def _load_items(self, path):
+        items = {}
+
+        image_dir = osp.join(path, CommonSegmentationPath.IMAGES_DIR)
+
+        if not osp.isdir(image_dir):
+            image_dir = glob.glob(
+                osp.join(path, "**", CommonSegmentationPath.IMAGES_DIR), recursive=True
+            )
+            if image_dir:
```
IRDonch: If this is false, then `image_dir` will remain a list, and the next call to `osp.isdir` will crash.

site/content/en/docs/formats/css.md
yasakova-anastasia: Added a link to AC documentation for this format.

datumaro/plugins/common_segmentation_format.py
yasakova-anastasia: Renamed to `common_semantic_segmentation`

datumaro/plugins/common_segmentation_format.py
yasakova-anastasia: Fixed

datumaro/plugins/common_segmentation_format.py
yasakova-anastasia: I agree. Fixed it.

datumaro/util/mask_tools.py
yasakova-anastasia: While testing this format on some dataset, I found this problem. The masks have pixels that could not be identified using the given classes. Such pixels should be defined as background.

datumaro/plugins/common_segmentation_format.py
yasakova-anastasia: Fixed

datumaro/plugins/common_segmentation_format.py
yasakova-anastasia: Removed this code

site/content/en/docs/formats/common_semantic_segmentation.md
```diff
@@ -0,0 +1,83 @@
...
+To add classes, you should use [`dataset_meta.json`](/docs/user-manual/supported_formats/#dataset-meta-file).
+
+To make sure that the selected dataset has been added to the project, you can
+run `datum project info`, which will display the project information.
+
+## Export to other formats
+
+Datumaro can convert a CSS dataset into any other format [Datumaro supports](/docs/user-manual/supported_formats/).
+To get the expected result, convert the dataset to formats
+that support the segmentation task (e.g. PASCAL VOC, CamVid, CItyscapes, etc.)
```
zhiltsov-max:
```suggestion
that support the segmentation task (e.g. PASCAL VOC, CamVid, Cityscapes, etc.)
```

tests/test_common_semantic_segmentation_format.py
```diff
@@ -0,0 +1,129 @@
...
+)
+from datumaro.util.test_utils import compare_datasets
+
+from .requirements import Requirements, mark_requirement
+
+DUMMY_DATASET_DIR = osp.join(
+    osp.dirname(__file__), "assets", "common_semantic_segmentation_dataset", "dataset"
+)
+
+DUMMY_DATASET_WITH_NON_STANDARD_STRUCTURE_DIR = osp.join(
```
zhiltsov-max: This dataset has custom filenames, but the structure is the same as in the "normal" case. I suggest to rename this dataset to make the difference more clear.

datumaro/util/mask_tools.py
zhiltsov-max: > Such pixels should be defined as background.

Where this comes from? Why not throw an error?

Also, we don't always have background at id 0 already, so maybe it should be a parameter.

yasakova-anastasia: I think that if a pixel does not belong to any presented class, it belongs to the background (it seemed to me that this class was partially needed for this). But I agree that this can be interpreted as an error in the dataset.

> Also, we don't always have background at id 0 already, so maybe it should be a parameter.

Do you mean add a parameter to the `load_mask` function for the background class number?

IRDonch: +1 to treating this as an error.

tests/test_common_semantic_segmentation_format.py
yasakova-anastasia: Renamed

datumaro/util/mask_tools.py
yasakova-anastasia: Added an error for this case.

zhiltsov-max: > I think that if a pixel does not belong to any presented class, it belongs to the background 

Why?

> it seemed to me that this class was partially needed for this. But I agree that this can be interpreted as an error in the dataset.

I agree, that it can be true in some cases, like id 255 / color 255,255,255 in VOC masks is used for "boundary" pixels, which are not supposed to be used for accuracy evaluation. But this fact cannot be treated as a common practice, or always supposed.

> Do you mean add a parameter to the load_mask function for the background class number?

No, I'm speaking about the `unpaint_mask` function, which is being discussed in this thread. We could add a parameter to map unknown pixels to background, if it is needed, but it should not be the default behavior.

datumaro/plugins/common_semantic_segmentation_format.py
```diff
@@ -0,0 +1,150 @@
...
+        parser = super().build_cmdline_parser(**kwargs)
+        parser.add_argument("--image-prefix", default="", help="Image prefix (default: '')")
+        parser.add_argument("--mask-prefix", default="", help="Mask prefix (default: '')")
+        return parser
+
+    @classmethod
+    def detect(cls, context: FormatDetectionContext) -> None:
+        context.require_file(f"**/{CommonSemanticSegmentationPath.IMAGES_DIR}/**/*.*")
+        context.require_file(f"**/{CommonSemanticSegmentationPath.MASKS_DIR}/**/*.*")
+        context.require_file(f"**/{DATASET_META_FILE}")
```
IRDonch: Similarly to the loading code, this doesn't need to search for images and masks everywhere in the tree. Once you have located the meta file, you can take its parent directory and search there.

datumaro/util/mask_tools.py
```diff
@@ -64,14 +65,22 @@ def unpaint_mask(painted_mask, inverse_colormap=None):
...
 
     painted_mask = painted_mask.astype(int)
     painted_mask = (
         painted_mask[:, :, 0] + (painted_mask[:, :, 1] << 8) + (painted_mask[:, :, 2] << 16)
     )
     uvals, unpainted_mask = np.unique(painted_mask, return_inverse=True)
-    palette = np.array([map_fn(v) for v in uvals], dtype=np.min_scalar_type(len(uvals)))
+    palette = []
+    for v in uvals:
+        color = map_fn(v)
```
IRDonch: `map_fn` doesn't return a color, it returns a class index.

datumaro/util/mask_tools.py
```diff
@@ -64,14 +65,22 @@ def unpaint_mask(painted_mask, inverse_colormap=None):
...
+        if color is None:
+            raise DatasetError(
```
IRDonch: I don't think `DatasetError` makes sense here, since there is no requirement that the mask must come from a dataset.

yasakova-anastasia: Replaced with `AnnotationError`

datumaro/plugins/common_semantic_segmentation_format.py
yasakova-anastasia: Fixed

datumaro/util/mask_tools.py
yasakova-anastasia: Fixed

datumaro/util/mask_tools.py
yasakova-anastasia: Replaced with `UndeclaredColorError`

IRDonch: @zhiltsov-max Please check that the new exception fits into your framework.

IRDonch (dismissed):

CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   (<https://github.com/openvinotoolkit/datumaro/pull/539>)
 - BraTS format (import-only) (.npy and .nii.gz), new `MultiframeImage`
   media type (<https://github.com/openvinotoolkit/datumaro/pull/628>)
+- Common Semantic Segmentation dataset format (import-only)
```
zhiltsov-max, line 18: Don't you think it's worth making not import-only? I feel, this format could be used extensively.

datumaro/plugins/common_semantic_segmentation_format.py
```diff
@@ -0,0 +1,152 @@
...
+
+        self._image_prefix = image_prefix
+        self._mask_prefix = mask_prefix
+
+        self._root_dir = path
+
+        if has_meta_file(path):
+            label_map = parse_meta_file(path)
+            self._categories = make_categories(label_map)
+        else:
```
zhiltsov-max: I can't see real difference between these branches. Can we use just the `else` branch? Also, where the meta file can be found, if not in the root?

datumaro/plugins/common_semantic_segmentation_format.py
```diff
@@ -0,0 +1,152 @@
...
+            mask = lazy_mask(mask_path, self._categories[AnnotationType.mask].inverse_colormap)
+            mask = mask()  # loading mask through cache
+
+            classes = np.unique(mask)
+            for label_id in classes:
+                annotations.append(
+                    Mask(image=self._lazy_extract_mask(mask, label_id), label=label_id)
+                )
+
+            items[item_id] = DatasetItem(id=item_id, media=image, annotations=annotations)
```
zhiltsov-max:
```suggestion
            items[item_id] = DatasetItem(id=item_id, subset=self._subset, media=image, annotations=annotations)
```

CHANGELOG.md
yasakova-anastasia: This format is now needed only for import for WB.

datumaro/util/mask_tools.py
```diff
@@ -64,14 +65,20 @@ def unpaint_mask(painted_mask, inverse_colormap=None):
     if callable(inverse_colormap):
         map_fn = lambda a: inverse_colormap((a >> 16) & 255, (a >> 8) & 255, a & 255)
     else:
-        map_fn = lambda a: inverse_colormap[((a >> 16) & 255, (a >> 8) & 255, a & 255)]
+        map_fn = lambda a: inverse_colormap.get(((a >> 16) & 255, (a >> 8) & 255, a & 255), None)
```
zhiltsov-max, line 70: Could you add docs and tests for this function?

site/content/en/docs/formats/common_semantic_segmentation.md
```diff
@@ -0,0 +1,83 @@
...
+## Import Common Semantic Segmentation dataset
+
+A Datumaro project with a CSS source can be created in the following way:
+
+``` bash
+datum create
+datum import --format common_semantic_segmentation <path/to/dataset>
+```
+
+Extra options for adding a source in the COCO format:
```
zhiltsov-max:
```suggestion
Extra import options:
```

COCO (panoptic) doesn't seem to be related to this format, because it stores masks in a different encoding. Also, they won't be located by this format, if the dataset follows the default Datumaro COCO layout.

site/content/en/docs/formats/common_semantic_segmentation.md
```diff
@@ -0,0 +1,83 @@
...
+    │   ├── <img1>.png
+    │   ├── <img2>.png
+    │   └── ...
+    └── masks/
+        ├── <img1>.png
+        ├── <img2>.png
+        └── ...
+```
+
+To add classes, you should use [`dataset_meta.json`](/docs/user-manual/supported_formats/#dataset-meta-file).
```
zhiltsov-max:
```suggestion
To describe classes and colors, you should use [`dataset_meta.json`](/docs/user-manual/supported_formats/#dataset-meta-file).
```

datumaro/util/mask_tools.py
zhiltsov-max: I like the error itself, but I'm not sure an `InvalidAnnotationError` (which is inherited from `DatasetImportError`) should be used in this context. I'd rather just throw a regular `KeyError` and handled in on the calling side.

CHANGELOG.md
zhiltsov-max: How WB is going to obtain such datasets? This format is crafted for OpenVINO and it is essential, that all our projects can work with it. Someone needs to be able to write in this format.

datumaro/plugins/common_semantic_segmentation_format.py
yasakova-anastasia: The meta file can be in a subfolder in some cases (located together with the images and masks folders)

CHANGELOG.md
yasakova-anastasia: I agree with this. We will need to support export in this format. But at the moment, the format is needed only for detection.
I think that the export task can wait.

datumaro/util/mask_tools.py
yasakova-anastasia: Added test for this function

zhiltsov-max (approved):

PR #686, opened 2022-03-21
Improve VOC parsing errors

zhiltsov-max:
### Summary

Depends on #684

- Improved VOC parsing errors

### Checklist

### License

```python
#
```


datumaro/plugins/voc_format/extractor.py
```diff
@@ -53,9 +65,10 @@ def __init__(self, path, task, **kwargs):
         )
         self._items = {item: None for item in self._load_subset_list(path)}
 
-    def _get_label_id(self, label):
+    def _get_label_id(self, label: str):
```
IRDonch: Probably worth adding a return type annotation too.

datumaro/plugins/voc_format/extractor.py
```diff
@@ -83,8 +96,9 @@ def _load_subset_list(self, subset_path):
                         if len(objects) == 3:
                             line = objects[1]
                         else:
-                            raise ValueError(
-                                "Line %s: unexpected number " "of quotes in filename" % line
+                            raise InvalidAnnotationError(
+                                f"{osp.basename(subset_path)}:{i}: "
```
IRDonch:
```suggestion
                                f"{osp.basename(subset_path)}:{i + 1}: "
```

Lines are usually numbered starting with 1.

datumaro/plugins/voc_format/extractor.py
```diff
@@ -163,116 +190,150 @@ def __iter__(self):
...
                 ann_file = osp.join(anno_dir, item_id + ".xml")
                 if osp.isfile(ann_file):
                     root_elem = ElementTree.parse(ann_file)
-                    height = root_elem.find("size/height")
-                    if height is not None:
-                        height = int(height.text)
-                    width = root_elem.find("size/width")
-                    if width is not None:
-                        width = int(width.text)
+                    if root_elem.getroot().tag != "annotation":
```
IRDonch: IMO, it would be better to move the `getroot` call to the initialization of `root_elem`, to make the name more justified. Right now `root_elem` is not even an element - it's the tree.

datumaro/plugins/voc_format/extractor.py
```diff
@@ -163,116 +190,150 @@ def __iter__(self):
...
-
-            truncated_elem = object_elem.find("truncated")
-            attributes["truncated"] = truncated_elem is not None and truncated_elem.text == "1"
-
-            occluded_elem = object_elem.find("occluded")
-            attributes["occluded"] = occluded_elem is not None and occluded_elem.text == "1"
+    @staticmethod
+    def _parse_field(
+        root, xpath: str, cls: Union[None, Type[T], Tuple[Type, ...]] = None, required: bool = True
+    ) -> Optional[Union[T, str]]:
```
IRDonch:
```suggestion
    @staticmethod
    def _parse_field(
        root: ElementTree.Element, xpath: str, cls: Type[T] = str, *, required: bool = True
    ) -> T:
```

Regarding the type of `cls`: the tuple alternative was not actually supported by the code; while the `None` alternative is unnecessary, as `str` gives the same result. You can prevent the construction of unnecessary strings by replacing `if cls` with `if cls is not str` in the function body.

datumaro/plugins/voc_format/extractor.py
```diff
@@ -163,116 +190,150 @@ def __iter__(self):
...
+
+                truncated_elem = object_elem.find("truncated")
+                if truncated_elem is not None and truncated_elem.text not in ["0", "1"]:
+                    raise InvalidFieldError("truncated")
+                attributes["truncated"] = truncated_elem is not None and truncated_elem.text == "1"
+
+                occluded_elem = object_elem.find("occluded")
+                if occluded_elem is not None and occluded_elem.text not in ["0", "1"]:
+                    raise InvalidFieldError("occluded")
+                attributes["occluded"] = occluded_elem is not None and occluded_elem.text == "1"
```
IRDonch: I think you could use a loop here.

datumaro/plugins/voc_format/extractor.py
```diff
@@ -163,116 +190,150 @@ def __iter__(self):
...
+
+                    if self._task is not VocTask.person_layout:
+                        break
+                    has_parts = True
+                    item_annotations.append(Bbox(*part_bbox, label=part_label_id, group=group))
+
+                attributes_elem = object_elem.find("attributes")
+                if attributes_elem is not None:
+                    for attr_elem in attributes_elem.iter("attribute"):
+                        attributes[attr_elem.find("name").text] = attr_elem.find("value").text
```
IRDonch: Shouldn't `_parse_field` be used here?

datumaro/plugins/voc_format/extractor.py
```diff
@@ -344,31 +405,36 @@ def _load_annotations(self, item_id):
...
+                    self._ctx.error_policy.report_annotation_error(
+                        UndeclaredLabelError(str(label_id)), item_id=(item_id, self._subset)
                     )
 
                 image = compiled_mask.lazy_extract(instance_id)
 
                 item_annotations.append(Mask(image=image, label=label_id, group=instance_id))
         elif class_mask is not None:
-            log.warning("Item %s: only class segmentations available" % item_id)
+            log.warning(f"Item {item_id}: only class segmentations available")
```
IRDonch: I think with logging it's better to do this:

```suggestion
            log.warning("Item %s: only class segmentations available", item_id)
```

since that prevents the message from being formatted unless the appropriate log level is enabled.

datumaro/plugins/voc_format/format.py
```diff
@@ -123,42 +125,55 @@ def make_voc_label_map():
...
+def parse_label_map(path: str) -> Dict[str, Tuple[RgbColor, List[str], List[str]]]:
+    """
+    Parses a label map file in the format:
+    'name : color (r, g, b) : parts (hand, feet, ...) : actions (siting, standing, ...)'
+
+    Parameters:
+        path: File path
+
+    Returns:
+        A dictionary: label -> (color, parts, actions)
```
IRDonch, line 151: This is wrong, because the actual return dictionary contains lists, not tuples (the function should probably be fixed, not the description).

IRDonch (changes requested):

datumaro/plugins/voc_format/extractor.py
zhiltsov-max: Yeah, it will work. Moved

datumaro/plugins/voc_format/extractor.py
zhiltsov-max: Applied here and in other cases

datumaro/plugins/voc_format/extractor.py
zhiltsov-max: Added

datumaro/plugins/voc_format/extractor.py
zhiltsov-max: Changed

datumaro/plugins/voc_format/extractor.py
zhiltsov-max: Added

datumaro/plugins/voc_format/extractor.py
zhiltsov-max: Yes, thanks.

datumaro/plugins/voc_format/extractor.py
zhiltsov-max: Applied

datumaro/plugins/voc_format/format.py
zhiltsov-max: The code often relies on element modification, moving to tuples would be inefficient for no reason and immutability from tuples is not supposed. It requires refactoring, but it is not in the scope of this PR. Added a comment on this and some type annotations.

datumaro/plugins/voc_format/extractor.py
```diff
@@ -42,20 +61,21 @@ def __init__(self, path, task, **kwargs):
...
                     (i, l, label_color(i))
                     for i, l in enumerate(self._categories[AnnotationType.label].items)
                 )
-            )
+            ),
         )
         self._items = {item: None for item in self._load_subset_list(path)}
 
-    def _get_label_id(self, label):
+    def _get_label_id(self, label: str) -> Optional[int]:
```
IRDonch: `Optional` seems wrong, because this function can't return `None`.

zhiltsov-max: Fixed

datumaro/plugins/voc_format/extractor.py
```diff
@@ -163,116 +197,141 @@ def __iter__(self):
...
+        if elem is None:
+            if required:
+                raise MissingFieldError(xpath)
+            else:
+                return None
 
-            pose_elem = object_elem.find("pose")
-            if pose_elem is not None:
-                attributes["pose"] = pose_elem.text
+        if not issubclass(cls, str):
```
IRDonch: This condition is incorrect. If `cls` is a subclass of `str` (but not `str` itself), then you _do_ need to cast `elem.text` to `cls`, otherwise it'll be of the wrong type.

So the condition needs to be `if cls is not str`. Actually, for readability it might be better to rewrite it as:

```python
if cls is str:
    return elem.text

try:
    return cls(elem.text)
except Exception as e:
    raise InvalidFieldError(xpath) from e
```

zhiltsov-max: Ok, makes sense

tests/test_voc_format.py
```diff
@@ -496,6 +537,228 @@ def test_can_pickle(self):
                 compare_datasets_strict(self, source, parsed)
 
 
+class VocExtractorTest(TestCase):
+    # ?xml... must be in the file beginning
+    XML_ANNOTATION_TEMPLATE = """<?xml version="1.0" encoding="UTF-8"?>
```
IRDonch:
```suggestion
    XML_ANNOTATION_TEMPLATE = """\
<?xml version="1.0" encoding="UTF-8"?>
```

Just for the sake of readability. You could also use `textwrap.dedent` to avoid violating natural indentation for the contents of the string.

zhiltsov-max: Agree with \\. 
Indentation is intended for simplified editing of this fragment.

tests/test_voc_format.py
```diff
@@ -496,6 +537,228 @@ def test_can_pickle(self):
...
+    """
+
+    @classmethod
+    def _write_xml_dataset(cls, root_dir, fmt_dir="Main", mangle_list_file=None, mangle_xml=None):
+        subset_file = osp.join(root_dir, "ImageSets", fmt_dir, "test.txt")
+        os.makedirs(osp.dirname(subset_file))
+        with open(subset_file, "w") as f:
+            text = "a\n" if fmt_dir != "Layout" else "a 0\n"
+            if mangle_list_file:
+                text = mangle_list_file(text)
```
IRDonch: The `mangle_list_file` argument is never used.

tests/test_voc_format.py
```diff
@@ -496,6 +537,228 @@ def test_can_pickle(self):
...
+            f.write(text)
+
+        ann_file = osp.join(root_dir, "Annotations", "a.xml")
+        os.makedirs(osp.dirname(ann_file))
+        with open(ann_file, "w") as f:
+            text = cls.XML_ANNOTATION_TEMPLATE
+            if mangle_xml:
+                text = mangle_xml(text)
```
IRDonch: IMO, mangling the XML content as _text_ leads to less clear and less robust tests. For example, there's a test below that matches for `<name>cat</name>`, but there's no guarantee that the match will happen inside the object. So if the template is later extended to contain multiple instances of `<name>cat</name>`, then the test might end up doing something unexpected.

It would be better to parse the XML string here, give the `mangle_xml` function the root element, then dump it back to XML. Then the mangle function could look like this:

```python
def mangle_xml(root):
    root.find('object/name').text = 'test'
```

zhiltsov-max: Fixed

tests/test_voc_format.py
zhiltsov-max: Fixed

IRDonch (approved):

PR #687, opened 2022-03-22
Improve YOLO parsing errors

zhiltsov-max:
### Summary

Depends on #686 

- Added more informative YOLO parsing errors
- Added more type annotations in extractor
- Allowed leading spaces in the `.data` file
- Added type docs for `util.image.ImageMeta`

### Checklist

### License

```python
#
```


datumaro/plugins/yolo_format/extractor.py
```diff
@@ -2,45 +2,61 @@
...
             self._parent = parent
-            self.items = OrderedDict()
+            self.items: Dict[str, Union[str, DatasetItem]] = OrderedDict()
 
         def __iter__(self):
             for item_id in self.items:
-                yield self._parent._get(item_id, self._name)
+                item = self._parent._get(item_id, self._name)
+                if item is not None:
+                    yield item
```
IRDonch, line 42: This enables questionable semantics: if any items are skipped in this way, then the number of items yielded will be different from the value returned by `len(subset)`.

tests/test_yolo_format.py
```diff
@@ -284,7 +271,7 @@ def test_can_save_and_load_without_path_prefix(self):
                     subset="valid",
                     media=Image(data=np.ones((8, 8, 3))),
                     annotations=[
-                        Bbox(0, 1, 5, 2, label=2),
+                        Bbox(0, 1, 5, 2, label=1),
```
IRDonch, line 274: What's the reason for this change?

zhiltsov-max: The annotation is invalid, there are only 2 labels in the dataset. Now it is checked, and the error is raised.

datumaro/plugins/yolo_format/extractor.py
zhiltsov-max: I agree about this, but it is not clear how to resolve the problem, and it has very limited ways to impact the situation, because it is only used until the imported dataset is cached. During caching/iteration, the actual count is computed. Actually, in VOC we have the same issue. Maybe, we could even consider it normal, because unless we loaded the data, it is the best we can report.

IRDonch: IMO, the proper solution would involve removing `__len__` from `IExtractor`.

IRDonch (approved):

PR #688, opened 2022-03-22
Allow custom subset names in YOLO

### LINKED ISSUE #565: export to yolo format with train and valid data
Hello Everyone, I noticed something. When I run the following command within a project:

```
datum transform -t split -- -t detection --subset train:.8 --subset val:.2 
```

then the two image folders train and val are created. The format of the data set is a coco_instances format.

If I now export the whole project as YOLO format, then the val data is not exported, because both the folder does not contain name valid and the annotation file. 

The console displays the following information:
```
WARNING: Skipping subset export 'val'. If specified, the only valid names are 'train', ‚valid'
```

I have currently solved it by renaming the folder and anntotations file from val to valid. After that it works.

Should this be so?

zhiltsov-max:
### Summary

Depends on #687 
Closes #565

- Allowed custom subset names in YOLO import and export, except `classes`, `names` and `backup`
- Exporting a subset with a reserved name now leads to an export error

### Checklist

### License

```python
#
```


datumaro/plugins/yolo_format/extractor.py
```diff
@@ -71,29 +71,24 @@ def __init__(
...
+        # The original format is like this:
+        #
+        # classes = 2
+        # train  = data/train.txt
+        # valid  = data/test.txt
+        # names = data/obj.names
+        # backup = backup/
+        #
+        # To support more subset names, we disallow subsets
+        # called 'classes' and 'backup'.
```
yasakova-anastasia: Need to add 'names'?

yasakova-anastasia (dismissed):

datumaro/plugins/yolo_format/extractor.py
zhiltsov-max: Added

PR #699, opened 2022-03-28
Fix the GitHub pages workflow

IRDonch:
### Summary
This workflow uses Ubuntu 20.04, which has pip 20.0.2 installed by default. However, the latest version of the `openvino` wheel uses the `manylinux_2_27_x86_64` platform tag, which was not supported until pip 20.3. pip 20.0.2 also doesn't have the new resolver enabled by default, so it's not smart enough to try an older version of `openvino-dev`. Therefore, this pip cannot install `openvino` and therefore `openvino-dev`.

This can be solved by upgrading pip, and there _is_ a pip upgrade command in the workflow. However, it has no effect, because the upgrade installs the new version as `/usr/local/bin/pip`, while bash remembers that pip was at `/usr/bin/pip` and keeps using that version.

Fix it by separating the pip upgrade into a separate step. It looks nicer this way in the build log, anyway.

### Checklist

### License

```python
#
```

### DIFF
--- a/.github/workflows/github_pages.yml
+++ b/.github/workflows/github_pages.yml
@@ -30,9 +30,15 @@ jobs:
         run: |
           npm ci
 
-      - name: Build docs
+      # The pip upgrade must be in a separate step, because otherwise bash will
+      # remember where the system-installed pip was, and will use it in any following
+      # commands instead of the newly-installed pip.
+      - name: Upgrade pip
         run: |
           pip install --upgrade pip
+
+      - name: Build docs
+        run: |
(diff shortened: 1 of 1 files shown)


zhiltsov-max (approved):

zhiltsov-max:
It looks like it partially helped, but now there is another issue https://github.com/openvinotoolkit/datumaro/runs/5735344738?check_suite_focus=true

IRDonch:
Yeah, looks like click published a release with a breaking change less than an hour after I submitted this fix. 😄

PR #700, opened 2022-03-28
CSR dataset format (import-only)

yasakova-anastasia:
### Summary

- Added CSR dataset format (import-only)
- Added `SuperResolutionAnnotation` annotation type

### Checklist

### License

```python
#
```


site/content/en/docs/user-manual/supported_formats.md
```diff
@@ -118,6 +118,9 @@ List of supported formats:
   - [Format specification](http://host.robots.ox.ac.uk/pascal/VOC/voc2012/htmldoc/index.html)
   - [Dataset example](https://github.com/openvinotoolkit/datumaro/tree/develop/tests/assets/voc_dataset)
   - [Format documentation](/docs/formats/pascal_voc)
+- Super Resolution
```
zhiltsov-max: Shouldn't it be prefixed Common as well as Semantic Segmentation?

site/content/en/docs/formats/super_resolution.md
```diff
@@ -0,0 +1,50 @@
...
+weight: 2
+---
+
+## Format specification
+
+Super resolution format specification is available [here](https://github.com/openvinotoolkit/open_model_zoo/blob/master/tools/accuracy_checker/openvino/tools/accuracy_checker/annotation_converters/README.md#supported-converters).
+
+Supported attributes:
+- `HR` (`Image`): high resolution image
+- `upsampled` (`Image`): upsampled image
```
zhiltsov-max: From the description I see that `upsampled` is optional, but `HR` is not. In the implementation, `HR` both are optional in extractor, but in detection `HR` is mandatory. Also, `HR` is in lower case in the implementation.

From the format semantics viewpoint, it would be better to represent these attributes as annotations. Attributes can work too, but I think it will be changed later, because they do not seem to be compatible with other formats.

site/content/en/docs/user-manual/supported_formats.md
yasakova-anastasia: I'm not sure about this. For WB it is, but in AC it is `super_resolution_converter`.

site/content/en/docs/formats/super_resolution.md
yasakova-anastasia: Do you suggest adding a new annotation type for this?

zhiltsov-max: Ideally, I'd prefer this solution.

site/content/en/docs/user-manual/supported_formats.md
zhiltsov-max: As I see, in AC there are 2 kinds of this format, and this one looks more like the directory-based variant. 

Being a user, I'd probably be confused about the current name, because it is too generic.

site/content/en/docs/formats/super_resolution.md
yasakova-anastasia: Added `ImageResolution` annotation type

yasakova-anastasia: Added `ImageResolution` annotation type

datumaro/components/annotation.py
```diff
@@ -7,6 +7,7 @@
 from enum import Enum, auto
 from functools import partial
 from itertools import zip_longest
+from tkinter import Image
```
zhiltsov-max: The import is invalid

datumaro/components/annotation.py
```diff
@@ -787,3 +789,14 @@ class Caption(Annotation):
...
+
+
+@attrs(slots=True, order=False)
+class ImageResolution(Annotation):
+    """
+    Represents high and low resolution images.
+    """
+
+    _type = AnnotationType.image_resolution
+    hr_image = field()
```
zhiltsov-max:
```suggestion
    hr_image: Image = field()
```

datumaro/components/annotation.py
```diff
@@ -787,3 +789,14 @@ class Caption(Annotation):
 
     _type = AnnotationType.caption
     caption: str = field(converter=str)
+
+
+@attrs(slots=True, order=False)
+class ImageResolution(Annotation):
+    """
+    Represents high and low resolution images.
```
zhiltsov-max: This annotation doesn't contain low resolution images.

datumaro/components/annotation.py
```diff
@@ -787,3 +789,14 @@ class Caption(Annotation):
...
+    """
+
+    _type = AnnotationType.image_resolution
+    hr_image = field()
+    upsampled_image: Optional[Image] = field(default=None)
```
zhiltsov-max: As I understand, in this task it is not an annotation. It's literally just the resized input image. I don't think storing it in annotations is the right way.

datumaro/components/annotation.py
```diff
@@ -787,3 +789,14 @@ class Caption(Annotation):
 
     _type = AnnotationType.caption
     caption: str = field(converter=str)
+
+
+@attrs(slots=True, order=False)
+class ImageResolution(Annotation):
```
zhiltsov-max: I propose to rename it to one of these:

`class SuperResolutionAnnotation(Annotation):`
`class ImageAnnotation(Annotation):`

`ImageResolution` doesn't sound like an annotation type. I'd also expect 2 numbers in that case instead of images.

datumaro/components/annotation.py
yasakova-anastasia: Fixed

datumaro/components/annotation.py
yasakova-anastasia: Fixed

datumaro/components/annotation.py
yasakova-anastasia: Yes, you're right. Moved it to attributes.

datumaro/components/annotation.py
yasakova-anastasia: Renamed to `SuperResolutionAnnotation`

datumaro/components/annotation.py
yasakova-anastasia: Fixed

zhiltsov-max (approved):

PR #726, opened 2022-09-05
Release 0.3.1

chuneuny-emily:
### Summary
This PR fix

- [CVS-84321](https://jira.devtools.intel.com/browse/CVS-84321) Export Annotated Dataset - lack of image's depth value for VOC dataset
- [CVS-86848](https://jira.devtools.intel.com/browse/CVS-86848) Annotations inside another annotation are duplicated during import for VOC format
- [CVS-85120](https://jira.devtools.intel.com/browse/CVS-85120) Cannot export detection->classification dataset

### How to test

- I performed the existing unit tests locally, and there was no regression on my changes.
- I performed e2e tests manually, and the issues above were not reproducible anymore.

### Checklist

- [x] I have added description of my changes into [CHANGELOG](https://github.com/chuneuny-emily/datumaro/blob/emily/release-0.3.1/CHANGELOG.md)

### License

```python
#
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,7 +83,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   (<https://jira.devtools.intel.com/browse/CVS-84321>)
 - Adding to handle the numerical labels in task chains properly
   (<https://jira.devtools.intel.com/browse/CVS-85120>)
-- Fixing the issue that annotations inside another annotation (polygon) 
+- Fixing the issue that annotations inside another annotation (polygon)
   are duplicated during import for VOC format
   (<https://jira.devtools.intel.com/browse/CVS-86848>)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,30 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Security
 - TBD
 
+## 05/09/2022 - Release v0.3.1 (hotfix)
+### Added
+- TBD
+
+### Changed
+- TBD
+
+### Deprecated
+- TBD
+
+### Removed
+- TBD
+
+### Fixed
+- Adding depth value of image when dataset is exported in VOC format
+  (<https://jira.devtools.intel.com/browse/CVS-84321>)
+- Adding to handle the numerical labels in task chains properly
+  (<https://jira.devtools.intel.com/browse/CVS-85120>)
+- Fixing the issue that annotations inside another annotation (polygon) are duplicated during import for VOC format
+  (<https://jira.devtools.intel.com/browse/CVS-86848>)
+
+### Security
+- TBD
+
 ## 21/02/2022 - Release v0.3
 ### Added
 - Ability to import a video as frames with the `video_frames` format and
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -65,6 +65,31 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  (<https://jira.devtools.intel.com/browse/CVS-84321>)
+- Adding to handle the numerical labels in task chains properly
+  (<https://jira.devtools.intel.com/browse/CVS-85120>)
+- Fixing the issue that annotations inside another annotation (polygon)
+  are duplicated during import for VOC format
+  (<https://jira.devtools.intel.com/browse/CVS-86848>)
+
+### Security
+- TBD
+
```
goodsong81: Why don't you remove sections without changes?
And please replace Jira ticket links by PR links.
(CVS tickets are not accessible from external)

datumaro/components/dataset_filter.py
```diff
@@ -64,6 +64,8 @@ def encode_annotation_base(cls, annotation):
         ET.SubElement(ann_elem, "type").text = str(annotation.type.name)
 
         for k, v in annotation.attributes.items():
+            if k.isdigit():
+                k = "_" + k
```
goodsong81, line 68: So, you decided to add _ prefix for export? Do you extract _ in case of import?

CHANGELOG.md
chuneuny-emily: I kept the way they were doing at Datumaro. Removing is better?
Should I need to write that issues in issue tab in Datumaro and link it? or just removing the CSV links?

datumaro/components/dataset_filter.py
chuneuny-emily: Yes. As you suggested, I'm using that way with _ prefix. The sub labels in task chains are classified as action in Datumaro. And these _ prefix has been using in action elements with digit. But these action elements are filtered during import. This is the code that is being filtered during import. I checked both export and import, and compared the results between digit case and non-digit case and I checked they had the same results.

goodsong81 (approved):
LGTM.

datumaro/plugins/voc_format/converter.py
```diff
@@ -283,7 +283,10 @@ def _export_annotations(self, item: DatasetItem, *, image_filename: str, lists:
                 size_elem = ET.SubElement(root_elem, "size")
                 ET.SubElement(size_elem, "width").text = str(w)
                 ET.SubElement(size_elem, "height").text = str(h)
-                ET.SubElement(size_elem, "depth").text = ""
+                depth = ""
+                if item.media.data is not None:
+                    depth = str(item.media.data.shape[-1])
```
zhiltsov-max, line 288: In many cases VOC export is going to take much more time now, because the images have to be loaded. For width and height we often have metainfo in other formats.

chuneuny-emily: The images with coco or other formats are not loaded, so item.media.data is None.
It is run when there are the loaded images. If not, it keeps "".

Any suggestion to set depth?

zhiltsov-max: `item.media.data` is None only when the image file is not available, but we have metainfo. If you wanted to avoid extra loading, you need to use [has_data](https://github.com/openvinotoolkit/datumaro/blob/develop/datumaro/components/media.py#L113)

chuneuny-emily: I see. Then I will update as using has_data you suggested soon. Do you think it should be included in v0.3.1?

zhiltsov-max: I can't answer on this question, release logic is out of my responsibility now.

PR #742, opened 2022-10-17
Hierarchical Labeling

bonhunko:
### Summary

Resolves 93482

- Add LabelGroup option to datumaro for representing 'exclusive'

### How to test

1. Generate a dataset containing hierarchical labels.
2. Export the dataset in 'datumaro' format.
3. The labels which disobey the single-selection rule will be filtered-out in the exported dataset

### Checklist

### License

```python
#
```


tests/test_labeling.py
```diff
@@ -0,0 +1,154 @@
...
+import numpy as np
+
+from datumaro.components.annotation import AnnotationType, Label, LabelCategories
+from datumaro.components.extractor import DatasetItem
+from datumaro.components.media import Image
+from datumaro.components.project import Dataset
+
+from .requirements import Requirements, mark_requirement
+
+TMP_DATASET_DIR = osp.join(osp.dirname(__file__), "tmp", "test_labeling")
```
vinnamkim: We can use `tempfile.TemporaryDirectory` for temporary usage (automatically reserve and clean up resources at code level). Please refer to https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory.

tests/test_labeling.py
```diff
@@ -0,0 +1,154 @@
...
+                            1,
+                            id=1,
+                        ),
+                    ],
+                ),
+            ],
+            categories={
+                AnnotationType.label: label_categories,
+            },
+        )
```
vinnamkim: You can initialize shared resources across tests in `setUpClass`. Please refer to https://docs.python.org/3/library/unittest.html#class-and-module-fixtures.

datumaro/plugins/datumaro_format/converter.py
```diff
@@ -383,3 +386,41 @@ def patch(cls, dataset, patch, save_dir, **kwargs):
...
+
+        name2parent_ss = {}
+        for label_category in categories[AnnotationType.label]:
+            name2parent_ss[label_category.name] = (
+                label_category.parent,
+                label_category.single_selection,
+            )
+
+        # collect childrens that have the same parent which only allow single-selection
+        parent2ss_indices = {}
```
vinnamkim: You can use `defaultdict(list)` in this case.

datumaro/plugins/datumaro_format/converter.py
```diff
@@ -383,3 +386,41 @@ def patch(cls, dataset, patch, save_dir, **kwargs):
...
+            label_name = categories[AnnotationType.label][annotation.label].name
+
+            parent, _ = name2parent_ss[label_name]
+            if len(parent) == 0:
+                continue
+
+            _, single_selection = name2parent_ss[parent]
+
+            if single_selection:
+                parent2ss_indices[parent] = parent2ss_indices.get(parent, []) + [i]
```
vinnamkim: If you use `parent2ss_indices = defaultdict(list)`, then this line will be `parent2ss_indices[parent] += [i]`.

datumaro/plugins/datumaro_format/converter.py
```diff
@@ -383,3 +386,41 @@ def patch(cls, dataset, patch, save_dir, **kwargs):
...
+    def _filterout_for_single_selection(item, categories):
+        if AnnotationType.label not in categories:
+            return item
+
+        name2parent_ss = {}
+        for label_category in categories[AnnotationType.label]:
+            name2parent_ss[label_category.name] = (
+                label_category.parent,
+                label_category.single_selection,
+            )
```
vinnamkim: Do we need to initialize this lookup dictionary for every item again and again? We can escalate this to the upper level and pass `name2parent_ss` to this function.

tests/test_labeling.py
bonhunko: Every test case is using slightly different LabelCategories or DatasetItems. Thus I think that we don't need to use shared resources in this time.

tests/test_labeling.py
bonhunko: cool!

datumaro/plugins/datumaro_format/converter.py
bonhunko: Indeed, I change that making name2parent_ss once per each export.

datumaro/plugins/datumaro_format/converter.py
bonhunko: cool!

vinnamkim (dismissed):
LGTM.
But I don't still get the concept of this work. Our current data structure of `categories` has a `parent` entity. It means that it can represent any tree structure with arbitrary depth and width.

How about this case?
```
        l1(T)
      /        \
 l2(F)        l3(F)
  /  \         /    \
l4    l5     l6     l7
```
and `Item1.annotations := [l1, l2, l3, l4, l5, l6, l7]` (if `label` has `single_selection=True`, then it denotes `label(T)`, otherwise, `label(F)`). Then, this filtering will make `Item1.annotations = [l1, l2, l4, l5, l6, l7]`. 

I feel this is a bit inconsistent because `l6` and `l7` is still alive although their parent `l3` is filtered out.

wonjuleee (dismissed):
LGTM

wonjuleee (commented):
Could you check whether Vinnam's concern is true or not?

bonhunko:
> Good, questions. This is a kind of architecture-level question. But I think that datumaro is needed to exports a general output. The detailed artchitecture-level filtering should be done by customer side. Because the detailed policy about how to handle hierarchical labeling should be determined by the customer. So please think that this implement only filter-out the error-cases that obviously wrong.

After facing some practical cases, we may change this filtering policy again.

bonhunko:
As vinnam suggested and discussed with wonju, I'll follow the LabelGroup implementation to mark the 'exclusive' after that I'll request review again.

bonhunko:
As we discussed, I import the LabelGroup format from the OTX, please review again.

tests/assets/datumaro_dataset/annotations/test.json
```diff
@@ -0,0 +1,77 @@
...
+                    10,
+                    5
+                ]
+            },
+            "media": {
+                "path": "../tests/assets/datumaro_dataset/images/test/c.jpg"
+            }
+        }
+    ]
+}
```
vinnamkim: Can we add new line at the end of file? This is because https://stackoverflow.com/a/729795/16880031.

tests/test_labeling.py
```diff
@@ -0,0 +1,85 @@
...
+
+        for item in dataset:
+            self.assertEqual(len(item.annotations), 2)
```
vinnamkim: Is this test duplicated with other existing tests?

tests/assets/datumaro_dataset/annotations/test.json
bonhunko: cool

tests/test_labeling.py
bonhunko: This is testing 'multi-class' supporting which is mentioned by emily that a single item may have multiple labels.
After your comments, I look around all previous test cases and the test bellow is already testing the 'multi-class' case, so I remove mine.
https://github.com/openvinotoolkit/datumaro/blob/develop/tests/test_coco_format.py#L272

vinnamkim (dismissed):
LGTM.

wonjuleee (approved):

PR #757, opened 2022-11-07
Export as CVAT video format

bonhunko:
### Summary

 - Support for exporting as CVAT video format
 - Add 'use_track' option to export DatasetItem into <track> tag
 - Implement CVS-93496
 - e.g.)
\<track id="0" label="label2"\>
    \<box frame="4" occluded="1" outside="0" keyframe="1" xtl="3.00" ytl="4.00" xbr="10.00" ybr="5.00" z_order="0"\>
    \</box\>
.....
    \<box frame="14" occluded="1" outside="0" keyframe="1" xtl="3.00" ytl="4.00" xbr="10.00" ybr="5.00" z_order="0"\>
    \</box\>
  \</track\>

### Checklist

### License

```python
#
```


datumaro/plugins/cvat_format/converter.py
```diff
@@ -449,7 +552,9 @@ def apply(self):
         for subset_name, subset in self._extractor.subsets().items():
             ann_path = osp.join(self._save_dir, "%s.xml" % subset_name)
             with open(ann_path, "w", encoding="utf-8") as f:
-                writer = _SubsetWriter(f, subset_name, subset, self)
+                writer = _SubsetWriter(
+                    f, subset_name, subset, self, self._use_track, self._original_size
```
vinnamkim: If we already have an argument `context=self`, then I think it would be not necessary to add new arguments for `self._use_track` and `self._original_size`.

tests/test_cvat_format.py
```diff
@@ -595,3 +597,76 @@ def test_inplace_save_writes_only_updated_data(self):
...
+            },
+        )
+
+        with TestDir() as path:
+            dataset.export(
+                format="cvat",
+                save_dir=path,
+                save_media=True,
+                use_track=True,
+                original_size=(video_hw[1], video_hw[0]),
```
vinnamkim: Could we infer this from `media` of `dataset`?

tests/test_cvat_format.py
```diff
@@ -595,3 +597,76 @@ def test_inplace_save_writes_only_updated_data(self):
...
+            items,
+            categories={
+                AnnotationType.label: LabelCategories.from_iterable(
+                    [
+                        ["label1"],
+                        ["label2"],
+                    ]
+                )
+            },
+        )
```
vinnamkim: I think this dataset does not include video `media` unlike `dataset_video`. Do you have any reason to make a new one rather than using `dataset_video`? I think that it is more natural to test `dataset_video` exportation.

datumaro/plugins/cvat_format/converter.py
```diff
@@ -276,17 +351,33 @@ def _get_label_attrs(self, label):
...
+        else:
+            shape_data = OrderedDict()
+
+        if write_frame:
+            shape_data.update(OrderedDict([("frame", str(int(shape.attributes.get("frame", 0))))]))
+        shape_data.update(
+            OrderedDict([("occluded", str(int(shape.attributes.get("occluded", False))))])
+        )
+        shape_data.update(
+            OrderedDict([("outside", str(int(shape.attributes.get("outside", False))))])
```
sooahleex, line 375: I did not catch CVAT property in detail. What does mean `"outside"` and `"keyframe"` in `shape_data`?

datumaro/plugins/cvat_format/converter.py
```diff
@@ -276,17 +351,33 @@ def _get_label_attrs(self, label):
...
+        )
+        shape_data.update(
+            OrderedDict([("keyframe", str(int(shape.attributes.get("keyframe", 0))))])
```
sooahleex: In here, `keyframe` could be `int` but in `test_video_export`, `keyframe` is set as `True`. Is both possible? I'm asking because I think it should be a number like `int`.

tests/test_cvat_format.py
bonhunko: "Dataset.import_from (.., "video_frames", ...)" sets categories to {}.
To my best knowledge, there is no option to add categories in the initialization.
I tried 'define_categories' https://github.com/openvinotoolkit/datumaro/blob/develop/datumaro/components/dataset.py#L857
but it throws CategoriesRedefinedError() https://github.com/openvinotoolkit/datumaro/blob/725110a6f1d0e37d1e2d92b194219e8a65a63c28/datumaro/components/dataset.py#L586

datumaro/plugins/cvat_format/converter.py
bonhunko: These are the attributions of an annotation.
'outside' = True indicates that the object is on the outside (not in the home)
'keyframe' = True means that this annotation is the keyframe of a track (I don't know the details but CVAT uses this feature internally when trying automated labeling for a track)
we set 'shape_data' = True when a label is a shape (polygon, box, ...), not a tag ('dog', 'human', ...)

datumaro/plugins/cvat_format/converter.py
bonhunko: Good point! I'll change the default value to False.

tests/test_cvat_format.py
vinnamkim: I guess that you can use this transformation

https://openvinotoolkit.github.io/datumaro/docs/user-manual/command-reference/transform/#project_labels
https://github.com/openvinotoolkit/datumaro/blob/d014342dfa021003b4be117598e9b0ba53e43177/datumaro/plugins/transforms.py#L717

bonhunko: Great! I changed to using ProjectLabels().
Please have a look, I couldn't sure that my changes are the best practice since I'm not familiar with 'transform'.

bonhunko: As vinnam suggested, I changed to use "dataset.transform(ProjectLabels, dst_labels=["label1", "label2"])"

vinnamkim (dismissed):
LGTM.

vinnamkim (dismissed):
LGTM.

datumaro/plugins/cvat_format/converter.py
```diff
@@ -432,12 +522,34 @@ def build_cmdline_parser(cls, **kwargs):
...
         self._reindex = reindex
         self._builtin_attrs = CvatPath.BUILTIN_ATTRS
         self._allow_undeclared_attrs = allow_undeclared_attrs
+        self._use_track = use_track
+        if use_track:
+            self._original_size = self.get_original_size(extractor)
+        else:
+            self._original_size = None
+
+    def get_original_size(self, extractor):
```
wonjuleee: Is there any plan to use this function in external? can we define this by private function?

bonhunko: Cool!

vinnamkim (dismissed):

bonhunko:
I resolve a conflict of CHANGELOG via [8ea3be2](https://github.com/openvinotoolkit/datumaro/pull/757/commits/8ea3be24ecdc2e3a62108dce24e76d25cf2a07e8) after the review approval. @vinnamkim, would you please give an approval again?

vinnamkim (approved):

wonjuleee (approved):
LGTM!

PR #772, opened 2022-11-30
COCO enable bbox with segement annotation

bonhunko:
### Summary

Resolves: CVS-98131

Currently, Datumaro removes the bbox annotations when segmentation annotation exits.
It confused the users and makes a dataset not fit for a detection task.

We thus keep the bbox annotations when importing a COCO dataset.
Then export the bbox annotation and relevant segmentation annotation to single annotation when exporting. 

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/plugins/data_formats/coco/base.py
+++ b/datumaro/plugins/data_formats/coco/base.py
@@ -412,17 +412,27 @@ class _CocoBase(SubsetBase):
                             rle=rle, label=label_id, id=ann_id, attributes=attributes, group=group
                         )
                     )
-            else:
-                bbox = self._parse_field(ann, "bbox", list)
+
+            bbox = self._parse_field(ann, "bbox", list)
+            if bbox and len(bbox) > 0:
                 if len(bbox) != 4:
                     raise InvalidAnnotationError(
                         f"Bbox has wrong value count {len(bbox)}. Expected 4 values."
                     )
-
                 x, y, w, h = bbox
                 parsed_annotations.append(
-                    Bbox(x, y, w, h, label=label_id, id=ann_id, attributes=attributes, group=group)
+                    Bbox(
+                        x,
+                        y,
+                        w,
+                        h,
+                        label=label_id,
+                        id=ann_id,
+                        attributes=attributes,
+                        group=group,
+                    )
                 )
+
         elif self._task is CocoTask.labels:
             label_id = self._get_label_id(ann)
             parsed_annotations.append(
--- a/datumaro/plugins/data_formats/coco/exporter.py
+++ b/datumaro/plugins/data_formats/coco/exporter.py
@@ -208,7 +208,10 @@ class _InstancesExporter(_TaskExporter):
 
         anns = boxes + polygons + masks
         leader = anno_tools.find_group_leader(anns)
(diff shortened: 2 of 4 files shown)


datumaro/plugins/data_formats/coco/exporter.py
```diff
@@ -208,7 +208,10 @@ def find_instance_parts(self, group, img_width, img_height):
 
         anns = boxes + polygons + masks
         leader = anno_tools.find_group_leader(anns)
-        bbox = anno_tools.max_bbox(anns)
+        if len(boxes) > 0:
+            bbox = anno_tools.max_bbox(boxes)
+        else:
+            bbox = anno_tools.max_bbox(anns)
```
wonjuleee, line 214: could you explain why this should be?

datumaro/plugins/data_formats/coco/base.py
```diff
@@ -412,17 +412,28 @@ def _load_annotations(self, ann, image_info=None, parsed_annotations=None):
...
+                            x,
+                            y,
+                            w,
+                            h,
+                            label=label_id,
+                            id=ann_id,
+                            attributes=attributes,
+                            group=group,
+                        )
+                    )
```
vinnamkim: We can omit `else` in this case instead of making a deeper tab level.

bonhunko: cool!

datumaro/plugins/data_formats/coco/exporter.py
bonhunko: As you know, An item may have multiple annotations for both COCO and Datumaro.
In the case of COCO, segmentation and bbox are grouped into a single annotation.
Thus, e.g. of annotations for an item for COCO is like as the follow:
[
    {....., bbox: [0.0, 1.0, 2.0, 3.0], segmentation: [1035, 30, 395, ...], ...},
    {....., bbox: [1.0, 1.0, 3.0, 2.0], segmentation: [10, 30, 3435, ...], ...},
    ....
]
When we import these annotations to Datumaro, we treat them to 4 annotation include 2 bbox and 2 segmentation (LRE or polygon)
To mark that some of them come from a single annotation, we give them ids thus, Datumaro annotations will be like the follow:
[
   {id: 1, type: bbox, ...},
   {id: 1, type: polygon, ...},
   {id: 2, type: bbox, ...},
   {id: 2, type: lre, ...},
]
when we export them back to COCO type, we should merge the annotations if they have the same id.
The code above merges the annotations.

When we determine the merged single bbox, we aggregate the candidates (anns), calculate bbox of each annon, and pick the maximum bbox to the merged bbox.
Thus, if a bbox from the segmentation annotation is larger then bbox from the bbox annotation,
we prefer the bbox from the segmentation and ignore bbox from the bbox annotation.
Because of this, when we import COCO, and export COCO back, many of bbox annotations are changed now.
I think this is weird.

The code changes above are to give a higher priority to bbox annotation when we determine the merged bbox annotation.
With this changes, we will ignore bbox from the segmentation annotation if we have any bbox annotation for an item.

wonjuleee (approved):
LGTM

vinnamkim (approved):
LGTM.

PR #773, opened 2022-11-30
Add data explorer feature

sooahleex:
### Checklist

### License

```python
#
```


datumaro/components/dataset_base.py
```diff
@@ -87,7 +87,11 @@ def __init__(
             media = point_cloud
 
         self.__attrs_init__(
-            id=id, subset=subset, media=media, annotations=annotations, attributes=attributes
+            id=id,
+            subset=subset,
+            media=media,
+            annotations=annotations,
+            attributes=attributes,
```
vinnamkim: Please revert those changes which are unrelated to this feature (and others).

datumaro/cli/commands/search.py
```diff
@@ -0,0 +1,120 @@
...
+        ]
+    }
+
+
+def search_command(args):
+    project = None
+    try:
+        project = scope_add(load_project(args.project_dir))
+    except ProjectNotFoundError:
+        if args.project_dir:
```
bonhunko, line 89: To my best understanding, if load_project() raises a ProjectNotFoundError and args.project_dir is not None or "",
Program will pass the outside of exception handling.
But I guess that 'project' may not be set correctly in this case.
Please give me an explain about this handling.

datumaro/cli/commands/search.py
```diff
@@ -0,0 +1,120 @@
...
+    results = searcher.search_topk(query_datasetitem, args.topk)
+
+    subset_list = []
+    id_list = []
+    result_path_list = []
+    for result in results:
+        subset_list.append(result.subset)
+        id_list.append(result.id)
+        result_path_list.append(result.media.path)
+    print(f"Most similar {args.topk} results of query in dataset: {result_path_list}")
```
bonhunko: Please use a logger, such as a 'warning', to print logs.

datumaro/cli/commands/search.py
```diff
@@ -0,0 +1,120 @@
...
+
+    visualizer = Visualizer(dataset, figsize=(20, 20), alpha=0)
+    fig = visualizer.vis_gallery(id_list, subset_list)
+
+    if args.save:
+        import numpy as np
```
bonhunko: Please consider to place all import to the top of source code file.

datumaro/components/launcher.py
```diff
@@ -36,12 +38,20 @@ def __init__(self, extractor, launcher, batch_size=1):
 
     def __iter__(self):
         for batch in take_by(self._extractor, self._batch_size):
-            inputs = np.array([np.atleast_3d(item.media.data) for item in batch])
+            inputs = []
+            for item in batch:
+                if type(item.media) in [Video, PointCloud, MultiframeImage]:
+                    raise MediaTypeError(
```
bonhunko: if 'item.media' has to be an object of Image, I think that we can check as the bellow:
if type(item.media) not Image:
    raise ...

datumaro/components/searcher.py
```diff
@@ -0,0 +1,160 @@
...
+
+            sims = np.zeros(shape=database_keys.shape[0] * len(query_hash_key_list))
+            for i, query_hash_key in enumerate(query_hash_key_list):
+                query_hash_key = self.unpack_hash_key(query_hash_key[0])
+                sims[i * db_len : (i + 1) * db_len] = calculate_hamming(
+                    query_hash_key, database_keys
+                )
+
+            def cal_ind(x):
+                if x >= db_len:
```
bonhunko: if db_len is >= 0, removing this if condition also provides the same result.

datumaro/components/searcher.py
```diff
@@ -0,0 +1,160 @@
...
+            for annotation in query.annotations:
+                if isinstance(annotation, HashKey):
+                    query_key = annotation.hash_key
+                    break
+
+            if not query_key:
+                try:
+                    if type(query.media) in [Video, PointCloud, MultiframeImage]:
+                        raise MediaTypeError(
+                            f"Media type should be Image, Current type={type(query.media)}"
```
bonhunko, line 128: if 'query.media' has to be an object of Image, I think that we can check as the below:
if type(query.media) not Image:
raise ...

tests/assets/searcher/clip_visual_ViT-B_32_interp.py
```diff
@@ -0,0 +1,15 @@
+# Copyright (C) 2023 Intel Corporation
+#
+# SPDX-License-Identifier: MIT
+
+from datumaro.components.annotation import AnnotationType, HashKey, LabelCategories
+
+
+def process_outputs(inputs, outputs):
```
bonhunko: 'inputs' is not utilized when calculating results, is it intended?

bonhunko (changes requested):
I left some minor reviews.

datumaro/components/dataset_base.py
sooahleex: Fix.

datumaro/cli/commands/search.py
sooahleex: Fix.

datumaro/components/launcher.py
sooahleex: Fix. I used `if not isinstance(item.media, Image)` instead of this.

vinnamkim (commented):
Please turn on CI green first.

tests/assets/searcher/clip_ViT-B_32_interp.py
```diff
@@ -0,0 +1,15 @@
+# Copyright (C) 2023 Intel Corporation
```
vinnamkim: Please move this file to somewhere under the package source directory, `./datumaro/~`.

datumaro/plugins/searcher.py
```diff
@@ -0,0 +1,96 @@
...
+        model_dir=None,
+        output_layers=None,
+    ):
+        model_name = "clip_visual_ViT-B_32"
+        if description or weights:
+            model_name = osp.splitext(description)[0]
+
+        url_folder = "http://s3.toolbox.iotg.sclab.intel.com/test/components-data/datumaro/models/"
+        if not model_dir:
+            model_dir = "tests/assets/searcher"
```
vinnamkim: Upload this model file to https://storage.openvinotoolkit.org/ and replace the model path in case of `model_dir == None` with a commonly used directory path, e.g. `~/.cache/datumaro`. Please refer to https://stackoverflow.com/questions/7401256/good-way-and-or-platform-independent-place-to-cache-data-from-web.

datumaro/cli/commands/search.py
sooahleex: I updated this like below based on your recommendation.
```
        if args.project_dir:
            log.info(
                f"Wrong argument: project_dir, {args.project_dir}, should be a path to project dir"
            )
            raise
```

datumaro/plugins/searcher.py
sooahleex: I changed to use `~/.cache/datumaro` folder instead of `tests/assets/searcher`. After I uploaded https://storage.openvinotoolkit.org/, I will update to this thread again.

requirements-core.txt
```diff
@@ -34,3 +34,4 @@ pandas>=1.1.5
 # Data explorer
 openvino==2022.3.0
 openvino-dev==2022.3.0
+transformers
```
vinnamkim: Is there any reason why `transformers` is added instead of [`tokenizers`](https://github.com/huggingface/tokenizers)? [`tokenizers`](https://github.com/huggingface/tokenizers) is much lighter.

bonhunko (dismissed):
LGTM! thank you!

requirements-core.txt
sooahleex: I tried to use same tokenizer which is used in CLIP. `transformers` supported pretrained CLIPTokenizer(https://huggingface.co/docs/transformers/v4.26.0/en/model_doc/clip#transformers.CLIPTokenizer) , so I used this.

vinnamkim: You can use it with `tokenizers` also. Please see https://github.com/huggingface/tokenizers/issues/1059.

vinnamkim: In addition, `openvino-dev==2022.3.0` is required?

datumaro/plugins/searcher.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+def normalize(x, axis=1, eps=1e-12):
+    denom = max(np.linalg.norm(x, axis=axis, keepdims=True), eps)
+    return x / denom
+
+
+def compute_hash(features):
+    features = encode_discrete(features)
+    features = normalize(features, axis=-1)
```
vinnamkim: Are those lines are same with the follwing?
```python
features = np.sign(features) / np.sqrt(len(features))
```

datumaro/plugins/searcher.py
```diff
@@ -0,0 +1,171 @@
...
+    hash_key = features >= 0
+    hash_key = hash_key * 1
```
vinnamkim: In addition, those four lines can be 
```
features = np.sign(features, dtype=np.int8)
features = np.clip(features, 0, None)
```

datumaro/components/launcher.py
```diff
@@ -36,12 +38,20 @@ def __init__(self, extractor, launcher, batch_size=1):
...
     def __iter__(self):
         for batch in take_by(self._extractor, self._batch_size):
-            inputs = np.array([np.atleast_3d(item.media.data) for item in batch])
+            inputs = []
+            for item in batch:
+                if not isinstance(item.media, Image):
+                    raise MediaTypeError(
+                        f"Media type should be Image, Current type={type(item.media)}"
+                    )
+                inputs.append(np.atleast_3d(item.media.data))
```
vinnamkim: Is there any reason to restrict `item.media` type to `Image`? This change may affect other parts.

datumaro/components/launcher.py
```diff
@@ -36,12 +38,20 @@ def __init__(self, extractor, launcher, batch_size=1):
...
+            inputs = np.array(inputs)
             inference = self._launcher.launch(inputs)
 
             for item, annotations in zip(batch, inference):
                 self._check_annotations(annotations)
-                yield self.wrap_item(item, annotations=annotations)
+                inferred_annotations = item.annotations + annotations
+                yield self.wrap_item(item, annotations=inferred_annotations)
```
vinnamkim: This change can also affect the others. Would you make a flag, e.g. `self.append (default=False)` and escalate it to `Dataset.run_model(..., append)`? Then, `Searcher` would be the only one who calls `dataset.run_model(..., append=True)` at this time.

datumaro/components/searcher.py
```diff
@@ -0,0 +1,161 @@
...
+            def cal_ind(x):
+                x = x % db_len
+                return x
+
+            ind = np.argsort(sims).tolist()
+            ind = list(map(cal_ind, ind))
+
+            item_list = np.array(self._item_list)[ind]
+            result = item_list[:topk].tolist()
+            return result
```
vinnamkim: I'll accept this for now. But, in the near future, you should improve this `O(n)` search algorithm (search by enumeration). I recommend you use `faiss` library. Please see https://github.com/facebookresearch/faiss/wiki/Binary-indexes.

datumaro/plugins/searcher.py
```diff
@@ -0,0 +1,171 @@
...
+
+        weights = osp.join(model_dir, model_name + ".bin")
+        if not osp.exists(weights):
+            cached_weights_url = osp.join(url_folder, model_name + ".bin")
+            log.info('Downloading: "{}" to {}\n'.format(cached_weights_url, weights))
+            download_file(cached_weights_url, weights)
+
+        if not interpreter:
+            openvino_plugin_samples_dir = "datumaro/plugins/openvino_plugin/samples"
+            interpreter = osp.join(openvino_plugin_samples_dir, "clip_ViT-B_32_interp.py")
```
vinnamkim: I don't know whether it is safe to use relative path than absolute path. I guess that it will be working only when you are on the package directory. Is it ok?

datumaro/cli/commands/search.py
```diff
@@ -0,0 +1,124 @@
...
+    log.info(f"Most similar {args.topk} results of query in dataset: {result_path_list}")
+
+    visualizer = Visualizer(dataset, figsize=(20, 20), alpha=0)
+    fig = visualizer.vis_gallery(id_list, subset_list)
+
+    if args.save:
+        fig.canvas.draw()
+        data = np.frombuffer(fig.canvas.tostring_rgb(), dtype=np.uint8)
+        data = data.reshape(fig.canvas.get_width_height()[::-1] + (3,))
+        save_image(osp.join("./searcher.png"), data, create_dir=True)
```
vinnamkim, line 125: I guess you should also print a list of `(id, subset)` of top-k candidates. If it is not going to this if statement (else of `if args.save`), what does a user obtain from this command?

datumaro/plugins/searcher.py
```diff
@@ -0,0 +1,171 @@
...
+        else:
+            inputs = inputs.squeeze()
+            if not inputs.any():
+                # media.data is None case
+                return None
+
+            if np.array(inputs).ndim == 2:
+                inputs = cv2.cvtColor(inputs, cv2.COLOR_GRAY2RGB)
+            elif np.array(inputs).ndim == 4:
+                inputs = cv2.cvtColor(inputs, cv2.COLOR_RGBA2RGB)
```
vinnamkim: Datumaro uses `BGR` format. 
https://github.com/openvinotoolkit/datumaro/blob/3aa474226f19cee00f2e8e78347d8a11a4c4ff8a/datumaro/util/image.py#L59-L86
If the model requires `RGB` format. You might also convert it when the number of channels is three too.

datumaro/components/annotation.py
```diff
@@ -207,6 +208,15 @@ class Label(Annotation):
     label: int = field(converter=int)
 
 
+@attrs(slots=True, order=False)
+class HashKey(Annotation):
+    _type = AnnotationType.hash_key
+    hash_key: List[str] = field(factory=list, validator=default_if_none(list))
```
vinnamkim: Can it be just `hash_key: np.ndarray`?

datumaro/components/annotation.py
```diff
@@ -207,6 +208,15 @@ class Label(Annotation):
...
+    label: Optional[int] = field(
+        converter=attr.converters.optional(int), default=None, kw_only=True
+    )
```
vinnamkim: Why do we need `label` in `HashKey`?

datumaro/plugins/searcher.py
sooahleex: I tried this, but the result of the suggestion is not same as that I intended. `encode_discrete` function is based on [this](https://github.com/zexuanqiu/CIBHash/blob/4c896ee1bab8e8a6060b7343aad06d483658ff78/model/CIBHash.py#L49-L50) and `normalize` function in here is same as [`normalize`](https://pytorch.org/docs/stable/generated/torch.nn.functional.normalize.html).

datumaro/plugins/searcher.py
sooahleex: I tried to use this like below. I think this result is not same as that I intended due to my answer above.
```
features = features.astype(np.int8)
features = np.sign(features)
features = np.clip(features, 0, None)
```

datumaro/components/launcher.py
sooahleex: Fix. I made `type_check` function for searcher launcher as your suggestion.

datumaro/components/launcher.py
sooahleex: Fix. I added `append_annotation` as instance variable which is only True for searcher.

datumaro/components/searcher.py
sooahleex: I see. I will add this in backlog.

datumaro/plugins/searcher.py
sooahleex: I made `get_samples_path` in `util/samples.py`. Please let me know if there is a better way.

datumaro/cli/commands/search.py
sooahleex: Users could get the plot result of candidates when `args.save = True`. And I updated to print `id, subset, path of image` through `log.info`.

datumaro/plugins/searcher.py
sooahleex: Fix. Thank you for reminding me this.

datumaro/components/annotation.py
sooahleex: `hash_key` consists of list of string like `['eced...']`. So I set hash_key as `List[str]`.

datumaro/components/annotation.py
sooahleex: I think No. But in [`_check_annotations`](https://github.com/openvinotoolkit/datumaro/blob/3aa474226f19cee00f2e8e78347d8a11a4c4ff8a/datumaro/components/launcher.py#L71-L72) function checked whether each annotation included label or not. I think I had no choice but to add `label` for `HashKey`.

datumaro/plugins/searcher.py
sooahleex: I updated this part as your suggestion. I removed `normalize` function.

datumaro/components/annotation.py
sooahleex: I removed `label` for `HashKey` and updated checking label part of annotation in `_check_annotation`.

datumaro/components/annotation.py
sooahleex: I set `hash_key: np.ndarray` and its shape is `(1,64)` with `np.uint8` due to using packbit.

vinnamkim (dismissed):
LGTM.

vinnamkim (dismissed):
LGTM.

vinnamkim (approved):
LGTM.

datumaro/plugins/searcher.py
sooahleex: Update data explorer models in [public storage](https://storage.openvinotoolkit.org/), please check https://github.com/openvinotoolkit/datumaro/pull/813.

PR #788, opened 2023-01-04
Fix TypeError in validate command

David-rn:
### Summary

When using the command `validate` in detection task, there is a TypeError in the dictionary that is being saved in a json file. 

To reproduce the error:
1. Create a project and import a dataset (cvat format in my case) for object detection
2. Execute the command: `datum validate -t detection`
3. It appears the following error: **TypeError: Dict key must be str**

The error appears because there are some keys that are `int` type, so a second check have been added alongside the check for type `tuple`.

### How to test

Taking into account the fix, if we execute the following command it generates the json file correctly instead of being empty:

`datum validate -t detection`

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/cli/commands/validate.py
+++ b/datumaro/cli/commands/validate.py
@@ -137,7 +137,7 @@ def validate_command(args):
     def _make_serializable(d):
         for key, val in list(d.items()):
             # tuple key to str
-            if isinstance(key, tuple):
+            if isinstance(key, tuple) or isinstance(key, int):
                 d[str(key)] = val
                 d.pop(key)
             if isinstance(val, dict):


vinnamkim (commented):
Hi, thanks for your contribution!

I tried to reproduce this error, but I found that there is no error with this CVAT dataset I created to test, [export_task.zip](https://github.com/openvinotoolkit/datumaro/files/10350426/export_task.zip).
Could you provide an example dataset to reproduce? Thank you!

David-rn:
Hi, thank you for the fast reply!

I've created this dummy [dataset](https://drive.google.com/file/d/1iiIc-06Se2F80lkWMbS3ON_bednE6Eov/view?usp=share_link) to reproduce the error I was getting. As I could not send the original data I have used some random images to point out the error. Some of the images are duplicated and also the bboxes for the class `building` does not correspond to an actual building in the image but the idea was to create a silly example to explain the problem to avoid creating a big dataset. 

I noticed that the error is raised when there are 'unusual' big/small bboxes as it will be written in the validation report under `bbox_distribution_in_label/items_far_from_mean` field as shown below. The problem is caused by this "0" that needs to be converted from `int` to a `str`.

```
"building": {
        "width": {
          "items_far_from_mean": {
            "('example_fail/image4', 'annotations')": {
              "0": 4429.0
            }
          },
...
```
In the same .zip file I have attached the validation_report file generated when it is added the suggested change.

vinnamkim (approved):
LGTM. I succeeded to reproduce this bug. It seems that `Annotation.id` has `int` type, so that the dictionary key can also have `int` type.

PR #796, opened 2023-01-19
Add MergeTileTransform

vinnamkim:
Signed-off-by: Kim, Vinnam <vinnam.kim@intel.com>

### Summary
- Resolves https://jira.devtools.intel.com/browse/CVS-100757.
- Add `MergeTile` transform

### How to test
Compare equivalence between the source dataset and the `Tile` -> `MergeTile` transformed dataset.

### Checklist

### License

```python
#
```


requirements-default.txt
```diff
@@ -1,2 +1,3 @@
 dvc>=2.7.0
+fsspec <= 2022.11.0; python_version < '3.8' # https://github.com/openvinotoolkit/datumaro/actions/runs/4003215621/jobs/6871114851#step:5:1647
```
wonjuleee, line 2: What is this for? As far as I know, adding dependency just before releases is definitely not preferred. @yunchu, Could you add the comment?

datumaro/plugins/tiling/untile.py
```diff
@@ -1,3 +0,0 @@
-# Copyright (C) 2023 Intel Corporation
-#
-# SPDX-License-Identifier: MIT
```
wonjuleee, line 3: why should we remove this?

datumaro/plugins/tiling/merge_tile.py
```diff
@@ -0,0 +1,303 @@
...
+        return attrs
+
+    def _merge_tiled_annotations(
+        self, items: List[DatasetItem], img_size: Tuple[int, int]
+    ) -> List[Annotation]:
+        anns_to_merge: Dict[AnnotationType, AnnotationsForMerge] = defaultdict(list)
+
+        for item in items:
+            roi = item.attributes.get("roi")
+            roi_box: sg.Polygon = sg.box(*xywh_to_x1y1x2y2(*roi))
```
wonjuleee, line 291: Under the consideration of not-perfectly divided tiles, e.g., divide a length 10 with a width 4, isn't it better to use x1y1x2y2 instead of xywh? in this case, rois (x,w) will be (0,4), (4,8), and (8,2)?

datumaro/plugins/tiling/merge_tile.py
```diff
@@ -0,0 +1,303 @@
...
+        return parser
+
+    def __init__(self, extractor):
+        super().__init__(extractor)
+
+    def __iter__(self):
+        items_to_merge = defaultdict(list)
+
+        for item in self._extractor:
+            item_id = item.attributes.get("tile_id")
```
wonjuleee, line 233: is here `tile_id`?

datumaro/plugins/tiling/merge_tile.py
```diff
@@ -0,0 +1,303 @@
...
+    def _merge_items(self, item_id: str, items: List[DatasetItem]) -> DatasetItem:
+        assert len(items) > 0
+
+        max_h = 0
+        max_w = 0
+        for item in items:
+            roi = item.attributes.get("roi")
+            x, y, w, h = roi
+            max_w = max(max_w, x + w)
+            max_h = max(max_h, y + h)
```
wonjuleee, line 251: w and h will be differ between tiles?

datumaro/plugins/tiling/merge_tile.py
```diff
@@ -0,0 +1,303 @@
...
+        x, y, w, h = roi_int
+        depth_img[y : y + h, x : x + w] = ann.image
+
+    return [ann.wrap(image=depth_img, attributes=deepcopy(ann.attributes))]
+
+
+def _merge_by_copy(
+    anns: AnnotationsForMerge, img_size: Tuple[int, int], *args, **kwargs
+) -> Union[Label, Caption]:
+    new_anns = {}
```
wonjuleee, line 177: should this be dict? if so, how about defaultdict()?

datumaro/plugins/tiling/merge_tile.py
```diff
@@ -0,0 +1,303 @@
...
+
+
+def _merge_depth_annotation(
+    anns: AnnotationsForMerge, img_size: Tuple[int, int], *args, **kwargs
+) -> List[DepthAnnotation]:
+    depth_img = np.zeros(shape=(img_size[0], img_size[1]))
+
+    for ann, roi_int, _ in anns:
+        x, y, w, h = roi_int
+        depth_img[y : y + h, x : x + w] = ann.image
```
wonjuleee, line 169: Will this be safe for putting ann.image to depth_img directly?

datumaro/plugins/tiling/merge_tile.py
```diff
@@ -0,0 +1,303 @@
...
+            )
+        ]
+
+    return merged_polylines
+
+
+def _merge_bbox(anns: AnnotationsForMerge, *args, **kwargs) -> List[Bbox]:
+    merged_bboxes = []
+
+    group_by_id = defaultdict(list)
```
wonjuleee, line 130: why this should be dict?

datumaro/plugins/tiling/merge_tile.py
vinnamkim: `tile_id` is the ID of the source item stored in tiled items when performing tile transformation. Therefore, we use them to merge when `MergeTile` transform

datumaro/plugins/tiling/merge_tile.py
vinnamkim: Yes. For a simple example, let's think the source image size is 11x11. Then, the 2x2 tiled images should have 5x5 or 6x6.

datumaro/plugins/tiling/merge_tile.py
vinnamkim: Sorry. I can't figure out what your claim is. Aren't `x1y1x2y2` and `xywh` formats perfectly convertible each other? In this line, I just convert `xywh(roi)` to `x1y1x2y2` to create `sg.Polygon` which only allows `x1y1x2y2` format.

datumaro/plugins/tiling/untile.py
vinnamkim: I removed this file and rename it as `merge_tile.py`.

requirements-default.txt
vinnamkim: Please see the recent CI failure https://github.com/openvinotoolkit/datumaro/actions/runs/4003215621/jobs/6871114851#step:5:1647.
`dvc` requires `fsspec` but `fsspec` is recently updated to `2023.1.0` from `2022.11.0` (https://pypi.org/project/fsspec/#history).
However, `fsspec==2023.1.0` tries to import `from fsspec.asyn import fsspec_loop` which is Python built-in module but not existed in Python=3.7.
Therefore, it is inevitable to downgrade `fsspec` for Python=3.7.

datumaro/plugins/tiling/merge_tile.py
vinnamkim: In the `Tile` transform, we split a bbox into multiple split bboxes inheriting the same id when on the edge of the tiled images. To merge them, we need to group the bboxes with the same id and this dictionary is used for it.

datumaro/plugins/tiling/merge_tile.py
vinnamkim: Yes, `ann.image` is a tiled depth annotation.

datumaro/plugins/tiling/merge_tile.py
vinnamkim: Yes, `dict` is needed to remove duplicate `Label`s and `Caption`s.
I think we should choose a data structure that is as lightweight as possible.
In this function, it seems that there is no need to automatically set the default value by `defaultdict()`.

requirements-default.txt
vinnamkim: [Update fsspec's license.txt in 3rd-party.txt](https://github.com/openvinotoolkit/datumaro/pull/796/commits/c797491f1a6a126d26fee3b533a33e33d2615e7b)

wonjuleee (approved):
LGTM

PR #800, opened 2023-02-01
Fixes ignored --loglevel by adding force=True when configuring with basicConfig.

### LINKED ISSUE #627: `loglevel` does not affect CLI output

Lucas-Steinmann:
Setting `force=True` removes any existing handlers from the root logger before configuring it.
If not specified and any handlers have been set before the call to basicConfig, then the changes are not applied.
During the import process, there are other calls to `logging.basicConfig`. Specifically, when tensorflow is not installed, then     [this line](https://github.com/openvinotoolkit/datumaro/blob/15c592609239ea23c3c49e30cc54c922d1ef5703/datumaro/components/extractor_tfds.py#L25) calls `log.debug`, which in turn calls `log.basicConfig`.

Since the call to basicConfig in `datumaro/cli/__main__.py` is the only deliberate call to basicConfig, it should be forced and overwrite any other inadvertent config changes.

### Summary
Resolves #627

Fixes bug, which prevented setting of log level when calling CLI tool.

### How to test
Before this patch is applied setting `--loglevel`did nothing (atleast if tensorflow is not installed, haven't tested with tf).
To test it just check if calling `datum --loglevel debug <any command>` works before and after applying this test.
 

### Checklist

### License

```python
#
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## \[Unreleased\]
 
 ### Fixed
-- Fix problem setting loglevel via CLI 
+- Fix problem setting loglevel via CLI
   (<https://github.com/openvinotoolkit/datumaro/pull/800>)
 
 ## 27/01/2023 - Release v0.5.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## \[Unreleased\]
 
+### Fixed
+- Fix problem setting loglevel via CLI 
+  (<https://github.com/openvinotoolkit/datumaro/pull/800>)
+
 ## 27/01/2023 - Release v0.5.0
 ### Added
 - Add Tile transformation
--- a/datumaro/cli/__main__.py
+++ b/datumaro/cli/__main__.py
@@ -41,7 +41,7 @@ class _LogManager:
         cls._define_loglevel_option(parser)
         args, _ = parser.parse_known_args(args)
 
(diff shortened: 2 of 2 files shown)


vinnamkim (approved):
Hi @Lucas-Steinmann, thank you for contributing to this project. It looks good to me but there is a formatting problem. Would you please fix it?

vinnamkim (changes requested):
Some failures are found in our unit tests.

Lucas-Steinmann:
I've fixed the formatting and I've revisited the tests.

The main problem is that the tests overwrite the logging configuration.
I guess unittests's `assertLog` was not written with the target in mind to run integration tests.
This is not meant as a critique, since I don't have a better idea on how to check the logging behavior.

I've tried to find a compromise, by not deleting handlers and instead only overwriting log level and formatters.
I'm not super happy with this, since this now tries to support some kind of prior modification to the logging configuration,
but cannot anticipate all possible prior modifications.

vinnamkim:
> I've fixed the formatting and I've revisited the tests.
> 
> The main problem is that the tests overwrite the logging configuration. I guess unittests's `assertLog` was not written with the target in mind to run integration tests. This is not meant as a critique, since I don't have a better idea on how to check the logging behavior.
> 
> I've tried to find a compromise, by not deleting handlers and instead only overwriting log level and formatters. I'm not super happy with this, since this now tries to support some kind of prior modification to the logging configuration, but cannot anticipate all possible prior modifications.

It seems that the previous implementation (`force=true`) removes an existing logging handler and create a new one. Therefore, `self.assertLogs()` stores a reference of the handler which will be removed by (`force=true`) subsequently. I think that this implementation not to remove the handler but modify it with our purpose is better than before.

vinnamkim (approved):
LGTM.

PR #807, opened 2023-02-08
Add Ellipse annotation type

vinnamkim:
### Summary
 - Related issue: https://jira.devtools.intel.com/browse/CVS-102621
 - Add Ellipse annotation type and update relevant other components who need to support Ellipse annotation type.

### How to test
I updated the existing tests related to this change.

### Checklist

### License

```python
#
```


datumaro/components/annotation.py
```diff
@@ -873,3 +880,116 @@ class DepthAnnotation(_ImageAnnotation):
...
+        y_points = self.c_y + l2 * np.sin(theta)
+
+        points = []
+        for x, y in zip(x_points, y_points):
+            points += [x, y]
+
+        return points
+
+    def iou(self, other: _Shape) -> Union[float, Literal[-1]]:
+        from datumaro.util.annotation_util import bbox_iou
```
bonhunko, line 995: Please move the import to the top of the source code file.

datumaro/components/annotation.py
```diff
@@ -673,7 +680,7 @@ def __eq__(self, other):
         self_polygon = sg.Polygon(self.get_points())
         other_polygon = sg.Polygon(other.get_points())
         inter_area = self_polygon.intersection(other_polygon).area
-        return abs(self_polygon.area - inter_area) == 0.0
+        return abs(self_polygon.area - inter_area) < 1e-7
```
bonhunko: Using `sys.float_info.epsilon` may be better than the hard-coded value 1e-7.

bonhunko (changes requested):
It looks fine to me. I just left some minor comments.

datumaro/components/annotation.py
vinnamkim: It's impossible because `datumaro.util.annotation_util` has circular dependency on this file.

datumaro/components/annotation.py
vinnamkim: `sys.float_info.epsilone` is `2.220446049250313e-16`. However, it's too tight so that there is a case that two polygons are same but the area difference can be greater than that.

vinnamkim: Instead of using it, I give a name to the number, `CHECK_POLYGON_EQ_EPSILONE = 1e-7`.

bonhunko: cool

datumaro/components/annotation.py
bonhunko: My concern is that `function iou()` may be called frequently and calling `import` is very expensive.
We may need to consider refactoring codes to remove circular imports next time.
At this time, please keep the current code.

I found a discussion about this in the stackflow.
https://stackoverflow.com/questions/31392035/python-workaround-for-circular-import

vinnamkim: As far as I know, Python doesn't reimport it until you explicitly force to do it. This circular dependency issue in `iou()` has been already existed in `Bbox`. I agree that this circular dependency should be resolved in the future, but it is not the scope of this PR.

bonhunko: Yes, It doesn't reimport but needs to check if a module is imported every import call.
Thus import in the frequently calling function is not recommended.

bonhunko (approved):

datumaro/components/annotation.py
vinnamkim: Ah ok I got it.

sooahleex (approved):
It looks good to me.

PR #812, opened 2023-02-17
Raise ImportError on importing malformed COCO directory

vinnamkim:
### Summary
 - Previously, COCOImporter implicitly set its `rootpath=""` and `images_dir=""` unless the directory structure is not well formed. This makes users cannot know what's happening although their dataset structure is malformed. As a result, after importing, users try to load its images but fail and don't know why.
 - This patch makes [COCO directory structure rule](https://github.com/cocodataset/cocoapi/blob/master/README.txt) more strictly at the dataset import level to prevent misbehavior in the next steps.
 - In addition, add a checker to check whether `Image` has correct dimensions: 2 or 3. There have been many test cases which let `Image()` not having 3 dimensions. This PR also fixes it.

### How to test
I added a unit test for it.

### Checklist

### License

```python
#
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   (<https://github.com/openvinotoolkit/datumaro/pull/804>)
 - Fix model remove CLI command
   (<https://github.com/openvinotoolkit/datumaro/pull/805>)
+- Raise ImportError on importing malformed COCO directory
+  (<https://github.com/openvinotoolkit/datumaro/pull/812>)
 
 ## 27/01/2023 - Release v0.5.0
 ### Added
--- a/datumaro/components/errors.py
+++ b/datumaro/components/errors.py
@@ -302,6 +302,10 @@ class MediaTypeError(DatumaroError):
     pass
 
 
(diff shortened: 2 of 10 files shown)


sooahleex (dismissed):
It looks good for me. I just have one question, you tested some case for wrong structure like `# Wrong structure: ./annotations -> ./labels` or `# Wrong structure: ./images -> ./imgs`, then do we need to consider those kinds of various wrong structure?

vinnamkim:
> It looks good for me. I just have one question, you tested some case for wrong structure like `# Wrong structure: ./annotations -> ./labels` or `# Wrong structure: ./images -> ./imgs`, then do we need to consider those kinds of various wrong structure?



> It looks good for me. I just have one question, you tested some case for wrong structure like `# Wrong structure: ./annotations -> ./labels` or `# Wrong structure: ./images -> ./imgs`, then do we need to consider those kinds of various wrong structure?

In those cases, this PR wants to raise `DatasetImportError` to tell a user that you should follow [COCO directory structure rule](https://github.com/cocodataset/cocoapi/blob/master/README.txt). After this PR, the only directory structure exactly fitted to [COCO directory structure rule](https://github.com/cocodataset/cocoapi/blob/master/README.txt) will be accepted.

wonjuleee (approved):

PR #830, opened 2023-02-27
Support PointCloud dataset by DatumaroBinary format

vinnamkim:
### Summary
 - Ticket no. 104650
 - Finish to support `PointCloud` dataset by `DatumaroBinary` format for compatibility with `Datumaro` format.

### Checklist

### License

```python
#
```


datumaro/plugins/data_formats/datumaro/exporter.py
```diff
@@ -64,68 +65,89 @@ def items(self):
...
+        """Implicitly change the media path and save it if save_media=True.
+        When done, revert it's path as before.
+        """
+        if item.media is None:
+            yield
+        elif isinstance(item.media, Image):
             image = item.media_as(Image)
             path = image.path
+
             if self._context._save_media:
```
wonjuleee, line 79: previously, when self._context is given by IDataset, was this correctly operated? The reason why I am asking is, IDataset doesn't have _save_media.

datumaro/plugins/data_formats/datumaro/exporter.py
```diff
@@ -64,68 +65,89 @@ def items(self):
...
-                path = self._context._make_image_filename(item)
-                self._context._save_image(
-                    item, osp.join(self._context._images_dir, item.subset, path)
+                # Temporarily update image path and save it.
+                image._path = osp.join(
+                    self._context._images_dir, item.subset, self._context._make_image_filename(item)
                 )
```
wonjuleee, line 83: is `_path` private variable of `MediaElement`? how about using `setter` and `getter` for path?

wonjuleee: It seems like an immutable variable after initialization.

datumaro/plugins/data_formats/datumaro/exporter.py
```diff
@@ -64,68 +65,89 @@ def items(self):
...
-                    item, osp.join(self._context._pcd_dir, item.subset, path)
+                # Temporarily update pcd path and save it.
+                pcd._path = osp.join(
+                    self._context._pcd_dir, item.subset, self._context._make_pcd_filename(item)
                 )
+                self._context._save_point_cloud(item, pcd.path)
+
+                # Temporarily update pcd related images paths and save them.
+                for i, img in enumerate(sorted(pcd.extra_images, key=lambda v: v.path)):
+                    img.__path = img.path
```
wonjuleee, line 101: I am sorry that I was failed to find `__path`. Could you let me know?

datumaro/plugins/data_formats/datumaro/exporter.py
```diff
@@ -64,68 +65,89 @@ def items(self):
...
 
-                    if img.has_data:
-                        img.save(
-                            osp.join(
-                                self._context._related_images_dir,
-                                item.subset,
-                                item.id,
-                                ri_desc["path"],
-                            )
-                        )
```
wonjuleee, line 120: We may not need to save the image if the image has data. Is this correct reason you removed?

datumaro/plugins/data_formats/datumaro/exporter.py
```diff
@@ -64,68 +65,89 @@ def items(self):
...
-            "id": item.id,
-            "annotations": annotations,
-        }
-
-        if item.attributes:
-            item_desc["attr"] = item.attributes
-
-        if isinstance(item.media, Image):
+    @contextmanager
+    def save_media(self, item: DatasetItem):
```
wonjuleee: We usually use `save_media` for boolean variable to save or not. How about changing the name for avoiding confusions?

datumaro/plugins/data_formats/datumaro_binary/mapper/media.py
```diff
@@ -0,0 +1,113 @@
...
+
+from datumaro.components.errors import DatumaroError
+from datumaro.components.media import Image, MediaElement, MediaType, PointCloud
+
+from .common import Mapper, StringMapper
+
+
+class MediaMapper(Mapper):
+    @classmethod
+    def forward(cls, obj: Optional[MediaElement]) -> bytes:
```
wonjuleee, line 16: I am a bit confused that we are going to forward only the media.path through this. Correct?

datumaro/plugins/data_formats/datumaro_binary/mapper/media.py
```diff
@@ -0,0 +1,113 @@
...
+        bytes_arr.extend(StringMapper.forward(obj.path))
+
+        return bytes(bytes_arr)
+
+    @classmethod
+    def backward(cls, _bytes: bytes, offset: int = 0) -> Tuple[Image, int]:
+        media_type, height, width = struct.unpack_from("<Iii", _bytes, offset)
+        assert media_type == cls.MEDIA_TYPE
+        size = (height, width)
+        offset += 12
```
wonjuleee: Could you clarify why this requires 12 bytes?

datumaro/plugins/data_formats/datumaro_binary/mapper/media.py
wonjuleee: If so, how about forwarding media.path instead of separating media types?

datumaro/plugins/data_formats/datumaro_binary/mapper/media.py
```diff
@@ -0,0 +1,113 @@
...
+        if obj is None:
+            return struct.pack(f"<I", MediaType.NONE)
+        elif obj.MEDIA_TYPE == MediaType.IMAGE:
+            return ImageMapper.forward(obj)
+        elif obj.MEDIA_TYPE == MediaType.POINT_CLOUD:
+            return PointCloudMapper.forward(obj)
+        elif obj.MEDIA_TYPE == MediaType.MEDIA_ELEMENT:
+            return MediaElementMapper.forward(obj)
+        else:
+            raise DatumaroError(f"{obj.MEDIA_TYPE} is not allowed for MediaMapper.")
```
wonjuleee: Could you expect the operation when path is too long to convert string to 8 byte?

datumaro/plugins/data_formats/datumaro/exporter.py
vinnamkim: Is it ok `context_save_media`?

datumaro/plugins/data_formats/datumaro/exporter.py
vinnamkim: Previously I wrongly typed it as `IDataset`. `Exporter` is correct one. Python `typing` does not affect to the code behavior. So, the code worked correctly but typing was incorrect.

datumaro/plugins/data_formats/datumaro/exporter.py
vinnamkim: This variable is used to save the original path temporariliy.

datumaro/plugins/data_formats/datumaro/exporter.py
vinnamkim: No all the things to save the media (images and point cloud) go into
```python
@contextmanager
    def save_media(self, item: DatasetItem):
```

datumaro/plugins/data_formats/datumaro_binary/mapper/media.py
vinnamkim: This acts a kind of factory for all kind of media types.
So, it has to write the media `type` first then goes on to write its data structure (including `path`, and etc....) subsequently.

datumaro/plugins/data_formats/datumaro_binary/mapper/media.py
```diff
@@ -0,0 +1,113 @@
...
+    MAGIC_SIZE_FOR_NONE = (-1583, -1597)
+    MEDIA_TYPE = MediaType.IMAGE
+
+    @classmethod
+    def forward(cls, obj: Image) -> bytes:
+        size = obj.size if obj.has_size else cls.MAGIC_SIZE_FOR_NONE
+
+        bytes_arr = bytearray()
+        bytes_arr.extend(struct.pack(f"<Iii", cls.MEDIA_TYPE, size[0], size[1]))
+        bytes_arr.extend(StringMapper.forward(obj.path))
```
vinnamkim: `path` will be serialized by this so that it can have an arbitrary size.

datumaro/plugins/data_formats/datumaro_binary/mapper/media.py
vinnamkim: Please see https://github.com/openvinotoolkit/datumaro/pull/830#discussion_r1128804561

datumaro/plugins/data_formats/datumaro_binary/mapper/media.py
vinnamkim: In L80, you can see we pack `<Iii` struct and it is unsigned int (4bits) + int (4bits) + int (4bits) = 12 bits.

datumaro/plugins/data_formats/datumaro/exporter.py
vinnamkim: Yes `path` was designed to be immutable but in this case, we have to make it mutable.
I think that the reason that we use `setter` in object oriented programming is when some member variables are dependent on the other variables. For example,

```python
class Whether:
def __init__(self, celsius: int):
     self._celsius = celsius
     self._text = "hot" if celsius > 30 else "warm"

@setter
def celsius(self, celsius: int):
     self._celsius = celsius
     self._text = "hot" if celsius > 30 else "warm"
```

You can see that `text` is dependent on `celsius` in this example. So, I decided to modify `_path` rather than creating the setter of it because it just increases the code lines like
```python
@setter
def path(self, path: str):
    self._path = path
```

datumaro/plugins/data_formats/datumaro/exporter.py
wonjuleee: it looks good to me.

datumaro/plugins/data_formats/datumaro/exporter.py
wonjuleee: Thank you for the clarification.

datumaro/plugins/data_formats/datumaro/exporter.py
wonjuleee: okay, thanks.

datumaro/plugins/data_formats/datumaro_binary/mapper/media.py
wonjuleee: I meant that we may be possible to locate forwarding `path` at here for all media types first, then the others (like size) can be forwarded according to each media types. What do you think?

datumaro/plugins/data_formats/datumaro_binary/mapper/media.py
wonjuleee: thanks.

datumaro/plugins/data_formats/datumaro/exporter.py
vinnamkim: Renamed it.

datumaro/plugins/data_formats/datumaro_binary/mapper/media.py
vinnamkim: I added `backward_dict()` to `MediaElementMapper` and make `ImageMapper` and `PointCloudMapper` to inherit `MediaElementMapper` to use the common function `backward_dict()`. It will be a way to handle common member variable `path` across all media classes.

datumaro/plugins/data_formats/datumaro/base.py
```diff
@@ -136,9 +136,7 @@ def _load_items(self, parsed):
                     related_images = [
                         Image(
                             size=ri.get("size"),
-                            path=osp.join(
-                                self._related_images_dir, self._subset, item_id, ri.get("path")
-                            ),
+                            path=osp.join(ri.get("path")),
```
bonhunko: unnecessary osp.join() call

bonhunko (dismissed):
Great work!
I left one minor comment.

datumaro/plugins/data_formats/datumaro/base.py
vinnamkim: Removed.

wonjuleee (approved):

PR #842, opened 2023-03-08
mark datumaro library version when exporting as datumaro

bonhunko:
### Summary

Resolves https://jira.devtools.intel.com/browse/CVS-105523

This PR adds the datumaro format version into an exported dataset.

### How to test

1. export a dataset as datumaro format
2. datumaro plugin will choose the proper reader of legacy datumaro format version

### Checklist

### License

```python
#
```


datumaro/components/dataset.py
```diff
@@ -1449,6 +1449,32 @@ def detect(path: str, *, env: Optional[Environment] = None, depth: int = 2) -> s
...
+
+        format_name = Dataset.detect(path, **detect_kwargs)
+
+        if env is None:
+            env = Environment()
+
+        version = env.detect_version(path=path, format_name=format_name)
+
+        return version
+
```
vinnamkim: I think that dataset versioning is only affected to Datumaro format so that we don't need to expose this to `Dataset` level. I guess this versioning can be dumpped to the `Dataset.info` dictionary, e.g., `key("dm_format_version")-value("1.0")`.

vinnamkim: Or if you really want to add this to `Dataset`(cvat has also versioning for its dataset format), I think it would be better to create a member variable, e.g., `Dataset._format_version` and make a setter for this variable in the dataset extractor, not here.

datumaro/components/importer.py
```diff
@@ -100,6 +102,10 @@ def detect(
 
         return FormatDetectionConfidence.LOW
 
+    @classmethod
+    def version(cls, path: str) -> str:
+        return cls.DEFAULT_VERSION
+
```
vinnamkim: I think you have to implement version checker too not just export/import version meta info.

datumaro/components/dataset.py
bonhunko: For `backward compatibility`, I think that we should be able to get version string from the dataset path before creating Dataset object. Thus, I think that member variable or using a `infos()` are not good place to store version.

datumaro/components/importer.py
bonhunko: That good idea. I'll prepare a version checker implementation.

datumaro/components/dataset.py
vinnamkim: My opinion is that the one who create `Dataset` is each `DatasetBase` (`Extractor` in the past, e.g., `DatumaroBase`) and it can decide that the current Datumaro has compatibility with the dataset version exactly when it tries to import the dataset from the file. For example,

```python
class DatumaroBase(DatasetBase):
    def __init__(self, path):
    version = parse_version(path)
    if not check_version_compatibility():
         raise DatasetImporterror(f"version {version} is not compatible. Please use the future version of Datumaro.")
    
    if version == "1.0":
         self._data = self.reader_v1_0()
    elif version == "1.1":
         self._data = self.reader_v1_1()
     ...
```

bonhunko: I agree that it's not necessary to expose the version to `Dataset` level.

Maybe some alternative could be generating a new file to store the overall meta-info of the Datumaro dataset not in the json for annotation.
@wonjuleee @vinnamkim Please share your opinion.

bonhunko: > My opinion is that the one who create `Dataset` is each `DatasetBase` (`Extractor` in the past, e.g., `DatumaroBase`) and it can decide that the current Datumaro has compatibility with the dataset version exactly when it tries to import the dataset from the file. For example,
> 
> ```python
> class DatumaroBase(DatasetBase):
>     def __init__(self, path):
>     version = parse_version(path)
>     if not check_version_compatibility():
>          raise DatasetImporterror(f"version {version} is not compatible. Please use the future version of Datumaro.")
>     
>     if version == "1.0":
>          self._data = self.reader_v1_0()
>     elif version == "1.1":
>          self._data = self.reader_v1_1()
>      ...
> ```

Yes, I guess that at this point, we don't fully generate Dataset object thus we should avoid storing the exported version in the Dataset obejct.

bonhunko: I discuss about this topic with @vinnamkim on the 1:1 chat.

I agree that the scope of versioning in the PR shouldn't affect all dataset types, only datumaro.
Thus I'll limit the changes on the datumaro plugin side.

About where to place the version string,
This should be volatility (removed after reading), I'll make a new field in json rather than using existing field, infos, because they are loaded to Dataset object while reading.

bonhunko:
I re-organized code as @vinnamkim suggested, please review this PR.

datumaro/plugins/data_formats/datumaro/base.py
```diff
@@ -28,9 +29,23 @@
 
 
 class DatumaroBase(SubsetBase):
+    REGACY_VERSION = "regacy"
```
vinnamkim:
```suggestion
    LEGACY_VERSION = "legacy"
```

datumaro/plugins/data_formats/datumaro/exporter.py
```diff
@@ -34,6 +34,7 @@
 from datumaro.components.exporter import Exporter
 from datumaro.components.media import Image, MediaElement, PointCloud
 from datumaro.util import cast, dump_json_file
+from datumaro.version import VERSION
```
vinnamkim: Datumaro "library" version and Datumaro "format" version should be independent. This is because there would be no change for Datumaro "format" even if Datumaro "library" version goes up. For example, the current version of CVAT is 2.2, but the CVAT image dataset format version is 1.1 (https://opencv.github.io/cvat/v2.3.0/docs/manual/advanced/xml_format/).

datumaro/plugins/data_formats/datumaro/base.py
```diff
@@ -53,6 +68,29 @@ def __init__(self, path):
...
+
+        Note that the regacy Datumaro doesn't store the version into exported dataset.
+        Thus it returns DatumaroBase.REGACY_VERSION when
+        """
+        versions = []
+        for annot_file in glob.glob(
+            osp.join(path, DatumaroPath.ANNOTATIONS_DIR, "**", "*" + DatumaroPath.ANNOTATION_EXT),
+            recursive=True,
+        ):
+            versions.append(parse_json_file(annot_file).get("dm_version", cls.REGACY_VERSION))
```
vinnamkim: This will make a JSON parsing step twice: 1) version check and 2) dataset content parsing. In the latest experiment I did, JSON parsing step for the Datumaro format of COCO instance segmentation takes ~10 seconds. This is non-negligible overhead. We need to make it do just once.

datumaro/plugins/data_formats/datumaro/exporter.py
bonhunko: Good point!

I placed DATUMARO_FORMAT_VERSION at https://github.com/openvinotoolkit/datumaro/pull/842/files#diff-07058dec3606d234ac036aa0116869b8aad42d7b9f5b3b85c820674fea958085R5

datumaro/plugins/data_formats/datumaro/base.py
bonhunko: Yes, exactly.

I let a DatumaroBase instance stores the parsed_json at the init() to avoid duplicated parsing.

datumaro/plugins/data_formats/datumaro/base.py
```diff
@@ -28,9 +28,24 @@
 
 
 class DatumaroBase(SubsetBase):
+    LEGACY_VERSION = "legacy"
```
vinnamkim, line 32:
```suggestion
    ALLOWED_VERSIONS = {LEGACY_VERSION, CURRENT_DATUMARO_FORMAT_VERSION}
    # If Datumaro format version goes up, it will be
    # ALLOWED_VERSIONS = {LEGACY_VERSION, 1.0, ..., CURRENT_DATUMARO_FORMAT_VERSION}
```

datumaro/plugins/data_formats/datumaro/base.py
```diff
@@ -28,9 +28,24 @@
...
         assert osp.isfile(path), path
 
+        self._parsed_anns = parse_json_file(path)
+        dm_version = self._get_dm_format_version(self._parsed_anns)
+
+        # when backward compatibility happend, we should implement version specific readers
+        if dm_version == DatumaroBase.LEGACY_VERSION:
+            self.default_reader(path=path)
+        else:
+            self.default_reader(path=path)
```
vinnamkim:
```suggestion
        if dm_version not in self.ALLOWED_VERSIONS:
            raise DatasetImportError(f"Datumaro format version of the given dataset is {dm_version}, but not supported by this Datumaro version: {<datumaro-library-version>}. The allowed datumaro format versions are {self.ALLOWED_VERSIONS}. Please install the latest Datumaro."
        else:
            self.default_reader(path=path)
```

tests/unit/data_formats/datumaro/test_datumaro_format.py
```diff
@@ -236,4 +238,36 @@ def test_inplace_save_writes_only_updated_data_with_transforms(self, test_dir, h
...
+
+        dataset.export(save_dir=test_dir, format="datumaro")
+        for annot_file in glob.glob(
+            osp.join(
+                test_dir, DatumaroPath.ANNOTATIONS_DIR, "**", "*" + DatumaroPath.ANNOTATION_EXT
+            ),
+            recursive=True,
+        ):
+            dm_base = DatumaroBase(path=annot_file)
+            assert dm_base._get_dm_format_version(dm_base._parsed_anns) == DATUMARO_FORMAT_VERSION
```
vinnamkim: I think this TC is already covered by the existing unit tests so that we can remove this. if you want to make a new TC, I recommend testing the broken version compatibility. For example,

```python
@pytest.fixture
def fxt_wrong_version_dir(fxt_test_datumaro_format_dataset, test_dir):
     dest_dir = osp.join(test_dir, "wrong_version")
     fxt_test_datumaro_format_dataset.export(dest_dir, "datumaro")
     < Somehow edit the annotation file to give some weird version to it >
     yield dest_dir

def test_version_compatibilty(self, fxt_wrong_version_dir):
     with pytest.raises(DatasetImportError):
            Dataset.import_from(fxt_wrong_version_dir, "datumaro")
```

bonhunko: This test is checking that
exported dm_dataset files have current datumaro format version or not.

As far as I know, existing unit tests doesn't check this.
if I missed some existing tests, please let me know.

BTW, testing the broken version compatibility is a good idea. I'll add the unit test.

vinnamkim: If the version compatibility is met, the dataset will be normally parsed without any error, and this is covered by the existing test. I think that code quality is important not only for the source but also for the test code (https://www.springboottutorial.com/code-quality-what-is-code-duplication).

bonhunko: If somehow datumaro version string is not stored in the exported dataset,

when exporting back, this dataset will be recognized as 'legacy' which is not appreciable but will be parsed well anyway.
Thus I think that we should check whether the exported datumaro has the current datumaro format version well or not.

vinnamkim: Basically, we need to satisfy the following two things after this PR.
1) Succeed to import the dataset without version info (legacy, what has been used)
2) Succeed to import the dataset with version = 1.0 (to-be)
3) Fail to import the dataset with version > 1.0

`2)` will be tested by the existing export and import TCs.

I thought that there has existed a TC for 1) but it doesn't. Sorry for my mistake.

So, it would be better to move `tests/assets/datumaro_dataset -> tests/assets/datumaro_dataset/legacy`, create a TC to import from `tests/assets/datumaro_dataset/legacy` and conduct `compare_datasets()`. For example,
https://github.com/openvinotoolkit/datumaro/blob/a44fc798760963d4a9510528be284946e3315bac/tests/unit/data_formats/test_common_semantic_segmentation_format.py#L135-L144

It is not appealing to call `dm_base._get_dm_format_version()` because it's a private function.

bonhunko: I added `test_version_compatibilty()` and `1)` tests as you suggested. Please have a look

bonhunko:
@vinnamkim after rebase to the lastest develop, I added `version` field to datumaro_binary format.
Please review this change which is not included previously https://github.com/openvinotoolkit/datumaro/pull/842/commits/98ddc5088f3c88831b05bd4a61ba97cea5954c3c

datumaro/plugins/data_formats/datumaro_binary/base.py
```diff
@@ -51,22 +60,30 @@ def _check_encryption_field(self):
...
+        if elem_type == dict:
+            header, _ = DictMapper.backward(_bytes)
+        elif elem_type == str:
+            header, _ = StringMapper.backward(_bytes)
+        else:
+            raise DatasetImportError(f"unsupported element type for binary format: {elem_type}")
         return header
 
+    def _read_version(self) -> str:
+        return self._read_header(elem_type=str)
```
vinnamkim:
```suggestion
        return self._read_header()["dm_format_version"]
```

datumaro/plugins/data_formats/datumaro_binary/base.py
```diff
@@ -51,22 +60,30 @@ def _check_encryption_field(self):
...
         len_byte = self._fp.read(4)
         _bytes = self._fp.read(struct.unpack("I", len_byte)[0])
         _bytes = self._crypter.decrypt(_bytes)
-        header, _ = DictMapper.backward(_bytes)
+        if elem_type == dict:
+            header, _ = DictMapper.backward(_bytes)
+        elif elem_type == str:
+            header, _ = StringMapper.backward(_bytes)
+        else:
+            raise DatasetImportError(f"unsupported element type for binary format: {elem_type}")
```
vinnamkim:
```suggestion
        header, _ = DictMapper.backward(_bytes)
```

datumaro/plugins/data_formats/datumaro_binary/exporter.py
```diff
@@ -44,20 +46,31 @@ def _dump_encryption_field(self) -> int:
...
 
-    def _dump_header(self, header: Any):
-        msg = DictMapper.forward(header)
+    def _dump_header(self, header: Any, elem_type: type):
+        if elem_type == str:
+            msg = StringMapper.forward(header)
+        elif elem_type == dict:
+            msg = DictMapper.forward(header)
+        else:
+            raise DatasetExportError(f"unsupported element type for binary format: {elem_type}")
```
vinnamkim:
```suggestion
        msg = DictMapper.forward(header)
```

datumaro/plugins/data_formats/datumaro_binary/exporter.py
```diff
@@ -44,20 +46,31 @@ def _dump_encryption_field(self) -> int:
...
 
         if self._crypter.key is not None:
             msg = self._crypter.encrypt(msg)
 
         length = struct.pack("I", len(msg))
         return self._fp.write(length + msg)
 
+    def _dump_version(self):
+        self._dump_header(DATUMARO_FORMAT_VERSION, str)
```
vinnamkim:
```suggestion
        self._dump_header({"dm_format_version": DATUMARO_FORMAT_VERSION})
```

vinnamkim: Please dump it with `dict` as same as `media_type` not to increase code branching.

datumaro/plugins/data_formats/datumaro_binary/base.py
```diff
@@ -24,13 +24,22 @@ def __init__(self, path: str, encryption_key: Optional[bytes] = None):
         self._crypter = Crypter(encryption_key)
         super().__init__(path)
 
+    def _get_dm_format_version(self, path: str) -> str:
+        with open(path, "rb") as fp:
+            self._fp = fp
+            self._check_signature()
+            self._check_encryption_field()
+            dm_format_version = self._read_version()
```
vinnamkim:
```suggestion
            self._fp = fp
            self._check_signature()
            dm_format_version = self._read_version()
            self._check_encryption_field()
```
Please check the version first before checking the encryption field.

datumaro/plugins/data_formats/datumaro_binary/exporter.py
```diff
@@ -75,9 +88,10 @@ def write(self):
                 self._fp = fp
                 self._sign()
                 self._dump_encryption_field()
-                self._dump_header(self.infos)
-                self._dump_header(self.categories)
-                self._dump_header(self._media_type)
+                self._dump_version()
```
vinnamkim, line 84: Same here. Please switch this line with `self._dump_encryption_field()`.

bonhunko: Cool! I switch encryption field and version field, and unify to dump as dictionary.

vinnamkim (approved):
LGTM.

PR #868, opened 2023-03-20
Fix detect unit tests to test false negatives as well

vinnamkim:
### Summary

 - Ticket no. 104511
 - To test false positive, change detect() test from assertIn() to assertEqual().
 - In addition, add false positive test code as well to some datasets which must have multiple candidates from
detect() in nature (e.g. kitti).
 - Refactor Importer.detect() confidence return logic.
 - Some addition fixes for dataset format plugins during this work.

### How to test
This changes includes tests as well.

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/components/format_detection.py
+++ b/datumaro/components/format_detection.py
@@ -35,14 +35,21 @@ class FormatDetectionConfidence(IntEnum):
     """
 
     NONE = 1
+    """
+    """
+    EXTREME_LOW = 5
+    """
+    EXTREME_LOW: This is currently only assigned to ImageDir format. This is because
+    ImageDir format can be detected in every image dataset format.
+    """
     LOW = 10
     """
-    The dataset seems to belong to the format, but the format is too loosely
+    LOW: The dataset seems to belong to the format, but the format is too loosely
     defined to be able to distinguish it from other formats.
     """
     MEDIUM = 20
     """
-    The dataset seems to belong to the format, and is likely not to belong
+    MEDIUM: The dataset seems to belong to the format, and is likely not to belong
     to any other format.
     """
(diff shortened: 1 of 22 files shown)


datumaro/components/format_detection.py
```diff
@@ -35,14 +35,21 @@ class FormatDetectionConfidence(IntEnum):
...
     LOW = 10
     """
-    The dataset seems to belong to the format, but the format is too loosely
+    LOW: The dataset seems to belong to the format, but the format is too loosely
     defined to be able to distinguish it from other formats.
     """
     MEDIUM = 20
     """
-    The dataset seems to belong to the format, and is likely not to belong
+    MEDIUM: The dataset seems to belong to the format, and is likely not to belong
```
wonjuleee, line 52: Do you have any plan to introduce `HIGH`?
Could you let me know what number (e.g., 1, 5, 10, and 20) means here?

vinnamkim: There is no plan for `HIGH`, and `EXTREME_LOW` was introduced to minimize change while maintaining momentum. The absolute number do not currently mean much, only relative differences between them to be matter. Therefore, the absolute numbers are given arbitrarily.

sooahleex (approved):
Looks good to me. I have one question, is there any reason to check whether images is under blacklisted folder only for imagenet dataset?

vinnamkim:
> Looks good to me. I have one question, is there any reason to check whether images is under blacklisted folder only for imagenet dataset?

Because `imagenet` format has very loose condition to be detected (sub-directories filled with images), we need to add the blacklist names to prevent it from being falsely detected.

wonjuleee (approved):
Thank you for your great work!

PR #878, opened 2023-03-22
Add prefix to secret key

vinnamkim:
### Summary
 - From @chuneuny-emily's stress test, we observed that the DatumaroBinary format encryption integration test occasionally fails because `argparse` expects one argument for `--encryption-key`.
```console
self = ArgumentParser(prog='datumaro_binary', usage=None, description=None, formatter_class=<class 'datumaro.cli.util.MultilineFormatter'>, conflict_handler='error', add_help=True)
status = 2, message = 'datumaro_binary: error: argument --encryption-key: expected one argument\n'
```
 - This is because the secret key can randomly start with `-` and `argparse` recognizes it as a new argument directive. 
```console
self = ArgumentParser(prog='datumaro_binary', usage=None, description=None, formatter_class=<class 'datumaro.cli.util.MultilineFormatter'>, conflict_handler='error', add_help=True)
arg_strings = ['--encryption-key', '-VaiTYwWN41UinEp6S5ke1zPnCh3hoZajOCEVptpyt8='], namespace = Namespace(encryption_key=None)

# '-VaiTYwWN41UinEp6S5ke1zPnCh3hoZajOCEVptpyt8=' starts with '-'.
```
 - This PR resolves this issue by adding `datum-` prefix to the secret key.
 - Remove a redundant file also.

### How to test
The existing tests cover this change.

### Checklist

### License

```python
#
```


chuneuny-emily (approved):

datumaro/components/crypter.py
```diff
@@ -7,19 +7,29 @@
...
     def __init__(self, key: Union[str, bytes]) -> None:
         if isinstance(key, str):
             key = key.encode()
-        self._key = key
+        if len(key) != self.FERNET_KEY_LEN:
+            raise DatumaroError(
+                f"Key length should be {self.FERNET_KEY_LEN}, "
+                f"but your key length is {len(key)} (key={key})."
+            )
+        self._key = key[self.KEY_PREFIX_LEN :]
```
wonjuleee, line 27: maybe `key[self.KEY_PREFIX_LEN:]`?

datumaro/components/crypter.py
```diff
@@ -29,17 +39,17 @@ def encrypt(self, msg: bytes) -> bytes:
 
     def handshake(self, key: bytes) -> bool:
         try:
-            return self.decrypt(key) == self._key
+            return self.decrypt(key) == self.key
```
wonjuleee, line 42: is this `self.key` instead of just `key`?

datumaro/components/crypter.py
vinnamkim: I did `black` for code formatting and it made it as it is.

datumaro/components/crypter.py
vinnamkim: This is because `handshake(key)` tries to check the equality between `self.key` and `key` which comes from the encrypted bytes in the annotation file. Therefore, we have to decrypt the bytes, `key` first (`self.decrypt(key)`) and compare it with `self.key`. Because both should have the `datum-` prefix, I changed it from `self._key` (without prefix) to `self.key` (with prefix).

wonjuleee (approved):

PR #867, opened 2023-03-20
Add publish to PyPI Github action

vinnamkim:
### Summary

- Automatically publish the package to PyPI when a release event occurs.

### Checklist

- [ ] I have added unit tests to cover my changes.​
- [ ] I have added integration tests to cover my changes.​
- [x] I have added the description of my changes into [CHANGELOG](https://github.com/openvinotoolkit/datumaro/blob/develop/CHANGELOG.md).​
- [ ] I have updated the [documentation](
  https://github.com/openvinotoolkit/datumaro/tree/develop/docs) accordingly

### License

- [ ] I submit _my code changes_ under the same [MIT License](
  https://github.com/openvinotoolkit/datumaro/blob/develop/LICENSE) that covers the project.
  Feel free to contact the maintainers if that's a concern.
- [ ] I have updated the license header for each file (see an example below)

```python
# Copyright (C) 2023 Intel Corporation
#
# SPDX-License-Identifier: MIT
```

PR #895, opened 2023-03-27
Change pypi_publish.yml to publish_sdist_to_pypi.yml

vinnamkim:
### Summary
 - The previous work https://github.com/openvinotoolkit/datumaro/pull/867 was failed because it try to build `wheel` on linux. However, PYPI doesn't accept this because https://peps.python.org/pep-0513/.
 - So, we need to just publish the source distribution (`sdist`) package only, not `wheel`.
 - Add if-else branching according to tag name for publishing to PyPI or TestPyPI
 - Fix to publish source distribution only
 - Upload source distribution to Github Release page too

### How to test
https://github.com/openvinotoolkit/datumaro/actions/runs/4528844180/jobs/7975997071
Upload to TestPyPI failure is because the same filename is not allowed to be uploaded again: https://github.com/pypa/packaging-problems/issues/74

### Checklist

### License

```python
#
```


cih9088 (approved):

bonhunko (approved):
LGTM

PR #897, opened 2023-03-27
Add multiprocessing to DatumaroBinaryBase

vinnamkim:
### Summary

- Ticket no. 106632
- Add `--num-workers` option to `DatumaroBinaryBase` for better performance.

### How to test
I updated the tests to cover this change.

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/plugins/data_formats/datumaro_binary/base.py
+++ b/datumaro/plugins/data_formats/datumaro_binary/base.py
@@ -5,9 +5,11 @@
 import os.path as osp
 import struct
 from io import BufferedReader
-from typing import Any, Dict, Optional
+from multiprocessing.pool import AsyncResult, Pool
+from typing import Any, Dict, List, Optional
 
 from datumaro.components.crypter import NULL_CRYPTER, Crypter
+from datumaro.components.dataset_base import DatasetItem
 from datumaro.components.errors import DatasetImportError
 from datumaro.components.media import Image, MediaElement, MediaType, PointCloud
 from datumaro.plugins.data_formats.datumaro_binary.format import DatumaroBinaryPath
@@ -16,15 +18,27 @@ from datumaro.plugins.data_formats.datumaro_binary.mapper.common import IntListM
 from datumaro.plugins.data_formats.datumaro_binary.mapper.dataset_item import DatasetItemMapper
 
 from ..datumaro.base import DatumaroBase
+from .format import DatumaroBinaryPath
 
 
 class DatumaroBinaryBase(DatumaroBase):
     """"""
 
-    def __init__(self, path: str, encryption_key: Optional[bytes] = None):
+    def __init__(self, path: str, encryption_key: Optional[bytes] = None, num_workers: int = 0):
+        """
+        Parameters
+        ----------
+        path
+            Directory path to import DatumaroBinary format dataset
+        encryption_key
+            If the dataset is encrypted, it (secret key) is needed to import the dataset.
+        num_workers
+            The number of multi-processing workers for import. If num_workers = 0, do not use multiprocessing.
+        """
         self._fp: Optional[BufferedReader] = None
         self._crypter = Crypter(encryption_key) if encryption_key is not None else NULL_CRYPTER
         self._media_encryption = False
+        self._num_workers = num_workers
         super().__init__(path)
 
     def _get_dm_format_version(self, path: str) -> str:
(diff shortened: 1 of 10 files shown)


datumaro/plugins/data_formats/datumaro_binary/base.py
```diff
@@ -91,27 +104,75 @@ def _read_media_type(self):
...
+        if self._num_workers > 0:
+            self._items = self._read_items_mp(blob_sizes, media_path_prefix)
+        else:
+            self._items = self._read_items_sp(blob_sizes, media_path_prefix)
+
+        for item in self._items:
+            if item.media is not None and self._media_encryption:
+                item.media.set_crypter(self._crypter)
+
+    def _read_items_mp(
```
JihwanEom, line 126: Don't we need try-except statement to avoid potential faults? Is it possible to prevent all the cases by the use of timeout?

site/content/en/docs/formats/datumaro_binary.md
```diff
@@ -123,6 +132,30 @@ datum import --format datumaro_binary <path/to/dataset> -- --encryption-key <sec
...
 `<secret-key>` is a 50-bytes long base64 encoded string prefixed with `datum-`. It is auto-generated in `<output/dir>/secret_key.txt` when the dataset is exported to DatumaroBinary format with `--encryption` option. You must have a correct `<secret-key>` to import the dataset encrypted by Datumaro.
 
+### Import datasets with multi-processing
+
+Using CLI
+
+```console
+# Import DatumaroBinary format dataset with 4 multi-processing workers
+datum create
+datum import --format datumaro_binary <path/to/dataset> -- --num-workers 4
```
JihwanEom, line 142: I'm not sure, but do we need additional hyphens? (`-- --num-workers 4`)
I hope.. 
`datum import --format datumaro_binary <path/to/dataset> --num-workers 4`

datumaro/plugins/data_formats/datumaro_binary/base.py
vinnamkim: In case of failure, we have to raise an error and notify to users that something goes wrong. The reason why I added a timeout is because the user must be notified if a situation occurs where the job never ends. If we don't have timeout, it just never stops and spins forever.

site/content/en/docs/formats/datumaro_binary.md
vinnamkim: This is a global rule from legacy for CLI arguments of all dataset formats. It would be good to make a ticket to enhance this rule for better UX.

datumaro/plugins/data_formats/datumaro_binary/format.py
```diff
@@ -23,6 +23,7 @@ class DatumaroBinaryPath:
     SECRET_KEY_FILE = "secret_key.txt"
 
     MAX_BLOB_SIZE = 2**20  # 1 Mega bytes
+    MP_TIMEOUT = 30.0  # 30 secs
```
sooahleex: Is there any reason to consider `MP_TIMEOUT` as 30 secs?

sooahleex (dismissed):
Looks good to me

datumaro/plugins/data_formats/datumaro_binary/format.py
vinnamkim: No reason, I just put in a value that I feel moderately long.

datumaro/plugins/data_formats/datumaro_binary/base.py
JihwanEom: I agree, but I have a concern about timeout. Is it okay with slow computing environment? It may require more time than our expectation to encrypt dataset. (or large dataset incomes.. with slow CPU or HDD.. )

vinnamkim: The timeout will be affected to each blob (not the entire dataset and the blob has maximum size), but I think it sounds reasonable. Let the timeout be longer to 5 minutes. Let this value to be merged into `develop`. I'll see this value is reasonable by running manual real-life tests for a production level dataset (COCO instance segmentation train/val).

JihwanEom: Thank you for detailed explanation!

JihwanEom (approved):
LGTM, thank you for awesome work!

sooahleex (approved):

PR #911, opened 2023-04-04
Refactor PointCloud

cih9088:
### Summary
Datumaro format and Datumaro binary format support two medias, namely `Image` and `PointCloud`.
The `PointCloud` should be able to digest functions, just like `Image`, for a single binary file export/import functionality, for example, Apache Arrow.

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/components/media.py
+++ b/datumaro/components/media.py
@@ -564,15 +564,48 @@ class PointCloud(MediaElement):
 
     def __init__(
         self,
-        path: str,
+        data: Union[bytes, Callable[[str], bytes], None] = None,
+        *,
+        path: Optional[str] = None,
         extra_images: Optional[List[Image]] = None,
         crypter: Crypter = NULL_CRYPTER,
     ):
+        if not isinstance(data, bytes):
+            assert path or callable(data), "PointCloud can not be empty"
+            assert data is None or callable(data)
+        self._bytes_data = data
         self._path = path
 
         self.extra_images: List[Image] = extra_images or []
 
         self._crypter = crypter
+
+    def get_bytes(self) -> bytes:
+        if callable(self._bytes_data):
+            return self._bytes_data(self.path)
+        if isinstance(self._bytes_data, bytes):
+            return self._bytes_data
+        else:
+            with open(self.path, "rb") as f:
+                bytes_data = f.read()
+            return bytes_data
+
+    def save(self, path, crypter: Crypter = NULL_CRYPTER):
+        if not crypter.is_null_crypter:
+            raise NotImplementedError(
+                f"{self.__class__.__name__} does not implement save() with non NullCrypter."
+            )
+
+        cur_path = osp.abspath(self.path) if self.path else None
+        path = osp.abspath(path)
+
+        os.makedirs(osp.dirname(path), exist_ok=True)
+        if cur_path is not None and osp.isfile(cur_path) and cur_path != path:
+            shutil.copyfile(cur_path, path)
+        elif self._bytes_data is not None:
+            data = self.get_bytes()
+            with open(path, "wb") as f:
+                f.write(data)
 
 
 class MultiframeImage(MediaElement):
--- a/datumaro/components/media.py
+++ b/datumaro/components/media.py
@@ -567,28 +567,41 @@ class PointCloud(MediaElement):
         data: Union[bytes, Callable[[str], bytes], None] = None,
         *,
         path: Optional[str] = None,
-        extra_images: Optional[List[Image]] = None,
+        extra_images: Union[List[Image], Callable[[], List[Image]], None] = None,
         crypter: Crypter = NULL_CRYPTER,
     ):
         if not isinstance(data, bytes):
             assert path or callable(data), "PointCloud can not be empty"
             assert data is None or callable(data)
+
         self._bytes_data = data
         self._path = path
 
-        self.extra_images: List[Image] = extra_images or []
+        self._extra_images = extra_images or []
 
         self._crypter = crypter
 
-    def get_bytes(self) -> bytes:
+    @property
+    def extra_images(self) -> List[Image]:
+        if callable(self._extra_images):
+            extra_images = self._extra_images()
+            assert isinstance(extra_images, list) and all(
+                [isinstance(image, Image) for image in extra_images]
+            )
+            return extra_images
+        return self._extra_images
+
+    def get_bytes(self) -> Optional[bytes]:
         if callable(self._bytes_data):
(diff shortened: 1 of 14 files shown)


datumaro/components/media.py
```diff
@@ -564,16 +564,62 @@ class PointCloud(MediaElement):
 
     def __init__(
         self,
-        path: str,
-        extra_images: Optional[List[Image]] = None,
+        data: Union[bytes, Callable[[str], bytes], None] = None,
```
vinnamkim: I think that it has been not good architecture of Datumaro to take various input types (`data` and `path`) in one class constructor. This makes the constructor code for lazy loading so complicated. In addition, we already saw a bad side-effect from the Datumaro user who is unfamiliar with our product, such as storing raw images in Geti dataset-ie microservice (using `data` for `Image`). Therefore, I have had a plan in my mind to refactor all `Media`s to have own single input data type. For example,

**As-is**
```python
class Image:
    def __init__(path, data):
```

**To-be**
```python
class ImageFromFile:
    def __init__(path):

# Default is ImageFromFile
Image = ImageFromFile

class ImageFromData:
    def __init__(data):
```

Therefore, I think that it would be good to apply this kind of refactoring in `PointCloud` first, if you would like to add `data` in this PR.

Additionally, I think that it would be also good for introducing Apache Arrow file format. For example,

```python
class ImageFromArrow(ImageFromFile):
     def __init__(path, address_in_file):
```

cih9088: Glad you have such a plan. Let me change the implementation in a way you suggested.

cih9088:
@vinnamkim How about this approach using factory?

vinnamkim:
> @vinnamkim How about this approach using factory?

LGTM. However, I think that it would be better explicitly to block `Media` from being created from its `constructor`. I mean a user must choose `PointCloud.from_file()` or `PointCloud.from_data()` to create `PointCloud` and `PointCloud()` is not allowed. Or, we can add warning for a user not to use the constructor directly. What do you think?

cih9088:
> LGTM. However, I think that it would be better explicitly to block `Media` from being created from its `constructor`. I mean a user must choose `PointCloud.from_file()` or `PointCloud.from_data()` to create `PointCloud` and `PointCloud()` is not allowed. Or, we can add warning for a user not to use the constructor directly. What do you think?

I thought the same thing and added assertion before your comment. What a coincidence.
How about now? If you agreed with this refactoring, I would propagate it to all other medias.

vinnamkim:
> > LGTM. However, I think that it would be better explicitly to block `Media` from being created from its `constructor`. I mean a user must choose `PointCloud.from_file()` or `PointCloud.from_data()` to create `PointCloud` and `PointCloud()` is not allowed. Or, we can add warning for a user not to use the constructor directly. What do you think?
> 
> I thought the same thing and added assertion before your comment. What a coincidence. How about now? If you agreed with this refactoring, I would propagate it to all other medias.

LGTM, but that expansion would be done in the next PR not to make a big one here.

cih9088:
> LGTM, but that expansion would be done in the next PR not to make a big one here.

Yes, of course. I will make separate PRs for each one. Thank you for your comments. Will ping you when the PR is ready.

datumaro/components/media.py
```diff
@@ -58,21 +88,32 @@ def ext(self) -> str:
         return osp.splitext(osp.basename(self.path))[1]
 
     @property
-    def is_encrypted(self) -> bool:
-        return not self._crypter.is_null_crypter
+    def data(self) -> Any:
```
vinnamkim:
```suggestion
    def data(self) -> Optional[Any]:
```

datumaro/components/media.py
```diff
@@ -58,21 +88,32 @@ def ext(self) -> str:
...
+        return f"{self.__class__.__name__}(path={self._path})"
 
-    def __eq__(self, other: object) -> bool:
-        # We need to compare exactly with this type
-        if type(other) is not __class__:  # pylint: disable=unidiomatic-typecheck
-            return False
-        return self._path == other._path
+
+class MediaElementFromDataMixin:
+    def __init__(self, data: Any, *args, **kwargs):
```
vinnamkim:
```suggestion
    def __init__(self, data: Union[Callable[[], Any], Any], *args, **kwargs):
```

I think we can replace `Any` with the union of data types used in all media types. It might be declared in the somewhere above.

datumaro/components/media.py
```diff
@@ -558,21 +605,109 @@ def __hash__(self):
...
+
 
 class PointCloud(MediaElement):
     _type = MediaType.POINT_CLOUD
 
     def __init__(
         self,
-        path: str,
-        extra_images: Optional[List[Image]] = None,
+        extra_images: Union[List[Image], Callable[[], List[Image]], None] = None,
```
vinnamkim:
```suggestion
        extra_images: Optional[List[Image], Callable[[], List[Image]]] = None,
```

datumaro/components/media.py
```diff
@@ -558,21 +605,109 @@ def __hash__(self):
...
+
+        cur_path = osp.abspath(self.path) if self.path else None
+        path = osp.abspath(path)
+
+        if cur_path is not None and osp.isfile(cur_path):
+            if cur_path != path:
+                os.makedirs(osp.dirname(path), exist_ok=True)
+                shutil.copyfile(cur_path, path)
+        else:
+            raise FileNotFoundError(errno.ENOENT, os.strerror(errno.ENOENT), cur_path)
```
vinnamkim, line 680: Not in this PR, but we could move saving extra images here in the future.

cih9088: Yes, I have a plan to save `extra_images` after refactoring `Image`.

datumaro/components/media.py
cih9088: https://github.com/openvinotoolkit/datumaro/pull/911/commits/741b44cec3b082a3047261bbf2ff42f82bbda70d

datumaro/components/media.py
cih9088: https://github.com/openvinotoolkit/datumaro/pull/911/commits/741b44cec3b082a3047261bbf2ff42f82bbda70d

datumaro/components/media.py
cih9088: https://github.com/openvinotoolkit/datumaro/pull/911/commits/741b44cec3b082a3047261bbf2ff42f82bbda70d

vinnamkim (approved):
LGTM.

sooahleex (approved):

PR #912, opened 2023-04-04
Refactor hl_ops.py

vinnamkim:
### Summary
 - Ticket no. 107274
 - Change the functions in hl_ops.py to static methods of `HLOps` to group them
 - Expose `HLOps` to the top, such as `import datumaro as dm; dm.HLOps.<functions>` for easy accessibility
 - **Suggestion for the better naming, directory structure, or refactoring direction are welcome** 

### How to test
I modified the existing tests to cover this.

### Checklist

### License

```python
#
```


datumaro/components/hl_ops/__init__.py
```diff
@@ -0,0 +1,231 @@
...
+        Returns: a wrapper around the input dataset
+        """
+
+        if isinstance(method, str):
+            if env is None:
+                env = Environment()
+            method = env.transforms[method]
+
+        if not (inspect.isclass(method) and issubclass(method, Transform)):
+            raise TypeError("Unexpected 'method' argument type: %s" % type(method))
```
JihwanEom: How about to use f-string?
```suggestion
            raise TypeError(f"Unexpected 'method' argument type: {type(method)}")
```

datumaro/components/hl_ops/__init__.py
```diff
@@ -0,0 +1,231 @@
...
+    def run_model(
+        dataset: IDataset,
+        model: Union[Launcher, Type[ModelTransform]],
+        *,
+        batch_size: int = 1,
+        append_annotation: bool = False,
+        **kwargs,
+    ) -> IDataset:
+        """
+        Applies a model to dataset items' media and produces a dataset with
```
JihwanEom:
```suggestion
        Applies a model to dataset item's media and produces a dataset with
```

datumaro/components/hl_ops/__init__.py
```diff
@@ -0,0 +1,231 @@
...
+                batch_size=batch_size,
+                append_annotation=append_annotation,
+                **kwargs,
+            )
+        elif inspect.isclass(model) and issubclass(model, ModelTransform):
+            return HLOps.transform(
+                dataset, model, batch_size=batch_size, append_annotation=append_annotation, **kwargs
+            )
+        else:
+            raise TypeError("Unexpected model argument type: %s" % type(model))
```
JihwanEom: Same as above, 
```suggestion
            raise TypeError(f"Unexpected model argument type: {type(model)}")
```

datumaro/components/hl_ops/__init__.py
```diff
@@ -0,0 +1,231 @@
...
+
+        if isinstance(format, str):
+            if env is None:
+                env = Environment()
+            exporter = env.exporters[format]
+        else:
+            exporter = format
+
+        if not (inspect.isclass(exporter) and issubclass(exporter, Exporter)):
+            raise TypeError("Unexpected 'format' argument type: %s" % type(exporter))
```
JihwanEom: Same as above,
```suggestion
            raise TypeError(f"Unexpected 'format' argument type: {type(exporter)}")
```

datumaro/components/hl_ops/__init__.py
vinnamkim: Replaced it with f-string.

datumaro/components/hl_ops/__init__.py
vinnamkim: Fixed.

datumaro/components/hl_ops/__init__.py
vinnamkim: Fixed.

datumaro/components/hl_ops/__init__.py
vinnamkim: Revised the model docstring for more clearance as follows.

Run the model on the dataset item media entities, such as images, to obtain pseudo labels and add them as dataset annotations.

JihwanEom (dismissed):

datumaro/components/hl_ops/__init__.py
```diff
@@ -0,0 +1,246 @@
...
+                - nothing + nothing = nothing,
+                - nothing + something = something
+                - something A + something B = conflict
+            - annotations are matched by value and shared
+            - in case of conflicts, throws an error
+
+        Returns: a wrapper around the input datasets
+        """
+
+        merger = ExactMerge()
```
wonjuleee, line 117: Do we only consider about `ExactMerge` for this?
Some datasets could have different categories and `ExactMerge` will be failed in this case.
So I mentioned that aggregator is going to be developed with `ExactMerge` (because subsets in a dataset will have exactly same categories), while it could be replaced with other Merger (such as UnionMerge) for the cases of merging two heterogeneous datasets. What do you think?

vinnamkim: The next work for ticket no. 107275 (Merge code refactoring) will deal with it. The goal of that task is to make `HLOps.merge()` tie `ExactMerge`, `UnionMerge` and `IntersectMerge` together at some higher level. For example,

```python
def merge(*dataset: IDataset, how: str = "exact"):
     if how == "exact"
           merger = ExactMerge()
     ...
```

vinnamkim: Also, I think `ExactMerge` can be perfectly replaced by `UnionMerge` in all use cases unless there is a user who want to stop merging explicitly if their datasets have different categories.

wonjuleee (dismissed):

sooahleex (dismissed):

sooahleex (approved):

wonjuleee (approved):

PR #922, opened 2023-04-07
Fix docs build warning from myst

yunchu:
### Summary

* Fixed cross-reference links on the docs.
* Added build-docs testenv to tox.ini
```bash
../datumaro$ tox -e build-docs
```

### Checklist

### License

```python
#
```


.pre-commit-config.yaml
```diff
@@ -1,5 +1,5 @@
 default_language_version:
-  node: system
+  node: 16.15.0
```
vinnamkim: Do we have to fix it?

yunchu: @vinnamkim, In case of my env, the node and its related things seem not to be installed correctly. so I've got some errors while running pre-commit hooks on my local. I think there would be some guys who have the same env like me. Is there any reasons to use system's node beyond the reducing burden to setup the node on the toxenv?

vinnamkim: My experience was the exact opposite of yours. If it requires the specific node version, I have to install that version mandatorily. What is the advantage of installing `node==16.15.0` to toxenv compared to installing or upgrading node on system using [nvm](https://github.com/nvm-sh/nvm) (or installing latest node to toxenv)? Does it have a better interface to install `node==16.15.0` on tox env rather than installing it on a system using nvm?

yunchu: > @vinnamkim, In case of my env, the node and its related things seem not to be installed correctly. so I've got some errors while running pre-commit hooks on my local. I think there would be some guys who have the same env like me. ~~Is there any reasons to use system's node beyond the reducing burden to setup the node on the toxenv?~~

sorry for the confusion. actually I've left totally wrong comment above. it's not related with tox.

I've installed node recently and checked that there are 3 different versions of node on my machine - 12.22.12, 16.18.0, 
and 18.15.0. and I've met the error "cannot find module node:fs" when run the pre-commit hooks with node:system in pre-commit config for both node to 16.18 or 18.15 on the system. but pre-commit worked well with node:16.15.0.
(I've used "n" instead of "nvm" and i don't have any ideas about what's different)

vinnamkim: I've been using `v18.15.0` but there is no issue for me.
```console
(datumaro) vinnamki@vinnamki:~/datumaro$ node -v
v18.15.0
```

yunchu: It's working on my side either. updated.

vinnamkim (dismissed):
LGTM.

wonjuleee (dismissed):

wonjuleee (approved):

vinnamkim (approved):

PR #931, opened 2023-04-12
Add Apache Arrow format support

cih9088:
### Summary
Support export into/import from `Apache Arrow`.

- `Subset.as_dataset()` inherits `source_path` and `format` from parent.
- Added export feature into `Apache Arrow` with sharding and multiprocessing support.
- Added import feature from `Apache Arrow`.
- Fixed comparing `Polygon` annotation error if they are invalid.
- Added `Sort` transform.
- A bit pettier `SimpleProgressReporter`.

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/components/format_detection.py
+++ b/datumaro/components/format_detection.py
@@ -228,7 +228,7 @@ class FormatDetectionContext:
 
         raise FormatRequirementsUnmet((requirement_desc,))
 
-    def is_root_dir(self):
+    def raise_if_root_is_not_dir(self):
         raise_if_path_is_not_dir(self.root_path)
 
     def require_file(
--- a/datumaro/components/format_detection.py
+++ b/datumaro/components/format_detection.py
@@ -142,6 +142,11 @@ class FormatDetectionUnsupported(_FormatRejected):
         return RejectionReason.detection_unsupported
 
 
+def raise_if_path_is_not_dir(path):
+    if not osp.isdir(path):
+        raise FormatRequirementsUnmet((f"root path {path} must refer to a directory.",))
+
+
 class FormatDetectionContext:
     """
     An instance of this class is given to a dataset format detector.
@@ -222,6 +227,9 @@ class FormatDetectionContext:
         self._start_requirement("fail")
 
         raise FormatRequirementsUnmet((requirement_desc,))
+
+    def is_root_dir(self):
+        raise_if_path_is_not_dir(self.root_path)
 
     def require_file(
         self,
@@ -456,9 +464,6 @@ def apply_format_detector(
     by the detector.
     """
     context = FormatDetectionContext(dataset_root_path)
-
-    if not osp.isdir(dataset_root_path):
-        context.fail(f"root path {dataset_root_path} must refer to a directory")
 
     return detector(context) or FormatDetectionConfidence.MEDIUM
 
--- a/datumaro/components/format_detection.py
+++ b/datumaro/components/format_detection.py
@@ -142,11 +142,6 @@ class FormatDetectionUnsupported(_FormatRejected):
         return RejectionReason.detection_unsupported
 
 
-def raise_if_path_is_not_dir(path):
-    if not osp.isdir(path):
-        raise FormatRequirementsUnmet((f"root path {path} must refer to a directory.",))
-
-
 class FormatDetectionContext:
     """
     An instance of this class is given to a dataset format detector.
@@ -227,9 +222,6 @@ class FormatDetectionContext:
         self._start_requirement("fail")
 
         raise FormatRequirementsUnmet((requirement_desc,))
-
-    def raise_if_root_is_not_dir(self):
-        raise_if_path_is_not_dir(self.root_path)
 
     def require_file(
         self,
@@ -464,6 +456,9 @@ def apply_format_detector(
     by the detector.
     """
     context = FormatDetectionContext(dataset_root_path)
+
+    if not osp.isdir(dataset_root_path):
+        context.fail(f"root path {dataset_root_path} must refer to a directory")
 
     return detector(context) or FormatDetectionConfidence.MEDIUM
 
--- a/datumaro/components/progress_reporting.py
+++ b/datumaro/components/progress_reporting.py
@@ -146,8 +146,14 @@ class SimpleProgressReporter(ProgressReporter):
 
     def report_status(self, progress: int):
         status = str(self._desc) if self._desc else ""
-        status += f"{progress}/{self._total} ({progress/self._total*100:.2f}%)"
+        status += (
+            f" {progress:0{len(str(self._total))}d}/{self._total}"
+            f" ({progress/self._total*100: 6.2f}%)"
+        )
         print(status)
+
+    def finish(self):
+        self.report_status(self._total)
 
     def split(self, count: int):
         return (SimpleProgressReporter(self._period, self._interval),) * count
--- a/datumaro/plugins/data_formats/arrow/arrow_dataset.py
+++ b/datumaro/plugins/data_formats/arrow/arrow_dataset.py
@@ -1,0 +1,204 @@
+# Copyright (C) 2023 Intel Corporation
+#
+# SPDX-License-Identifier: MIT
+
+import copy
+import errno
+import os
+from collections.abc import Iterator
+from typing import Dict, List, Optional, Tuple, TypeVar, Union, overload
+
+import numpy as np
+import pyarrow as pa
+
+
+PathLike = TypeVar("PathLike", str, os.PathLike, None)
+
+
+def _binary_search(arr, x, low=None, high=None, fn=None):
+    low = 0 if low is None else low
+    high = len(arr) - 1 if high is None else high
+    if fn is None:
+
+        def fn(arr, x):
+            return arr[x]
+
+    while low <= high:
+        mid = (high + low) // 2
+        val = fn(arr, mid)
+        if val > x:
+            high = mid - 1
+        else:
+            low = mid + 1
+    return high
+
+
+class ArrowDatasetIterator(Iterator):
+    def __init__(self, dataset: "ArrowDataset"):
+        self.dataset = dataset
+        self.index = 0
+
+    def __next__(self) -> Dict:
+        if self.index >= len(self.dataset):
+            raise StopIteration
+        item = self.dataset[self.index]
+        self.index += 1
+        return item
+
+
+class ArrowDataset:
+    def __init__(
+        self,
+        files: Union[PathLike, List[PathLike]],
+        keep_in_memory: bool = False,
+    ):
+        assert files
+        if not isinstance(files, (list, tuple)):
+            files = [files]
+
+        self._files = files
+        self._keep_in_memory = keep_in_memory
+
+        self.unload_arrows()
+        self.load_arrows()
+
+    @property
+    def table(self) -> pa.Table:
+        self.load_arrows()
+        return self.__table
+
+    @property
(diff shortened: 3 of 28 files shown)


datumaro/components/format_detection.py
```diff
@@ -457,9 +465,6 @@ def apply_format_detector(
     """
     context = FormatDetectionContext(dataset_root_path)
 
-    if not osp.isdir(dataset_root_path):
-        context.fail(f"root path {dataset_root_path} must refer to a directory")
```
vinnamkim: Why do we remove this line?

datumaro/components/progress_reporting.py
```diff
@@ -146,9 +146,15 @@ def start(self, total: int, *, desc: Optional[str] = None):
 
     def report_status(self, progress: int):
         status = str(self._desc) if self._desc else ""
-        status += f"{progress}/{self._total} ({progress/self._total*100:.2f}%)"
+        status += (
+            f" {progress:0{len(str(self._total))}d}/{self._total}"
```
vinnamkim, line 150: Do we need `progress:0` <- `"0"`?

datumaro/components/progress_reporting.py
```diff
@@ -146,9 +146,15 @@ def start(self, total: int, *, desc: Optional[str] = None):
...
+            f" ({progress/self._total*100: 6.2f}%)"
```
vinnamkim, line 151: Why is it `6.2f` than `.2f`?

datumaro/plugins/data_formats/arrow/arrow_dataset.py
```diff
@@ -0,0 +1,203 @@
...
+            return arr[x]
+
+    while low <= high:
+        mid = (high + low) // 2
+        val = fn(arr, mid)
+        if val > x:
+            high = mid - 1
+        else:
+            low = mid + 1
+    return high
```
vinnamkim, line 33: Do we have this in Python core library?
https://docs.python.org/3/library/bisect.html

datumaro/plugins/data_formats/arrow/arrow_dataset.py
```diff
@@ -0,0 +1,203 @@
...
+
+        if isinstance(key, range):
+            if key.step == 1 and key.stop >= key.start:
+                return self.slice(key.start, key.stop - key.start)
+
+        if isinstance(key, int):
+            if key < 0:
+                key += len(self)
+            return self.slice(key, 1)[0]
+        elif isinstance(key, range):
```
vinnamkim: I think this `elif` branch should move to L183 for better readability.

datumaro/plugins/data_formats/arrow/arrow_dataset.py
```diff
@@ -0,0 +1,203 @@
...
+        for file in self._files:
+            if not os.path.exists(file):
+                raise FileNotFoundError(errno.ENOENT, os.strerror(errno.ENOENT), file)
+
+            if self._keep_in_memory:
+                stream = pa.input_stream(file)
+            else:
+                stream = pa.memory_map(file)
+            opened_stream = pa.ipc.open_stream(stream)
+            table_ = opened_stream.read_all()
```
vinnamkim, line 120: What is the difference between `self._keep_in_memory=True or False`? I think that there seems no difference after `table_ = opened_stream.read_all()` because it seems to load everything on the memory from the file.

vinnamkim, line 119: What is the difference between `self._keep_in_memory=True or False`? I think that there seems no difference after `table_ = opened_stream.read_all()` because it seems to load everything on the memory from the file.

datumaro/plugins/data_formats/arrow/arrow_dataset.py
```diff
@@ -0,0 +1,203 @@
...
+            if key < 0:
+                key += len(self)
+            return self.slice(key, 1)[0]
+        elif isinstance(key, range):
+            return [
+                self.slice(key_, 1)[0] if key_ >= 0 else self.slice(key_ + len(self), 1)[0]
+                for key_ in key
+            ]
+
+        raise KeyError()
```
vinnamkim: I think we need to make a unit test for this function because it has many branches.

datumaro/plugins/data_formats/arrow/exporter.py
```diff
@@ -0,0 +1,418 @@
...
+            help=f"Image encoding scheme. (default: {cls.DEFAULT_IMAGE_EXT})",
+            choices=cls.AVAILABLE_IMAGE_EXTS,
+        )
+
+        #  parser.add_argument(
+        #      "--split-by-subsets",
+        #      action="store_true",
+        #      default=False,
+        #      help="Split arrow dataset by subset. (default: %(default)s)",
+        #  )
```
vinnamkim: Can we remove?

tests/unit/data_formats/arrow/test_arrow_format.py
```diff
@@ -0,0 +1,331 @@
...
+            # python on windows in github action can not remove arrow files
+            # with an access denied error but files can be removable from CMD.
+            # It might be related to not closed arrow writer object
+            # though some files are removable.
+            if platform.system() != "Windows":
+                return
+            for file in os.listdir(save_dir):
+                if not file.endswith(".arrow"):
+                    continue
+                os.system(f"rm -rf {os.path.join(save_dir, file)}")
```
vinnamkim: I think you can resolve this in more `pytest`-onic way by overriding `test_dir` fixture.

datumaro/components/format_detection.py
cih9088: `dataset_root_path` can be a file not just a directory to provide import feature from a single arrow file.

datumaro/components/progress_reporting.py
cih9088: Would you elaborate it more? I don't get what you mean.

datumaro/components/progress_reporting.py
cih9088: To align whole number part as well. It just makes thing a bit prettier like below.
```
(  1.00%)
( 10.00%)
(100.00%)
```

datumaro/plugins/data_formats/arrow/arrow_dataset.py
cih9088: Ah, yes, I totally forget about `bisect`. Let me check if I can replace it with `bisect`.

datumaro/components/progress_reporting.py
vinnamkim: Sorry, I was mistaken that it was outside the bracket, e.g. `f"progress:0{len(str(self._total))}d}/{self._total}"`

datumaro/components/format_detection.py
vinnamkim: `root_path` literally means directory, and entering a directory path has traditionally been the behavior of this function that globally has been affecting to all Datumaro components. It looks like it might be possible to follow this rule in the ArrowDataset (by fixing filename pattern).

datumaro/plugins/data_formats/arrow/arrow_dataset.py
cih9088: Fixed in https://github.com/openvinotoolkit/datumaro/pull/931/commits/8799b145fcae5765b5952ebce33b0c7e71e5c446

datumaro/plugins/data_formats/arrow/exporter.py
cih9088: Fixed in https://github.com/openvinotoolkit/datumaro/pull/931/commits/0a8eda950dd54f0b83c2762fb1731f7a61bdc0e9

tests/unit/data_formats/arrow/test_arrow_format.py
cih9088: Thanks for the suggestion. Fixed in https://github.com/openvinotoolkit/datumaro/pull/931/commits/d5b94307e4e7c29a56e974fea8f047c43f2bb9ba

datumaro/plugins/data_formats/arrow/arrow_dataset.py
cih9088: In the case of `keep_in_memory=False`, `read_all()` does not allocate new memory since it is a memory mapped file. The file is treated as a virtual memory and only relevant parts would be paged into memory, not the whole file.

In the case of `keep_in_memory=True`, `read_all()` does allocate new memory and read all the contents in it into the allocated memory.

datumaro/components/format_detection.py
cih9088: I changed it in this way because the dataset can be saved into a single arrow file. I felt that since we can export into a single file, we need to also import from a single file as well.

Furthermore, doc indicates it can receive a single file as well.
https://github.com/openvinotoolkit/datumaro/blob/37b62bb5fef85211ebc8f804bf0e20a02e2fb7ad/datumaro/components/dataset.py#L1350

Limiting the input as directory for all formats would not be a preferable solution, I believe. But if you still think the previous was the right way, I will revert the change. In arrow format perspective, it does not matter much.

vinnamkim: My opinion is that I don't understand why a change that affects the overall behavior is necessary in this PR although there are ways to avoid it. Even defining the data format based on the directory structure is a rule followed by all datasets except `ArrowDataset`.  Therefore, it is recommended to inspect the directory structure rather than probing the data file in `detect()`.  In your case, for example,
```python
def detect(root_path):
    for fname in os.listdir(root_path):
         if os.path.splitext(fname)[-1] == ".arrow":
              return confidence.MEDIUM
        ....
```
 Lastly, the only one who consumes `apply_format_detector()` is `detect_dataset_format()` and `detect_dataset_format()` is only consumed by
https://github.com/openvinotoolkit/datumaro/blob/37b62bb5fef85211ebc8f804bf0e20a02e2fb7ad/datumaro/components/environment.py#L252-L277
and L276 shows that it doesn't like a file path.

datumaro/plugins/data_formats/arrow/arrow_dataset.py
vinnamkim: Because there is no time, please add TODO comment and do it in the future.

datumaro/plugins/data_formats/arrow/arrow_dataset.py
vinnamkim: Because there is no time, please add TODO comment and do it in the future.

datumaro/components/format_detection.py
cih9088: Fair point. Reverted in https://github.com/openvinotoolkit/datumaro/pull/931/commits/44eff3a90ef23a86adeea5ca3a4258b80f88f811

datumaro/plugins/data_formats/arrow/arrow_dataset.py
cih9088: Added in https://github.com/openvinotoolkit/datumaro/pull/931/commits/c5730b7fc806bc7de4e486eafeeed0cfb3e726e6

datumaro/plugins/data_formats/arrow/arrow_dataset.py
cih9088: Added in https://github.com/openvinotoolkit/datumaro/pull/931/commits/c5730b7fc806bc7de4e486eafeeed0cfb3e726e6

vinnamkim (approved):
LGTM.

wonjuleee (approved):

PR #940, opened 2023-04-15
Refactor explorer CLI command

sooahleex:
### Summary

- Ticket no.107264
- Modify `search` or `searcher` to `explore` or `explorer` to fix term
- Update explorer to run w/ or w/o project

### Checklist

### License

```python
#
```


vinnamkim (commented):
> Update explorer to run w/ or w/o project

Should this command be "w/ project" only?

I think this feature remains at a very primitive level and doesn't reach a production level. For example,

1. Introduction of a vector search library such as Faiss to avoid linear time search complexity. I told you about this last time when we merged this feature to develop.
2.  Related to the "'w/ project' only?" problem above, there has been missing the part about how and where to store the computed feature vectors. We cannot compute feature vector again and again each time the explore command is executed. So, we have to store them somewhere, and the concept of project seems most appropriate for it.

sooahleex:
> > Update explorer to run w/ or w/o project
> 
> Should this command be "w/ project" only?
> 
> I think this feature remains at a very primitive level and doesn't reach a production level. For example,
> 
> 1. Introduction of a vector search library such as Faiss to avoid linear time search complexity. I told you about this last time when we merged this feature to develop.
> 2. Related to the "'w/ project' only?" problem above, there has been missing the part about how and where to store the computed feature vectors. We cannot compute feature vector again and again each time the explore command is executed. So, we have to store them somewhere, and the concept of project seems most appropriate for it.

I agreed to the first one. I will update algorithm for search. But it is not related to refactoring cli for explorer.
And for second one, I thought `search` command is only for one time command. so, we agreed before that compute feature vector again and again whenever explore command is executed.

vinnamkim:
> I agreed to the first one. I will update algorithm for search. But it is not related to refactoring cli for explorer. And for second one, I thought `search` command is only for one time command. so, we agreed before that compute feature vector again and again whenever explore command is executed.

First of all, my question is not directly related to this PR. However, this feature has been remained unproductive for a long time. Therefore, I raised my concern since we are building up the documentation for our feature for this sprint, but this seems not ready for production. I don't know what the documentation means if the product quality is low.

I don't understand why the first concern is agreeable but the second is not. They are on the same page that is **saving computation**. Rather than replacing linear search with an elaborate search algorithm, I think it 10,000 times much more reduces the cost to save the pre-computed feature vector in the project and avoid calculating the feature vector for all dataset samples each time the explorer command is executed (also the vector search library should build an index and this will take longer than 1-time linear search if not reused).

To achieve this, I'm not asking inventing something new. I think that it can be sufficiently dealt with using basic methods.

sooahleex:
> > I agreed to the first one. I will update algorithm for search. But it is not related to refactoring cli for explorer. And for second one, I thought `search` command is only for one time command. so, we agreed before that compute feature vector again and again whenever explore command is executed.
> 
> First of all, my question is not directly related to this PR. However, this feature has been remained unproductive for a long time. Therefore, I raised my concern since we are building up the documentation for our feature for this sprint, but this seems not ready for production. I don't know what the documentation means if the product quality is low.
> 
> I don't understand why the first concern is agreeable but the second is not. They are on the same page that is **saving computation**. Rather than replacing linear search with an elaborate search algorithm, I think it 10,000 times much more reduces the cost to save the pre-computed feature vector in the project and avoid calculating the feature vector for all dataset samples each time the explorer command is executed (also the vector search library should build an index and this will take longer than 1-time linear search if not reused).
> 
> To achieve this, I'm not asking inventing something new. I think that it can be sufficiently dealt with using basic methods.

Okay, then I will close this pr and revisit this after implementing linear search algorithm. However, for the second one, we avoid to use database to store hash for all datasetitems. So I don't know there is any alternative for this. If we find anything to apply, I will revisit it too.

PR #981, opened 2023-04-28
Save and load hashkey for explorer

sooahleex:
### Summary

- Ticket no.107264
- Save and load `HashKey` for dataset after `explore` command
- Get list of dataset in explorer
- Export `HashKey` annotation in `datumaro` format
- Usecase for explorer
  - w/w.o target for explore command
  - explore -> add -> explore
  - explore -> merge -> explore
  - Support versioning

### Checklist

### License

```python
#
```


datumaro/cli/commands/explore.py
```diff
@@ -101,24 +110,21 @@ def explore_command(args):
...
+        log.info(f"id: {result.id} | subset: {result.subset} | path : {path}")
 
     if args.save:
-        fig.canvas.draw()
-        data = np.frombuffer(fig.canvas.tostring_rgb(), dtype=np.uint8)
-        data = data.reshape(fig.canvas.get_width_height()[::-1] + (3,))
-        save_image(osp.join("./explorer.png"), data, create_dir=True)
+        saved_result_path = osp.join(args.project_dir, "explore_result")
+        if osp.exists(saved_result_path):
+            os.rmdir(saved_result_path)
```
JihwanEom: How about to use `shutil.rmtree` instead of `os.rmdir`? `os.rmdir` does not remove directory trees, it can only remove empty directory without any additional files. https://docs.python.org/3/library/os.html#os.rmdir

datumaro/cli/commands/explore.py
```diff
@@ -101,24 +110,21 @@ def explore_command(args):
...
+        os.makedirs(saved_result_path)
+        for result in results:
+            if not osp.exists(osp.join(saved_result_path, result.subset)):
+                os.makedirs(osp.join(saved_result_path, result.subset))
```
JihwanEom: How about to define a variable which has `osp.join(save_result_path, results.subset)`?

datumaro/components/explorer.py
```diff
@@ -43,22 +43,25 @@ def __init__(
...
-            for annotation in item.annotations:
-                if isinstance(annotation, HashKey):
-                    try:
-                        hash_key = annotation.hash_key[0]
-                        hash_key = np.unpackbits(hash_key, axis=-1)
-                        database_keys.append(hash_key)
-                        item_list.append(item)
-                    except Exception:
-                        hash_key = None
+        if not isinstance(dataset, List):
```
JihwanEom: How do you think about below suggestion if the dataset argument incomes only as `List[IDataset]` or `IDataset`.
```suggestion
        if isinstance(dataset, IDataset):
```

datumaro/components/explorer.py
```diff
@@ -43,22 +43,25 @@ def __init__(
...
+
+            for item in inference:
+                for annotation in item.annotations:
+                    if isinstance(annotation, HashKey):
+                        try:
+                            hash_key = annotation.hash_key[0]
+                            hash_key = np.unpackbits(hash_key, axis=-1)
+                            database_keys.append(hash_key)
+                            item_list.append(item)
+                        except Exception:
```
JihwanEom, line 70: Could you narrow the kind of `Exception`?

datumaro/components/explorer.py
```diff
@@ -43,22 +43,25 @@ def __init__(
...
-                        hash_key = np.unpackbits(hash_key, axis=-1)
-                        database_keys.append(hash_key)
-                        item_list.append(item)
-                    except Exception:
-                        hash_key = None
+        if not isinstance(dataset, List):
+            dataset = [dataset]
+
+        for dataset_ in dataset:
+            inference = dataset_.run_model(self._model, append_annotation=True)
```
JihwanEom: How about.. `inference_items` or `inference_item`?

datumaro/components/launcher.py
```diff
@@ -46,7 +49,13 @@ def __iter__(self):
                     continue
                 inputs.append(np.atleast_3d(item.media.data))
             inputs = np.array(inputs)
-            inference = self._launcher.launch(inputs)
+
+            if self._launcher.inferenced_check(batch):
```
JihwanEom: Could you explain more about `inferenced_check` function? It seems like always return `False`, when `if` statement(L53-L56) will be used?

sooahleex: I defined `inferenced_check` in `datumaro/plugins/explorer.py`. This function is related to explorer to check whether the item is inferenced before or not.

datumaro/components/explorer.py
sooahleex: I think I could not specify an Exception. I hope there is no case that `hash_key` is not computed, but I added this try-exception clause because there could be a case that part of the data did not include media.

datumaro/components/project.py
```diff
@@ -1079,7 +1092,9 @@ def add(self, paths, base=None):
             repo_root = osp.abspath(self._project_dir)
             assert is_subpath(base, base=repo_root), "Base path should be inside of the repo"
             base = osp.relpath(base, repo_root)
-            path_rewriter = lambda entry: osp.relpath(entry.path, base).replace("\\", "/")
+
+            def path_rewriter(entry):
+                return osp.relpath(entry.path, base).replace("\\", "/")
```
vinnamkim: Is this necessary?

datumaro/components/project.py
```diff
@@ -1584,6 +1599,10 @@ def find_project_dir(path: str) -> Optional[str]:
 
         return None
 
+    @staticmethod
+    def wrap_item(item, **kwargs):
+        return item.wrap(**kwargs)
+
```
vinnamkim: Is this necessary?

datumaro/components/explorer.py
```diff
@@ -43,22 +43,25 @@ def __init__(
...
-                        hash_key = annotation.hash_key[0]
-                        hash_key = np.unpackbits(hash_key, axis=-1)
-                        database_keys.append(hash_key)
-                        item_list.append(item)
-                    except Exception:
-                        hash_key = None
+        if isinstance(dataset, IDataset):
+            dataset = [dataset]
+
+        for dataset_ in dataset:
```
vinnamkim:
```suggestion
        for dataset in datasets:
```

datumaro/components/explorer.py
```diff
@@ -28,7 +28,7 @@ def calculate_hamming(B1, B2):
 class Explorer:
     def __init__(
         self,
-        dataset: IDataset,
+        dataset: Union[List[IDataset], IDataset],
         topk: int = 10,
```
vinnamkim:
```suggestion
        *datasets: Sequence[Dataset],
        *,
        topk: int = 10,
```

datumaro/plugins/data_formats/datumaro/exporter.py
vinnamkim, line 1: Please move saving hash key functionality to `components/exporter.py::Exporter`, so that it can be usable for all data formats. This saving function should store the following file with the following directory structure.

```console
dataset_directory/
     - hash_key_meta/
            - hash_keys.json # Hash key data (you can use binary file also)
            - index.bin # FAISS index (future)
     ...
```

datumaro/plugins/data_formats/datumaro/base.py
vinnamkim, line 1: Please make loading hash key functionality global (all data formats). I think that you need to implement `class DatasetExtractor` by extending `class DatasetBase` to take in `path` to get the hash key checkpoint:

```python
class DatasetExtractor(DatasetBase):
    def __init__(
        self,
        path: str,
        *,
        length: Optional[int] = None,
        subsets: Optional[Sequence[str]] = None,
        media_type: Type[MediaElement] = Image,
        ctx: Optional[ImportContext] = None,
    ):
        ...
        self._load_hash_key(path)
     
     def _load_hash_key(self, path):
        ...
```
Subsequently, make all data format plugins to inherit `DatasetExtractor`.

However, I think it can make this PR too huge. Please do it as a separate PR and implement `self._load_hash_key(path)` function to `DatasetExtractor` and `Datumaro` format in this PR.

datumaro/components/explorer.py
```diff
@@ -43,22 +43,25 @@ def __init__(
...
-                        hash_key = np.unpackbits(hash_key, axis=-1)
-                        database_keys.append(hash_key)
-                        item_list.append(item)
-                    except Exception:
-                        hash_key = None
+        if isinstance(dataset, IDataset):
+            dataset = [dataset]
+
+        for dataset_ in dataset:
+            inference_items = dataset_.run_model(self._model, append_annotation=True)
```
vinnamkim: You have to check all items have their hash key first before `run_model` to avoid loading the OpenVINO launcher. We don't need OpenVINO launcher if all hash keys are generated already.

datumaro/components/launcher.py
vinnamkim: Revert the changes in `ModelTransform` and give no hash key items to `ModelTransform` outside of this function. To pick up no hash key `DatasetItem` and group them to `IDataset`, you can use `DatasetItemStorage` and `DatasetItemStorageDatasetView`.
Please refer to
https://github.com/openvinotoolkit/datumaro/blob/12ddca41e91b8bb4adac5586b0c6b8acf1528170/datumaro/components/merge/base.py#L73-L79

datumaro/components/project.py
sooahleex: Yes, If this part is removed, `sources` are deleted during `checkout`.

datumaro/plugins/data_formats/datumaro/exporter.py
sooahleex: I will cover this in future PR.

datumaro/plugins/data_formats/datumaro/base.py
sooahleex: I will cover this in future PR, too.

datumaro/components/project.py
vinnamkim: Is this same with `path_rewriter = lambda entry: osp.relpath(entry.path, base).replace("\\", "/")`?

sooahleex: Yes, they worked the same way. I reverted this.

datumaro/components/explorer.py
```diff
@@ -43,22 +51,29 @@ def __init__(
...
-                        hash_key = annotation.hash_key[0]
-                        hash_key = np.unpackbits(hash_key, axis=-1)
-                        database_keys.append(hash_key)
-                        item_list.append(item)
-                    except Exception:
-                        hash_key = None
+        datasets = datasets[0]
+        if isinstance(datasets, IDataset):
+            datasets = [datasets]
+
```
vinnamkim:
```suggestion
```

datumaro/components/explorer.py
```diff
@@ -43,22 +51,29 @@ def __init__(
...
+            for item in dataset:
+                for annotation in item.annotations:
+                    if isinstance(annotation, HashKey):
+                        try:
+                            hash_key = annotation.hash_key[0]
+                            hash_key = np.unpackbits(hash_key, axis=-1)
+                            database_keys.append(hash_key)
+                            item_list.append(item)
+                        except Exception:
+                            hash_key = None
```
vinnamkim:
```suggestion
                            continue
```
Do we have to set `hash_key = None` explicitly?

datumaro/components/explorer.py
```diff
@@ -43,22 +51,29 @@ def __init__(
...
-                        database_keys.append(hash_key)
-                        item_list.append(item)
-                    except Exception:
-                        hash_key = None
+        datasets = datasets[0]
+        if isinstance(datasets, IDataset):
+            datasets = [datasets]
+
+        for dataset in datasets:
+            uninferenced_dataset = select_uninferenced_dataset(dataset)
```
vinnamkim, line 63:
```suggestion
    datasets_to_infer = [select_uninferenced_dataset(dataset) for dataset in datasets]
    datasets = self.compute_hash_key(datasets, datasets_to_infer)

def compute_hash_key(self, datasets, datasets_to_infer):
      for dataset in datasets_to_infer:
            if len(dataset) > 0:
                  dataset.run_model(self.model, append_annotation)

        for dataset, dataset_to_infer in zip(datasets, datasets_to_infer):
             dataset.update(dataset_to_infer)
        return datasets

@property
def model(self):
     if self._model is None:
           self._model = ExplorerLauncher(model_name="clip_visual_ViT-B_32")
     return self._model
```

Allow lazy loading for the model launcher when there are no model inference requirements.

datumaro/plugins/data_formats/datumaro/exporter.py
```diff
@@ -303,6 +306,11 @@ def _convert_cuboid_3d_object(self, obj):
     def _convert_ellipse_object(self, obj: Ellipse):
         return self._convert_shape_object(obj)
 
+    def _convert_hashkey_object(self, obj):
+        converted = self._convert_annotation(obj)
+        converted.update({"hashkey": cast(obj.hash_key, np.array)})
+        return converted
```
vinnamkim: You don't need this because `self._save_hashkey_file()` will save hash key to the separate file.

datumaro/plugins/data_formats/datumaro/base.py
```diff
@@ -317,6 +322,14 @@ def _load_annotations(item):
...
                 )
 
+            elif ann_type == AnnotationType.hash_key:
+                hash_key = np.asarray(hash_key, dtype=np.uint8)
+                loaded.append(
+                    HashKey(
+                        hash_key=np.array(hash_key),
+                    )
+                )
+
```
vinnamkim: You don't need this because `parse_hashkey_file()` will load hash_key from the separate file.

vinnamkim (changes requested):

datumaro/components/explorer.py
sooahleex: This part is needed to use first one of `datasets`. `datasets` is tuple and we need to use the first one.

JihwanEom (dismissed):

datumaro/components/explorer.py
vinnamkim: I don't understand what you mean. You give a sequence of `Dataset` to `Explore`.

https://github.com/openvinotoolkit/datumaro/blob/1b8c8f4399a00cc97720ce4d3b8abe103699ecac/datumaro/cli/commands/explore.py#L109-L119

But, why do we have to just use the first one `Dataset` of the sequence? Didn't you want to explore all datasets present in the sequence? If not, isn't it correct to only accept a single dataset as the input?

datumaro/components/config_model.py
```diff
@@ -92,6 +92,7 @@ def get_stage(self, stage):
     .add("build_targets", lambda: _DictConfig(lambda v=None: BuildTarget(v)))
     .add("base_dir", str, internal=True)
     .add("config_path", str, internal=True)
+    .add("hashkey", dict)
```
vinnamkim: Is it necessary?

datumaro/components/explorer.py
sooahleex: The possible input for `datasets` are `Union[List[IDataset], IDataset]`, so the first one of datasets as input in now is like `([dataset1, dataset2],)` or `(dataset1,)`. To avoid this, I need to convert `source_datasets` in line 112 to tuple which you attached above. If it is better, I will change it to use tuple instead of list.

vinnamkim: I think that you can just change L119 to `explorer = Explorer(*source_datasets)`.

datumaro/plugins/data_formats/datumaro/base.py
```diff
@@ -190,6 +199,10 @@ def _load_items(self, parsed):
                 media = MediaElement(path=media_desc.get("path"))
                 self._media_type = MediaElement
 
+            if has_hashkey_file(self._rootpath):
```
vinnamkim: You are trying to invoke disk I/O for every `DatasetItem` (O(N)). This will make it super slow.

datumaro/components/explorer.py
```diff
@@ -41,28 +49,50 @@ def __init__(
...
+        if self._text_model is None:
+            self._text_model = ExplorerLauncher(model_name="clip_text_ViT-B_32")
+        return self._text_model
+
+    def compute_hash_key(self, datasets, datasets_to_infer):
+        for dataset in datasets_to_infer:
+            if len(dataset) > 0:
+                dataset.run_model(self.model, append_annotation=True)
+            for dataset, dataset_to_infer in zip(datasets, datasets_to_infer):
+                dataset.update(dataset_to_infer)
```
vinnamkim:
```suggestion
        for dataset, dataset_to_infer in zip(datasets, datasets_to_infer):
            dataset.update(dataset_to_infer)
```

datumaro/plugins/data_formats/datumaro/base.py
vinnamkim:
```python
    def _load_impl(self, path: str) -> None:
        """Actual implementation of loading Datumaro format."""
        self._infos = self._load_infos(self._parsed_anns)
        self._categories = self._load_categories(self._parsed_anns)
        items = self._load_items(self._parsed_anns)
        self._items = self._load_hash_key(items)

    def _load_hash_key(self, items):
        if not has_hashkey_file(self._rootpath):
            return items

        hashkey_dict = parse_hashkey_file(self._rootpath)
        for item in items:
            item.annotations.append(hashkey_dict[(item.id, item.subset)])
        return items
```

datumaro/util/meta_file_util.py
```diff
@@ -72,3 +85,43 @@ def save_meta_file(path, categories):
...
+    if not hashkey_dict:
+        hashkey_dict = {}
+
+    for item in item_list:
+        item_id = item.id
+        for annotation in item.annotations:
+            if isinstance(annotation, HashKey):
+                hashkey = annotation.hash_key
+                break
+        hashkey_dict.update({item_id: hashkey.tolist()})
```
vinnamkim: I found you only use `item.id` for the key of `hashkey_dict`. This will make a bug when there are two items which have the same `id` but different `subset`s. Please fix this and add a test case which must fail in this commit but not failed in the fixed commit.

vinnamkim (changes requested):

datumaro/plugins/data_formats/datumaro/base.py
```diff
@@ -99,7 +103,9 @@ def _load_impl(self, path: str) -> None:
         """Actual implementation of loading Datumaro format."""
         self._infos = self._load_infos(self._parsed_anns)
         self._categories = self._load_categories(self._parsed_anns)
-        self._items = self._load_items(self._parsed_anns)
+        # self._items = self._load_items(self._parsed_anns)
```
vinnamkim:
```suggestion
```

datumaro/plugins/data_formats/datumaro/exporter.py
```diff
@@ -419,6 +422,11 @@ def apply(self, pool: Optional[Pool] = None, *args, **kwargs):
             subset = item.subset or DEFAULT_SUBSET_NAME
             writers[subset].add_item(item, pool)
 
+            for annotation in item.annotations:
+                if isinstance(annotation, HashKey):
+                    self._save_hashkey_meta = True
+                    break
```
vinnamkim:
```suggestion
            self._check_hash_key_existence(item)

       def _check_hash_key_existence(self, item):
            if self._save_hashkey_meta:
                 return
            for annotation in item.annotations:
                if isinstance(annotation, HashKey):
                    self._save_hashkey_meta = True
                    return
```

datumaro/plugins/explorer.py
```diff
@@ -63,7 +63,7 @@ def infer(self, inputs):
             inputs = self._tokenize(prompt_text)
             inputs = {self._input_blob: inputs}
         else:
-            if not inputs.any():
+            if inputs is None:
```
vinnamkim:
```suggestion
       elif isinstance(inputs, np.ndarray):
            if len(inputs.shape) == 3:
                inputs = np.expand_dims(inputs, axis=0)
            inputs = self.process_inputs(inputs)
       else:
            raise ValueError(f"inputs={inputs} is not allowed type.")
```

vinnamkim (approved):
LGTM.

JihwanEom (approved):

PR #983, opened 2023-05-02
Remedying apache-beam dependency conflicts

yunchu:
### Summary

Fixing the failure of regression tests on MacOS11-Python3.10.

Verified on this actions: https://github.com/openvinotoolkit/datumaro/actions/runs/4861593576

image

### DIFF
--- a/.github/workflows/nightly_check.yml
+++ b/.github/workflows/nightly_check.yml
@@ -11,8 +11,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ['macos-11']
-        python-version: ['3.10']
+        os: ['macos-11', 'ubuntu-20.04', 'windows-2019']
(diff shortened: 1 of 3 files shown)


wonjuleee (approved):

.github/workflows/nightly_check.yml
```diff
@@ -21,6 +21,7 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
+          cache: pip
```
vinnamkim, line 24: Can this option be shared everywhere?

vinnamkim (approved):
LGTM.

.github/workflows/nightly_check.yml
yunchu: Not 100% sure but I want to check that works with this workflow.

PR #987, opened 2023-05-04
Fix Synthia data format and Add Synthia AL and SF formats

wonjuleee:
### Checklist

### License

```python
#
```


datumaro/plugins/data_formats/synthia/importer.py
```diff
@@ -0,0 +1,51 @@
...
+    def find_sources(cls, path):
+        return [{"url": path, "format": cls.FORMAT}]
+
+
+class SynthiaRandImporter(_SynthiaImporter):
+    FORMAT = SynthiaFormatType.synthia_rand.name
+    META_FOLDERS = []
+    for prefix in vars(SynthiaRandPath).values():
+        if isinstance(prefix, str) and "datumaro" not in prefix:
+            META_FOLDERS.append(prefix)
```
vinnamkim:
```suggestion
    META_FOLDERS = SynthiaRandPath.meta_folders()
```
I recommend you wrap these lines up to a `classmethod` of `SynthiaRandPath` for better readability.

datumaro/plugins/data_formats/synthia/base.py
```diff
@@ -96,78 +72,89 @@ def parse_label_map(path):
...
-                color_mask = lazy_mask(gt_img, inverse_cls_colormap)
-                color_mask = color_mask()
+
+                seg_img = load_image(seg_img_path, dtype=np.uint16)
+                color_mask = np.zeros_like(seg_img[:, :, 0])
+                for i in range(seg_img.shape[0]):
+                    for j in range(seg_img.shape[1]):
+                        pixel_color = tuple(seg_img[i, j, :])
+                        if pixel_color in inverse_cls_colormap.keys():
+                            color_mask[i, j] = inverse_cls_colormap[pixel_color]
```
vinnamkim: Basically, I think it is not good idea to loop over (H,W) dimensions with Python because of performance issues. On the other hand, this code will dump all masks in memory (not lazy loading). The original code probably wasn't lazy loading as well (calling `color_mask()`). However, we can modify the previous logic to achieve fully lazy loading as follows.

```python
lazy_color_mask = lazy_mask(gt_img, inverse_cls_colormap)
lazy_color_mask_callable = lambda: lazy_color_mask() == label_id

...

for label_id in classes:
    anno.append(
          Mask(image=lazy_color_mask_callable, label=label_id)
    )
```

datumaro/plugins/data_formats/synthia/importer.py
wonjuleee: I updated at the [241b5548af138265e96f7c871eec0b522a225448](https://github.com/openvinotoolkit/datumaro/pull/987/commits/241b5548af138265e96f7c871eec0b522a225448)

datumaro/plugins/data_formats/synthia/base.py
wonjuleee: I updated at the [241b5548af138265e96f7c871eec0b522a225448](https://github.com/openvinotoolkit/datumaro/pull/987/commits/241b5548af138265e96f7c871eec0b522a225448)

wonjuleee: By the way, when using `load_mask` (or `lazy_mask`), we have some errors during importing Synthia dataset because of https://github.com/openvinotoolkit/datumaro/blob/98298beaae9f5805f1a51e6d213f0b5f74a7aacf/datumaro/util/mask_tools.py#L82. This is because, some pixels are not correctly matched into a given label map (pixel value to label id). In this case, we should map them into a predefined class (such as `background` or `ignored`). What do you think?

vinnamkim: Is there any pixel from the segmentation masks of Synthia which cannot be mapped with the table? I guess that Synthia leaves this `("Void", (0, 0, 0))` mapping for `background` or `ignored`. Isn't it our mistake that we can't find the mapping rule for those pixels?

datumaro/plugins/data_formats/synthia/format.py
```diff
@@ -0,0 +1,111 @@
...
+        ("Truck", (0, 80, 100)),
+        ("Bus", (0, 60, 100)),
+        ("Train", (0, 80, 100)),
+        ("Motorcycle", (0, 0, 230)),
+        ("Bicycle", (119, 11, 31)),
+        ("RoadLines", (157, 234, 50)),
+        ("Other", (72, 0, 98)),
+        ("RoadWorks", (167, 106, 29)),
+    ]
+)
```
wonjuleee, line 91: Here is the label map for Synthia-Rand, while there are some undeclared pixels in data as shown below.
image

datumaro/plugins/data_formats/synthia/base.py
wonjuleee: I left another comment to compare the predefined label map and undeclared pixels in the Synthia-SF dataset.
Please take a look at below.

datumaro/plugins/data_formats/synthia/format.py
wonjuleee: So my suggestion is map them into label 0 for a while. And we may need to specify the default class id into the mask tools,

vinnamkim: It is suspicious to happen when some files are corrupted due to network issues while downloading it. However, yes, we can introduce `default_id: Optional[int]` argument to `unpaint_mask()` function.
https://github.com/openvinotoolkit/datumaro/blob/98298beaae9f5805f1a51e6d213f0b5f74a7aacf/datumaro/util/mask_tools.py#L56-L87

if `default_id=None`, it will raise an error as is, otherwise it will give warning and assign `default_id`, but no error raise.

wonjuleee: good, I updated `unpaint_mask` and `load_mask` with `default_id`.

vinnamkim (approved):
LGTM.

datumaro/plugins/data_formats/synthia/base.py
```diff
@@ -4,52 +4,28 @@
...
+    SynthiaAlLabelMap,
+    SynthiaAlPath,
+    SynthiaRandLabelMap,
+    SynthiaRandPath,
+    SynthiaSfLabelMap,
+    SynthiaSfPath,
 )
 
 
 def make_categories(label_map=None):
```
cih9088, line 28:
```suggestion
def make_categories(label_map):
```

cih9088 (approved):

cih9088 (approved):

PR #992, opened 2023-05-08
Enrich stack trace while importing dataset to improve user experience

cih9088:
### Summary
- Sorted detection results to make it consistent if something went wrong
- Added a verbose error output with what and where Datumaro tried to import
- Changed to consume `ctx` for all data formats supported officially to lift up one level of stack trace
- Narrowed down exception type for verbosity

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/components/dataset.py
+++ b/datumaro/components/dataset.py
@@ -1249,10 +1249,12 @@ class Dataset(IDataset):
             dataset = cls.from_extractors(*extractors, env=env, merge_policy=merge_policy)
             if eager:
                 dataset.init_cache()
-        except (_ImportFail, DatasetImportError, NotADirectoryError, FileNotFoundError) as e:
-            cause = e.__cause__ if isinstance(e, _ImportFail) else e
+        except _ImportFail as e:
+            cause = e.__cause__ if getattr(e, "__cause__", None) is not None else e
             cause.__traceback__ = e.__traceback__
             raise DatasetImportError(f"Failed to import dataset '{format}' at '{path}'.") from cause
+        except Exception as e:
+            raise DatasetImportError(f"Failed to import dataset '{format}' at '{path}'.") from e
 
         dataset._source_path = path
         dataset._format = format
--- a/datumaro/components/dataset.py
+++ b/datumaro/components/dataset.py
@@ -1249,8 +1249,8 @@ class Dataset(IDataset):
             dataset = cls.from_extractors(*extractors, env=env, merge_policy=merge_policy)
             if eager:
                 dataset.init_cache()
-        except _ImportFail as e:
-            cause = e.__cause__
+        except (_ImportFail, DatasetImportError) as e:
+            cause = e.__cause__ if isinstance(e, _ImportFail) else e
             cause.__traceback__ = e.__traceback__
             raise DatasetImportError(f"Failed to import dataset '{format}' at '{path}'.") from cause
 
--- a/datumaro/components/dataset.py
+++ b/datumaro/components/dataset.py
@@ -32,6 +32,7 @@ from datumaro.components.environment import DEFAULT_ENVIRONMENT, Environment
 from datumaro.components.errors import (
     CategoriesRedefinedError,
     ConflictingCategoriesError,
+    DatasetImportError,
     DatasetInfosRedefinedError,
     MediaTypeError,
     MultipleFormatsMatchError,
@@ -1249,7 +1250,9 @@ class Dataset(IDataset):
             if eager:
                 dataset.init_cache()
         except _ImportFail as e:
-            raise e.__cause__
+            cause = e.__cause__
+            cause.__traceback__ = e.__traceback__
+            raise DatasetImportError(f"Failed to import dataset '{format}' at '{path}'.") from cause
 
         dataset._source_path = path
         dataset._format = format
--- a/datumaro/components/dataset.py
+++ b/datumaro/components/dataset.py
@@ -1249,7 +1249,7 @@ class Dataset(IDataset):
             dataset = cls.from_extractors(*extractors, env=env, merge_policy=merge_policy)
             if eager:
                 dataset.init_cache()
-        except (_ImportFail, DatasetImportError) as e:
+        except (_ImportFail, DatasetImportError, NotADirectoryError, FileNotFoundError) as e:
             cause = e.__cause__ if isinstance(e, _ImportFail) else e
             cause.__traceback__ = e.__traceback__
             raise DatasetImportError(f"Failed to import dataset '{format}' at '{path}'.") from cause
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Enhancements
 - Use autosummary for fully-automatic Python module docs generation
   (<https://github.com/openvinotoolkit/datumaro/pull/973>)
+- Enrich stack trace for better user experience when importing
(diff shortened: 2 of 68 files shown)


datumaro/components/dataset.py
```diff
@@ -1248,8 +1249,10 @@ def import_from(
             dataset = cls.from_extractors(*extractors, env=env, merge_policy=merge_policy)
             if eager:
                 dataset.init_cache()
-        except _ImportFail as e:
-            raise e.__cause__
+        except (_ImportFail, DatasetImportError, NotADirectoryError, FileNotFoundError) as e:
+            cause = e.__cause__ if isinstance(e, _ImportFail) else e
+            cause.__traceback__ = e.__traceback__
+            raise DatasetImportError(f"Failed to import dataset '{format}' at '{path}'.") from cause
```
vinnamkim: Is it better to catch every exception type here rather than specifying some types (L1252)? and why do we need to pull the stack level up one step in L1253-1254? My point is that it can be enough just to make exception chaining on those lines as follows.

```python
except Exception as cause:
    raise DatasetImportError(f"Failed to import dataset '{format}' at '{path}'.") from cause
```

cih9088: We can catch all exceptions like you suggested but I wanted to make it extra clear that Datumaro cares about following exceptions only and other exceptions not caught should be handled by each data format carefully to give informative exception message to users by explicitly rasing a `DatasetImportError`-ish exception.

`_ImportFail` raised simply like this: `raise _ImportFail from exception`. Without it, it shows uninformative additional stack trace.

Current behaviour:
```
Traceback (most recent call last):
  File "/home/andy/workspace/datumaro/datumaro/components/dataset.py", line 1229, in import_from
    env.make_extractor(src_conf.format, src_conf.url, **extractor_kwargs)
  File "/home/andy/workspace/datumaro/datumaro/components/environment.py", line 244, in make_extractor
    return self.extractors.get(name)(*args, **kwargs)
  File "/home/andy/workspace/datumaro/datumaro/plugins/data_formats/coco/base.py", line 531, in __init__
    super().__init__(path, **kwargs)
  File "/home/andy/workspace/datumaro/datumaro/plugins/data_formats/coco/base.py", line 125, in __init__
    self._ctx.error_policy.fail(DatasetImportError("a"))
  File "/home/andy/workspace/datumaro/datumaro/components/importer.py", line 72, in fail
    raise _ImportFail from error
datumaro.components.errors.DatasetImportError: a

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "test2.py", line 20, in <module>
    dataset = Dataset.import_from(
  File "/home/andy/workspace/datumaro/datumaro/components/dataset.py", line 1257, in import_from
    raise DatasetImportError(f"Failed to import dataset '{format}' at '{path}'.") from cause
datumaro.components.errors.DatasetImportError: Failed to import dataset 'coco' at '/mnt/data/dataset/coco_minitrain_25k'.
```

Without it:
```
datumaro.components.errors.DatasetImportError: a

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/andy/workspace/datumaro/datumaro/components/dataset.py", line 1229, in import_from
    env.make_extractor(src_conf.format, src_conf.url, **extractor_kwargs)
  File "/home/andy/workspace/datumaro/datumaro/components/environment.py", line 244, in make_extractor
    return self.extractors.get(name)(*args, **kwargs)
  File "/home/andy/workspace/datumaro/datumaro/plugins/data_formats/coco/base.py", line 531, in __init__
    super().__init__(path, **kwargs)
  File "/home/andy/workspace/datumaro/datumaro/plugins/data_formats/coco/base.py", line 125, in __init__
    self._ctx.error_policy.fail(DatasetImportError("a"))
  File "/home/andy/workspace/datumaro/datumaro/components/importer.py", line 72, in fail
    raise _ImportFail from error
datumaro.components.importer._ImportFail

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "test2.py", line 20, in <module>
    dataset = Dataset.import_from(
  File "/home/andy/workspace/datumaro/datumaro/components/dataset.py", line 1257, in import_from
    raise DatasetImportError(f"Failed to import dataset '{format}' at '{path}'.") from cause
datumaro.components.errors.DatasetImportError: Failed to import dataset 'coco' at '/mnt/data/dataset/coco_minitrain_25k'.
```

vinnamkim: I definitely agree that L1252 should raise `DatasetImportError()`. However, to make it clear that Datumaro fails to import the dataset, all exceptions should be catched in L1252. This is because there is a possibility that some exception can bypass L1252 and users will not know it this invokes the dataset import error.

Secondly, is *Without it:* the one who doesn't have L1253-1254? It looks good to me since 
```console
datumaro.components.errors.DatasetImportError: a <- root cause

The above exception was the direct cause of the following exception: <- chain 1
...
The above exception was the direct cause of the following exception: <- chain 2
...
```

cih9088: 1. The exceptions not caught by L1252 should give taceback which users **can** know something went from importing process. It just does not give additional information saying `"Failed to import dataset '{format}' at '{path}'.` I have no strong objection to catch all exceptions though.
2. I was not saying this is wrong but we can remove uninformative stack `datumaro.components.importer._ImportFail` like original implementation did.

vinnamkim: 1. I meant that they cannot easily know **something went wrong from importing process** if we don't chain `DatasetImportError`. Without this chaining, users should look at L1212-1246 and learn those lines are about to import the dataset.
2. If you want to remove uninformative stacks from the following
```python
  File "/home/andy/workspace/datumaro/datumaro/plugins/data_formats/coco/base.py", line 125, in __init__
    self._ctx.error_policy.fail(DatasetImportError("a"))
  File "/home/andy/workspace/datumaro/datumaro/components/importer.py", line 72, in fail
    raise _ImportFail from error
```
, I think that it would be good to raise a special exception type from `datumaro/components/importer.py` and catch it from L1252 to remove the uninformative stacks. For example,

```python
#  File "/home/andy/workspace/datumaro/datumaro/components/importer.py", line 72, in fail
    raise _ImporterContextFail from error

# L1252
        except _ImporterContextFail as e:
            cause = e.__cause__ if isinstance(e, _ImporterContextFail) else e
            cause.__traceback__ = e.__traceback__
            raise DatasetImportError(f"Failed to import dataset '{format}' at '{path}'.") from cause
        except Exception as e:
            raise DatasetImportError(f"Failed to import dataset '{format}' at '{path}'.") from e
```

cih9088: This one looks much better. Thanks for the suggestion. Fixed to catch all exceptions in https://github.com/openvinotoolkit/datumaro/pull/992/commits/a39b16d3eb98caf76e30a7f99cc90d06977d8842

vinnamkim (approved):
LGTM.

datumaro/components/dataset.py
```diff
@@ -1249,7 +1250,11 @@ def import_from(
             if eager:
                 dataset.init_cache()
         except _ImportFail as e:
-            raise e.__cause__
+            cause = e.__cause__ if getattr(e, "__cause__", None) is not None else e
```
bonhunko, line 1253: I recommend to use `hasattr`

bonhunko (approved):
LGTM! I left a minor comment.

datumaro/components/dataset.py
cih9088: Thanks for the comment but `hasattr` would return `True` always since `__cause__` attribute is set to another `Exception` if the exception was raised from it, otherwise `None`.

PR #998, opened 2023-05-17
update code-scan workflow

yunchu:
### Summary

Updated code-scan workflow to enable vulnerability scan using the Trivy

### DIFF
--- a/.github/workflows/code_scan.yml
+++ b/.github/workflows/code_scan.yml
@@ -7,6 +7,28 @@ on:
     - cron: "0 18 * * 1-5"
 
 jobs:
+  Trivy:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
+      - name: Install dependencies
+        run: python -m pip install tox
+      - name: Trivy Scanning
+        env:
+          TRIVY_DOWNLOAD_URL: ${{ vars.TRIVY_DOWNLOAD_URL }}
+        run: tox -e trivy-scan
+      - name: Upload Trivy results artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: trivy-results
+          path: |
+            .tox/trivy-scan-results.txt
+            .tox/trivy-spdx-datumaro.json
   Bandit:
     runs-on: ubuntu-latest
     steps:
@@ -28,7 +50,7 @@ jobs:
         # Use always() to always run this step to publish scan results when there are test failures
         if: ${{ always() }}
   call-notify-to-teams:
-    needs: [Bandit]
+    needs: [Trivy, Bandit]
     if: |
       always() &&
       contains(needs.*.result, 'failure')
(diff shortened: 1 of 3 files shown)


.github/workflows/code_scan.yml
```diff
@@ -7,6 +7,28 @@ on:
...
+        uses: actions/checkout@v3
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
+      - name: Install dependencies
+        run: python -m pip install tox
+      - name: Trivy Scanning
+        env:
+          TRIVY_DOWNLOAD_URL: ${{ vars.TRIVY_DOWNLOAD_URL }}
```
vinnamkim, line 23: Should this URL be retrieved from the variable? Why not just be specified here? I guess that URL has a version number, and it is the reason to write it to the source file (`.yml`).

yunchu: @vinnamkim because I don't want to update worflow file itself when there is a new release of the trivy tool. we can just update the download URL on the variables.

vinnamkim (dismissed):
LGTM.

.github/workflows/code_scan.yml
```diff
@@ -7,6 +7,28 @@ on:
...
+        run: tox -e trivy-scan
+      - name: Upload Trivy results artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: trivy-results
+          path: |
+            .tox/trivy-scan-results.txt
+            .tox/trivy-spdx-otx.json
```
wonjuleee: This might be `trivy-spdx-datumaro.json`

tox.ini
```diff
@@ -34,27 +34,25 @@ commands =
...
-    {toxworkdir}/snyk-to-html-linux -i {toxworkdir}/snyk.json -o {toxworkdir}/snyk.html -d
-    rm {toxworkdir}/snyk
-    rm {toxworkdir}/snyk-to-html-linux
-    rm snyk-req.txt
+    bash -c "pip freeze > {toxworkdir}/requirements.txt"
+    cat {toxworkdir}/requirements.txt
+    curl -L0 {env:TRIVY_DOWNLOAD_URL} -o {toxworkdir}/trivy.tar.gz
+    tar -xzf {toxworkdir}/trivy.tar.gz -C {toxworkdir}
+    {toxworkdir}/trivy fs -d -c {toxinidir}/trivy.yaml --dependency-tree -o {toxworkdir}/trivy-scan-results.txt {toxworkdir}/requirements.txt
+    {toxworkdir}/trivy fs -d -c {toxinidir}/trivy.yaml --format spdx-json -o {toxworkdir}/trivy-spdx-otx.json {toxworkdir}/requirements.txt
```
wonjuleee: here too.

.github/workflows/code_scan.yml
```diff
@@ -7,6 +7,28 @@ on:
...
 jobs:
+  Trivy:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
```
wonjuleee, line 18: Is there any reason to set python version to `3.10` for trivy scanning?

yunchu: No, the python is required for installing datumaro package & its dependency, not for the Trivy. so any versions are okay if datumaro is supported.

vinnamkim (approved):

wonjuleee (approved):

PR #1001, opened 2023-05-18
Drop a malformed transform from StackedTransform automatically

vinnamkim:
### Summary

- Ticket no. 106054
- A malformed transform can be stacked on `_StackTransform`. Therefore, it would fail if the user tries to apply several transforms to the dataset sequentially and there exist malformed transforms. Also, once a malformed transform is given to `Dataset.transform()`, that transform permanently remains in the dataset, so creating a new dataset is the only way to solve it.
- This patch pops up the malformed transform from the transform stack automatically if it raises an error at construction.

### How to test
Unit tests are added for this change.

### Checklist

### License

```python
#
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   (<https://github.com/openvinotoolkit/datumaro/pull/987>)
 - Fix `person_layout` categories and `action_classification` attributes in imported Pascal-VOC dataset
   (<https://github.com/openvinotoolkit/datumaro/pull/997>)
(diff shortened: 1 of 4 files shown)


sooahleex (approved):
Looks good to me

bonhunko (approved):
LGTM

wonjuleee (approved):
It looks good to me for avoiding an error, but it may require to remove the root cause of an error when stacking transforms. Is there any way to resolve this?

wonjuleee (approved):
It looks good to me for avoiding an error, but it may require to remove the root cause of an error when stacking transforms. Is there any way to resolve this?

cih9088 (approved):

PR #1004, opened 2023-05-18
Improve RemoveAnnotations to cope with annotation ids

wonjuleee:
### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/plugins/transforms.py
+++ b/datumaro/plugins/transforms.py
@@ -9,7 +9,7 @@ import logging as log
 import os.path as osp
 import random
 import re
-from collections import defaultdict, Counter
+from collections import Counter, defaultdict
 from copy import deepcopy
 from enum import Enum, auto
 from itertools import chain
--- a/datumaro/plugins/transforms.py
+++ b/datumaro/plugins/transforms.py
@@ -1121,6 +1121,8 @@ class RemoveAnnotations(ItemTransform):
     @staticmethod
     def _parse_id(s):
         full_id = s.split(":")
+        if len(full_id) == 3:
+            full_id[-1] = int(full_id[-1])
         if len(full_id) != 2 or len(full_id) != 3:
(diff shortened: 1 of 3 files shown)


cih9088 (dismissed):

datumaro/plugins/transforms.py
```diff
@@ -1136,18 +1138,37 @@ def build_cmdline_parser(cls, **kwargs):
...
+        for v in ids:
+            key = tuple(v[:2])
+            val = v[2] if len(v) == 3 else None
+            if key in self._ids:
+                if val and self._ids[key]:
+                    self._ids[key].append(val)
+                else:
+                    self._ids[key] = None
+            else:
+                self._ids[key] = None if val is None else [val]
```
vinnamkim:
```suggestion
        self._ids = defaultdict(list)
        for v in ids:
            key = tuple(v[:2])
            val = v[2] if len(v) == 3 else None
            if val is not None:
                self._ids[key].append(val)
            else:
                self._ids[key] = []
```

datumaro/plugins/transforms.py
```diff
@@ -1121,9 +1121,11 @@ class RemoveAnnotations(ItemTransform):
     @staticmethod
     def _parse_id(s):
         full_id = s.split(":")
```
vinnamkim, line 1123:
```suggestion
        full_id = s.split(":")
        if len(full_id) == 3:
            full_id[-1] = int(full_id[-1])
```

`s` will be string in CLI arguments, so that we need to apply type conversion. It would be better to add a CLI integration test for this.

bonhunko (dismissed):

vinnamkim (dismissed):
LGTM.

cih9088 (dismissed):

cih9088 (approved):

vinnamkim (approved):
LGTM.

PR #1031, opened 2023-05-30
Enhance import performance for built-in plugins

vinnamkim:
### Summary

- Ticket no. 112204
- Apply lazy import for built-in plugins (`datumaro/components/lazy_plugin.py`, `datumaro/plugins/specs.py`)
- Use `DEFAULT_ENVIRONMENT` rather than creating `Environment()` for CLI codes.
- Add `openvino-dev` to default extra deps and update `contributing.md`
- Decrease `datum` CLI latency by 1 sec (5 -> 4).

(before)
```console
(datumaro) vinnamki@vinnamki:~/datumaro$ time datum -h
usage: datum [-h] [--version] [--loglevel LOGLEVEL]
...

real    0m5.052s
user    0m5.208s
sys     0m3.363s
```

(after)
```console
(datumaro) vinnamki@vinnamki:~/datumaro$ time datum -h
usage: datum [-h] [--version] [--loglevel LOGLEVEL]
...

real    0m3.955s
user    0m4.323s
sys     0m3.172s
```

### How to test
I added a test for this change.

### Checklist

### License

```python
#
```


datumaro/components/environment.py
```diff
@@ -322,5 +354,11 @@ def _register(registry: PluginRegistry):
 
         return merged
 
+    @classmethod
+    @contextmanager
+    def release_builtin_plugins(cls):
```
cih9088, line 357: This method seems no use to users. Why do we have to implement it in `Environment` rather than in test file? Would you give some example use-cases?

vinnamkim: I wanted to open it for users who wants to reload built-in plugins somehow.

cih9088 (dismissed):

vinnamkim:
https://github.com/openvinotoolkit/datumaro/pull/1031/commits/d8f1fd0055fead402d7dde7a5000fa9271b165cb added `extra_deps` mechanism to the lazy plugin import. It requires for the system who doesn't have extra deps for some plugins. Without this, that system terminates with `ImportError` when trying to import a lazy plugin.

wonjuleee (approved):

cih9088 (approved):

PR #1034, opened 2023-06-01
Update workflows for testing to use tox

yunchu:
### Summary
This PR is for resolving some issues that could be introduced while run the tests with the package installation instead of 'editable'.

* changed flat layout to src layout - this will resolve `ModuleNotFoundError` for `datumaro._capi` when run the tests with package installation. some background information is here: https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/
* added testenvs for testing in tox - tox provides the easiest way to run tests with package installation option

### Updates were tested & verified with some workflows below manually
* Build (develop health check): https://github.com/openvinotoolkit/datumaro/actions/runs/5151541593
* Nightly regression: https://github.com/openvinotoolkit/datumaro/actions/runs/5151029735
* Weekly: https://github.com/openvinotoolkit/datumaro/actions/runs/5152396591
* code-scanning: https://github.com/openvinotoolkit/datumaro/actions/runs/5151897614
* Build-docs: locally


tox.ini
```diff
@@ -4,8 +4,18 @@ skip_missing_interpreters = true
 
 
 [testenv]
+passenv =
+    USERNAME
+    LOCALAPPDATA
+setenv =
+    MPLBACKEND=agg
 deps =
     -r{toxinidir}/requirements.txt
```
vinnamkim:
```suggestion
    -r{toxinidir}/requirements.txt
```

Is it possible to remove this?

pyproject.toml
```diff
@@ -40,6 +40,12 @@ exclude_lines = [
     'if __name__ == .__main__.:',
 ]
 
+[tool.pytest.ini_options]
+markers = [
+    "v1_3_0: marks tests for release 1.3.0",
+    "v1_4_0: marks tests for release 1.4.0",
```
vinnamkim, line 46: What is the meaning of these new markers?

tox.ini
yunchu: oh... you mean that the dependency in the requirements.txt should be installed in the package installation phase, right? let me check that again. thanks.

vinnamkim: Yes, for example, `pip install datumaro` doesn't require `pip install -r requirements.txt` explicitly.

pyproject.toml
yunchu: we can collect and run TCs to generate some validation report for the next release by using the marker to the newly implemented test cases for the new features of the release.

wonjuleee (commented):
Should we adhere to have `datumaro` directory? How about replacing `datumaro` to `src` isntead?

tox.ini
yunchu: updated.

yunchu:
> Should we adhere to have `datumaro` directory? How about replacing `datumaro` to `src` isntead?

we cannot omit `datumaro` directory because the first level of namespace for the package should be that.

vinnamkim (approved):
LGTM.

cih9088 (approved):

PR #1055, opened 2023-06-21
Refactor Launcher and ModelInterpreter

vinnamkim:
### Summary

- This is a pre-work of implementing `OVMSLauncher`.
- Did extensive refactoring on `Launcher`. Made an effort to add typing, docstrings, and enhance model pre & post-processing conducted by `Launcher` more comprehensible to users who want to implement their own `ModelInterpreter` for their model.
- Fixed hamming distance computation of `Explore`.

### How to test
- Fix the existing change according to the refactoring.

### Checklist

### License

```python
#
```


src/datumaro/components/abstracts/model_interpreter.py
```diff
@@ -3,23 +3,49 @@
...
         raise NotImplementedError("Function should be implemented.")
 
     @abstractmethod
-    def normalize(self, inputs):
+    def postprocess(self, pred: ModelPred, info: PrepInfo) -> List[Annotation]:
+        """Postprocess a model prediction.
+
+        Parameters:
+            pred: Model prediction
+            info: Preprocessing information
```
wonjuleee: Typo: `Postprocessing`

src/datumaro/components/algorithms/rise.py
```diff
@@ -94,7 +94,8 @@ def apply(self, image, progressive=False):
             samples = min(self.max_samples, samples)
         batch_size = self.batch_size
 
-        result = next(iter(model.launch(_expand(image, 0))))
+        pred = next(iter(model.infer(_expand(image, 0))))
```
wonjuleee, line 97: `_expand(image,0)` could be declared as preprocessing. This may be useful to consider the input normalization according to the target model interpreter.

src/datumaro/components/launcher.py
```diff
@@ -2,104 +2,121 @@
...
+        inputs_info = []
+
+        for item in batch:
+            prep_img, prep_info = self.preprocess(item.media)
+            inputs_img.append(prep_img)
+            inputs_info.append(prep_info)
+
+        preds = self.infer(
+            np.stack(
+                [np.atleast_3d(img) for img in inputs_img],
```
wonjuleee: `preprocess` has already defined to return batch of BGR formatted image. Should `np.atleast_3d` require?

src/datumaro/components/abstracts/model_interpreter.py
vinnamkim: It is not type. I added more descriptions for this in https://github.com/openvinotoolkit/datumaro/pull/1055/commits/429e2b13da67a8b1a48276749c3e1dd0077d81b1

src/datumaro/components/launcher.py
vinnamkim: Removed in https://github.com/openvinotoolkit/datumaro/pull/1055/commits/429e2b13da67a8b1a48276749c3e1dd0077d81b1

src/datumaro/components/algorithms/rise.py
vinnamkim: Since the RISE code is already very outdated and would require significant modifications to get it right, I only made minimal changes following this refactoring.

wonjuleee (approved):

src/datumaro/components/annotation.py
```diff
@@ -215,7 +216,32 @@ class Label(Annotation):
 @attrs(slots=True, eq=False, order=False)
 class HashKey(Annotation):
     _type = AnnotationType.hash_key
-    hash_key: np.ndarray = field(factory=lambda: np.zeros((1, 64), dtype=np.uint8))
+    hash_key: np.ndarray = field(validator=attr.validators.instance_of(np.ndarray))
+
+    @hash_key.validator
```
sooahleex, line 221: Is there any reason to use validator for `HashKey`?

src/datumaro/components/explorer.py
```diff
@@ -20,9 +20,7 @@ def calculate_hamming(B1, B2):
     :param B2:  vector [r*n]
     :return: hamming distance [r]
     """
-    q = B2.shape[1]  # max inner product value
-    distH = 0.5 * (q - B1 @ B2.transpose())
-    return distH
+    return np.count_nonzero(B1 != B2, axis=1)
```
sooahleex, line 23: Is there any reason to replace this? I think this could be take more time for bigger matrix.

src/datumaro/components/annotation.py
vinnamkim: I think that we don't need default factory. What we need to a validator which checks its input type, dim, dtype, and shape.

src/datumaro/components/explorer.py
vinnamkim: First of all, the previous implement was not computing the hamming distance. It was actually dot product between two vectors in {0,1}^d. Secondly, both (the dot product and this implementation) have the same complexity.

src/datumaro/components/annotation.py
vinnamkim: Every `Explorer` logic requires `HashKey.hash_key` to be 1D numpy array with `dtype=np.uint8` and `shape[0]=64`, so that I think that it is better to check the input data when creating `HashKey`.

sooahleex (approved):

PR #1079, opened 2023-07-06
Bugfix for explore cli test

sooahleex:
### Summary

- Fix explore cli fail tests : https://github.com/openvinotoolkit/datumaro/actions/runs/5464396553/jobs/9960479229
- Due to `train_path` not matched with `dataset._source_path`, `query` is considered as `string` input, not coverted to `Datasetitem` through `dataset.get_datasetitem_by_path(args.query)`.
- Match `train_path` of CLI path as `project/source-1/images/train/1.jpg'.

### Checklist

### License

```python
#
```

### DIFF
--- a/src/datumaro/cli/commands/explore.py
+++ b/src/datumaro/cli/commands/explore.py
@@ -133,7 +133,7 @@ def explore_command(args):
                 query_datasetitem = dataset.get_datasetitem_by_path(args.query)
             except Exception:
                 continue
-            if not query_datasetitem:
+            if query_datasetitem:
                 break
     else:
         query_datasetitem = args.query
--- a/tests/integration/cli/test_explore.py
+++ b/tests/integration/cli/test_explore.py
@@ -1,7 +1,7 @@
 import os.path as osp
 import platform
 from glob import glob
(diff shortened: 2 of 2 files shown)


vinnamkim (commented):
What is the cause of the problem?

sooahleex:
> What is the cause of the problem?

I set same images for `train` and `test` subset, and tried to find similar image of `train/1.jpg`. Then, the result could be `train/1.jpg` and `train/2.jpg` for `topk=2`, not only `train/1.jpg`. So I set to check whether `train` folder is included in `result`.

vinnamkim:
> > What is the cause of the problem?
> 
> I set same images for `train` and `test` subset, and tried to find similar image of `train/1.jpg`. Then, the result could be `train/1.jpg` and `train/2.jpg` for `topk=2`, not only `train/1.jpg`. So I set to check whether `train` folder is included in `result`.

I guess that you designed the test which should produce `1.jpg` for similarity search of the query `1.jpg`. But this PR is like "There happened a problem in 'A'. Ok, let's just remove 'A' and finish the case. Everybody is happy.".

PR #1086, opened 2023-07-10
Fix UnionMerge

wonjuleee:
### Summary

When importing `cable` category in MVTec anomaly data, the current implementation shows `KeyError`.
Because `self._matching_table` only stores the label mapping with different indices and some labels should keep the original indices, the codes raises the `KeyError` for keeping labels.

### Checklist

### License

```python
#
```


vinnamkim (commented):
Would you describe why this fix is needed and add a test case to reproduce the bug?

wonjuleee:
> Would you describe why this fix is needed and add a test case to reproduce the bug?

I have added the test taking care of mixed categories between datasets (same to MVTec dataset).

vinnamkim (approved):
LGTM.

PR #1092, opened 2023-07-12
Bugfix for explore cli test to get datasetitem through path

sooahleex:
### Summary

- Due to `train_path` not matched with `dataset._source_path`, `query` is considered as `string` input, not coverted to `Datasetitem` or could not bring the proper `Datasetitem` which is matched with `path` through `dataset.get_datasetitem_by_path(args.query)`. 
- Match `train_path` of CLI path as `project/source-1/images/train/1.jpg'.

### How to test

- I fixed the path in integration test.

### Checklist

### License

```python
#
```

### DIFF
--- a/src/datumaro/cli/commands/explore.py
+++ b/src/datumaro/cli/commands/explore.py
@@ -10,7 +10,10 @@ import random
 import shutil
 
 from datumaro.components.algorithms.hash_key_inference.explorer import Explorer
-from datumaro.components.algorithms.hash_key_inference.hashkey_util import match_query_subset
+from datumaro.components.algorithms.hash_key_inference.hashkey_util import (
+    check_and_convert_to_list,
+    match_query_subset,
+)
 from datumaro.components.dataset_base import DatasetItem
 from datumaro.components.errors import ProjectNotFoundError
 from datumaro.components.media import Image
@@ -58,6 +61,12 @@ def build_parser(parser_ctor=argparse.ArgumentParser):
         default=None,
         type=str,
         help="Datasetitem id of query to explore similar data",
+    )
+    query_parser.add_argument(
+        "--query-item-subset",
+        default=None,
+        type=str,
+        help="Datasetitem subset of query to explore similar data",
     )
     query_parser.add_argument(
         "--query-str",
@@ -141,11 +150,7 @@ def explore_command(args):
         project.working_tree.save()
 
     if args.query_img_path:
-        querys = (
-            [args.query_img_path]
-            if not isinstance(args.query_img_path, list)
-            else args.query_img_path
-        )
+        querys = check_and_convert_to_list(args.query_img_path)
         query_datasetitems = []
         for query_ in querys:
             query_datasetitem = DatasetItem(
@@ -159,12 +164,12 @@ def explore_command(args):
         query_datasetitems = []
         for query in querys:
             for dataset in source_datasets:
-                query_datasetitem = match_query_subset(query, dataset)
+                query_datasetitem = match_query_subset(
+                    query, dataset, subset=args.query_item_subset
+                )
             query_datasetitems.append(query_datasetitem)
     elif args.query_str:
         query_datasetitems = args.query_str
-    else:
-        raise
 
     results = explorer.explore_topk(query_datasetitems, args.topk)
 
--- a/src/datumaro/cli/commands/explore.py
+++ b/src/datumaro/cli/commands/explore.py
@@ -133,7 +133,7 @@ def explore_command(args):
                 query_datasetitem = dataset.get_datasetitem_by_path(args.query)
             except Exception:
                 continue
-            if not query_datasetitem:
+            if query_datasetitem:
                 break
     else:
         query_datasetitem = args.query
--- a/src/datumaro/cli/commands/explore.py
+++ b/src/datumaro/cli/commands/explore.py
@@ -6,14 +6,14 @@ import argparse
 import logging as log
 import os
 import os.path as osp
+import random
 import shutil
 
 from datumaro.components.algorithms.hash_key_inference.explorer import Explorer
-from datumaro.components.algorithms.hash_key_inference.hashkey_util import (
-    match_query_path,
-    match_query_subset,
-)
+from datumaro.components.algorithms.hash_key_inference.hashkey_util import match_query_subset
+from datumaro.components.dataset_base import DatasetItem
 from datumaro.components.errors import ProjectNotFoundError
+from datumaro.components.media import Image
 from datumaro.util import str_to_bool
 from datumaro.util.scope import scope_add, scoped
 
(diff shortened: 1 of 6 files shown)


vinnamkim (changes requested):
I recommend you to break the query explicitly into two cases:

1. DatasetItem id (`--query-item-id`)
2. Arbitrary image file path (not constrained to the project or dataset) (`--query-img-path`)

src/datumaro/cli/commands/explore.py
```diff
@@ -42,12 +46,24 @@ def build_parser(parser_ctor=argparse.ArgumentParser):
...
+        "--query-img-path",
+        default=None,
+        type=str,
+        help="",
+    )
+    query_parser.add_argument(
+        "--query-str",
+        default=None,
+        type=str,
+        help="",
```
vinnamkim: Please introduce https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_mutually_exclusive_group.
It seems that we need only one of them.

src/datumaro/components/algorithms/hash_key_inference/hashkey_util.py
```diff
@@ -172,3 +174,30 @@ def calculate_hamming(B1, B2):
...
+    result_path = None
+
+    for item in dataset:
+        query_rel = os.path.relpath(query_path, common_base)
+        item_rel = os.path.relpath(item.media.path, common_base)
+
+        for q, i in zip(query_rel.split(os.sep)[::-1], item_rel.split(os.sep)[::-1]):
+            if q == i:
+                result_path = item.media.path
+    return result_path
```
vinnamkim: Create `DatasetItem` from the given image path (we do not care whether it is in the dataset or not) and `hash_key` as well. Then, use the `hash_key` as the query.

src/datumaro/cli/commands/explore.py
```diff
@@ -125,20 +140,33 @@ def explore_command(args):
...
+                query = match_query_path(query_, dataset)
+                try:
+                    query_datasetitem = dataset.get_datasetitem_by_path(query)
+                except Exception:
+                    continue
+                if query_datasetitem:
+                    break
+            query_datasetitems.append(query_datasetitem)
+    elif args.query_item_id:
+        querys = [args.query_item_id]
```
vinnamkim: Why do we have to wrap it up with `list` mandatorily?

src/datumaro/components/algorithms/hash_key_inference/hashkey_util.py
sooahleex: I wonder if we use `hash_key` as query, then we need to update `explore_topk` function to get `hash_key` as input and we already find `hash_key` in `Datasetitem` through `explore_topk`. So in my thought, in this part, we only find the proper `DatatsetItem` whether it is in the project or not. If my understanding is wrong, please let me know.

vinnamkim: Is there any problem to create a new `Datasetitem` from the given image path and pass it to `explore_topk()`? I mean that the `DatasetItem` which is passed to this function should always come from the dataset where the `Explorer` refer?
https://github.com/openvinotoolkit/datumaro/blob/f100bb848f723d13acf09947042f46623450cede/src/datumaro/components/algorithms/hash_key_inference/explorer.py#L64-L68

sooahleex: I see. I will test this and update about it.

src/datumaro/components/algorithms/hash_key_inference/hashkey_util.py
```diff
@@ -172,3 +172,15 @@ def calculate_hamming(B1, B2):
...
+def match_query_subset(query_id, dataset):
+    subset_names = list(dataset.subsets().keys())
+    for subset_name in subset_names:
+        try:
+            query_datasetitem = dataset.get(query_id, subset_name)
+        except Exception:
+            continue
+        if query_datasetitem:
+            break
+    return query_datasetitem
```
vinnamkim: Please add `subset = None` argument to this function and CLI argument parser. Then, iterate over subsets only if `subset == None`. In addition, I think this code should have a problem if there is no item for the given `query_id`. You need to raise an appropriate error in this case.

src/datumaro/cli/commands/explore.py
```diff
@@ -125,20 +140,33 @@ def explore_command(args):
...
-            if not query_datasetitem:
-                break
+    if args.query_img_path:
+        querys = (
+            [args.query_img_path]
+            if not isinstance(args.query_img_path, list)
+            else args.query_img_path
+        )
+        query_datasetitems = []
+        for query_ in querys:
```
vinnamkim, line 155: Check the file existence first and raise proper errors if not exist.

src/datumaro/cli/commands/explore.py
```diff
@@ -125,20 +140,33 @@ def explore_command(args):
...
+        query_datasetitems = []
+        for query in querys:
+            for dataset in source_datasets:
+                query_datasetitem = match_query_subset(query, dataset)
+            query_datasetitems.append(query_datasetitem)
+    elif args.query_str:
+        query_datasetitems = args.query_str
     else:
-        query_datasetitem = args.query
+        raise
```
vinnamkim: I feel that you can remove these two lines since you added the mutually exclusive groups with `required=true`.

src/datumaro/cli/commands/explore.py
```diff
@@ -125,20 +149,29 @@ def explore_command(args):
...
-    else:
-        query_datasetitem = args.query
-
-    results = explorer.explore_topk(query_datasetitem, args.topk)
+    if args.query_img_path:
+        querys = check_and_convert_to_list(args.query_img_path)
+        query_datasetitems = []
+        for query_ in querys:
+            query_datasetitem = DatasetItem(
+                id=random.randrange(100, 1000), media=Image.from_file(query_)
```
vinnamkim: Basically, `id` type should be `str`, not `int`. If you want to generate a random unique id, please use `uuid4`: https://docs.python.org/ko/3/library/uuid.html.

src/datumaro/cli/commands/explore.py
```diff
@@ -153,9 +153,7 @@ def explore_command(args):
         querys = check_and_convert_to_list(args.query_img_path)
         query_datasetitems = []
         for query_ in querys:
-            query_datasetitem = DatasetItem(
-                id=random.randrange(100, 1000), media=Image.from_file(query_)
-            )
+            query_datasetitem = DatasetItem(id=uuid.uuid4(), media=Image.from_file(query_))
```
vinnamkim:
```suggestion
            query_datasetitem = DatasetItem(id=str(uuid.uuid4()), media=Image.from_file(query_))
```

vinnamkim (approved):
LGTM.

PR #1101, opened 2023-07-17
Fix: Dataset infos() can be broken if a transform not redefining infos() is stacked on the top

vinnamkim:
### Summary
- Ticket no. 115725
- Fix: Dataset infos() can be broken if a transform not redefining infos() is stacked on the top
- Enhance the StreamDatasetStorage transform tests added in #1077.
- Test `call_count` as well in the tests to validate stacked transforms.

### Checklist

### License

```python
#
```

### DIFF
--- a/tests/unit/components/test_dataset_storage.py
+++ b/tests/unit/components/test_dataset_storage.py
@@ -151,11 +151,11 @@ class StreamDatasetStorageTest:
         # Check extractor categories
         assert storage.categories() == fxt_categories
 
-        # Check extractor infos
         mapping = {"car": "apple", "cat": "banana", "dog": "cinnamon"}
         storage.transform(RemapLabels, mapping=mapping)
         assert fxt_stream_extractor.__iter__.call_count == 0
 
+        # Stack Rename (ItemTransform) on the top
         storage.transform(Rename, regex="|item_|rename_|")
         assert fxt_stream_extractor.__iter__.call_count == 0
 
--- a/tests/unit/components/test_dataset_storage.py
+++ b/tests/unit/components/test_dataset_storage.py
@@ -75,29 +75,35 @@ class StreamDatasetStorageTest:
         n_calls = 1
 
         self._test_loop(fxt_stream_extractor, storage, n_calls)
+        assert fxt_stream_extractor.__iter__.call_count == 1
 
         # Stack transform 1 level
         storage.transform(Rename, regex="|item_|rename_|")
         self._test_loop(fxt_stream_extractor, storage, n_calls, id_pattern="rename_{idx}")
+        assert fxt_stream_extractor.__iter__.call_count == 2
 
         # Stack transform 2 level
         storage.transform(Rename, regex="|rename_|renameagain_|")
         self._test_loop(fxt_stream_extractor, storage, n_calls, id_pattern="renameagain_{idx}")
+        assert fxt_stream_extractor.__iter__.call_count == 3
 
     def test_subset_transform(self, fxt_stream_extractor: MagicMock):
         storage = StreamDatasetStorage(source=fxt_stream_extractor)
 
         self._test_subsets(fxt_stream_extractor, storage)
+        assert fxt_stream_extractor.__iter__.call_count == 1
 
         # Stack transform 1 level
         storage.transform(RandomSplit, splits=[("train", 0.5), ("val", 0.5)], seed=3003)
         self._test_subsets(fxt_stream_extractor, storage, expect={"train", "val"})
+        assert fxt_stream_extractor.__iter__.call_count == 2
 
         # Stack transform 2 level
         storage.transform(
             MapSubsets, mapping={"train": DEFAULT_SUBSET_NAME, "val": DEFAULT_SUBSET_NAME}
         )
         self._test_subsets(fxt_stream_extractor, storage)
+        assert fxt_stream_extractor.__iter__.call_count == 3
 
     def test_info_transform(self, fxt_stream_extractor: MagicMock, fxt_infos: DatasetInfo):
         storage = StreamDatasetStorage(source=fxt_stream_extractor)
@@ -108,6 +114,7 @@ class StreamDatasetStorageTest:
         storage.transform(ProjectInfos, dst_infos=dst_infos)
 
         assert storage.infos().get("new") == "info"
+        assert fxt_stream_extractor.__iter__.call_count == 0
 
     def test_categories_transform(
         self, fxt_stream_extractor: MagicMock, fxt_categories: CategoriesInfo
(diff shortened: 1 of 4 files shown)


tests/unit/components/test_dataset_storage.py
```diff
@@ -108,6 +114,7 @@ def test_info_transform(self, fxt_stream_extractor: MagicMock, fxt_infos: Datase
         storage.transform(ProjectInfos, dst_infos=dst_infos)
 
         assert storage.infos().get("new") == "info"
+        assert fxt_stream_extractor.__iter__.call_count == 0
```
wonjuleee, line 117: I have one quick question. If it was `DatasetStorage`, `storage` was created from the `fxt_extractor` and it called the `infos()`, which may run `init_cache()`. If so, all items within `fxt_extractor` is also called? What I asked is, `fxt_extractor.__iter__.call_count == 1` for `DatasetStorage`?

wonjuleee (commented):
Don't we need to have `DatasetStorageTest` to compare them?

tests/unit/components/test_dataset_storage.py
```diff
@@ -122,3 +129,5 @@ def test_categories_transform(
         actual = set(cat.name for cat in storage.categories()[AnnotationType.label])
         expect = set(mapping.values())
         assert actual == expect
+
```
wonjuleee, line 132: With `StreamDatasetStorage`, `storage.infos()` returns only `self.stacked_transform.infos()`. What happens if there is no transform in `StreamDatasetStorage`?

tests/unit/components/test_dataset_storage.py
vinnamkim: Yes

vinnamkim:
> Don't we need to have `DatasetStorageTest` to compare them?

`tests/unit/components/test_dataset_storage.py` has the tests to validate `StreamDatasetStorage` behave as we expect. `DatasetStorage` has been used as the core of other components, so that our many existing tests have been tested `DatasetStorage` I think.

wonjuleee:
> > Don't we need to have `DatasetStorageTest` to compare them?
> 
> `tests/unit/components/test_dataset_storage.py` has the tests to validate `StreamDatasetStorage` behave as we expect. `DatasetStorage` has been used as the core of other components, so that our many existing tests have been tested `DatasetStorage` I think.

Partially agree. In my opinion, it was definitely tested in terms of functional, but it was not for performance. Your test for `StreamDatasetStorage` looks like related to performance (how many times an iterator called) and how it operated internally. If you agree, let me prepare the tests for `DatasetStorage` with similar shape in another PR. Please give me your opinion on this. 

Moreover, we may need to evaluate the quality of operations in terms of time and memory consumption in the future.

vinnamkim:
> > > Don't we need to have `DatasetStorageTest` to compare them?
> > 
> > 
> > `tests/unit/components/test_dataset_storage.py` has the tests to validate `StreamDatasetStorage` behave as we expect. `DatasetStorage` has been used as the core of other components, so that our many existing tests have been tested `DatasetStorage` I think.
> 
> Partially agree. In my opinion, it was definitely tested in terms of functional, but it was not for performance. Your test for `StreamDatasetStorage` looks like related to performance (how many times an iterator called) and how it operated internally. If you agree, let me prepare the tests for `DatasetStorage` with similar shape in another PR. Please give me your opinion on this.
> 
> Moreover, we may need to evaluate the quality of operations in terms of time and memory consumption in the future.

No, `tests/unit/components/test_dataset_storage.py` was 100% designed for the functional. The reason why we confirm how many the `__iter__()` of extractor is called is that we want to confirm that `StreamDatasetStorage` behaves in a stream manner (to test its functional, not performance). If it is `DatasetStorage`, `__iter__().call_count` always remains one although we stack any number of transforms (because of automatic caching).

To test the performance, I think that the unit test is not right place for it. Unit tests are aim to test the functionality of small components as possible. Therefore, there should exist pass or fail always. However, the performance test can be easily flaky by many variable factors (machine status, network status, ...). In my opinion, performance tests usually are conducted periodically and produce a report rather than testing pass/fail like as we do in NNCF.

wonjuleee:
> > > > Don't we need to have `DatasetStorageTest` to compare them?
> > > 
> > > 
> > > `tests/unit/components/test_dataset_storage.py` has the tests to validate `StreamDatasetStorage` behave as we expect. `DatasetStorage` has been used as the core of other components, so that our many existing tests have been tested `DatasetStorage` I think.
> > 
> > 
> > Partially agree. In my opinion, it was definitely tested in terms of functional, but it was not for performance. Your test for `StreamDatasetStorage` looks like related to performance (how many times an iterator called) and how it operated internally. If you agree, let me prepare the tests for `DatasetStorage` with similar shape in another PR. Please give me your opinion on this.
> > Moreover, we may need to evaluate the quality of operations in terms of time and memory consumption in the future.
> 
> No, `tests/unit/components/test_dataset_storage.py` was 100% designed for the functional. The reason why we confirm how many the `__iter__()` of extractor is called is that we want to confirm that `StreamDatasetStorage` behaves in a stream manner (to test its functional, not performance). If it is `DatasetStorage`, `__iter__().call_count` always remains one although we stack any number of transforms (because of automatic caching).
> 
> To test the performance, I think that the unit test is not right place for it. Unit tests are aim to test the functionality of small components as possible. Therefore, there should exist pass or fail always. However, the performance test can be easily flaky by many variable factors (machine status, network status, ...). In my opinion, performance tests usually are conducted periodically and produce a report rather than testing pass/fail like as we do in NNCF.

I don't think it is functional. I think we should run `StreamDatasetStorage` with other components (such as merge, transform, etc) and should check the same expected results for the functional tests.

tests/unit/components/test_dataset_storage.py
vinnamkim: It actually has a problem. Therefore, I made a bug ticket in the JIRA (https://jira.devtools.intel.com/browse/CVS-115725). https://github.com/openvinotoolkit/datumaro/pull/1101/commits/0f18b96cf74ce2506d6714db6f0a727cef37942e resolves it and it also adds the tests which will be failed if it is not corrected.

vinnamkim: This problem is not exactly what you ask. Therefore, the answer to your question is that it just returns the extractor's `infos()`. Please see this test:
https://github.com/openvinotoolkit/datumaro/blob/32317f1f40894cd3bf6ab8b95c84777eebb43a0c/tests/unit/components/test_dataset_storage.py#L62-L66

wonjuleee (approved):

vinnamkim:
> I don't think it is functional. I think we should run `StreamDatasetStorage` with other components (such as merge, transform, etc) and should check the same expected results for the functional tests.

We exactly have been doing that by parametrizing `Dataset` and `StreamDataset` for the importer and exporter tests (yes, not for merge, but transform is did in this test, `tests/unit/components/test_dataset_storage.py`, but for the `DatasetStorage` level, not for `Dataset` level still). However, I mean that this test is aimed to prove it's core mechanism (no caching and streaming) and behaviors.

PR #1115, opened 2023-08-02
Remove xfail marks from the convert integration tests

vinnamkim:
### Summary

- Ticket no. 115954
- Remove xfail marks from the convert integration tests
- This is because some data format should require annotation's `id` and `group` to be re-indexed, which should be taken into account when comparing the two datasets for equality.
- It means that there is no functional failures, so this PR just change the testing.

### How to test
I updated the target tests.

### Checklist

### License

```python
#
```

### DIFF
--- a/tests/utils/test_utils.py
+++ b/tests/utils/test_utils.py
@@ -119,24 +119,47 @@ def compare_categories(test, expected, actual):
 IGNORE_ALL = "*"
 
 
-def _compare_annotations(expected, actual, ignored_attrs=None):
-    if not ignored_attrs:
-        return expected == actual
+def _compare_annotations(
+    expected,
+    actual,
+    ignored_attrs=None,
+    ignore_ann_id: bool = False,
+    ignore_ann_group: bool = False,
+):
+    a_id, b_id = getattr(expected, "id"), getattr(actual, "id")
+    a_group, b_group = getattr(expected, "group"), getattr(actual, "group")
+    a_attr, b_attr = expected.attributes, actual.attributes
 
-    a_attr = expected.attributes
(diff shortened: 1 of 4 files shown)


tests/utils/test_utils.py
```diff
@@ -119,24 +119,47 @@ def compare_categories(test, expected, actual):
...
-    actual.attributes = b_attr
+    if ignore_ann_id:
+        setattr(expected, "id", a_id)
+        setattr(actual, "id", b_id)
+
+    if ignore_ann_group:
+        setattr(expected, "group", a_group)
+        setattr(actual, "group", b_group)
+
+    if ignored_attrs != None:
```
bonhunko:
```suggestion
    if ignored_attrs is not None:
```

tests/utils/test_utils.py
```diff
@@ -119,24 +119,47 @@ def compare_categories(test, expected, actual):
...
 
-    if ignored_attrs != IGNORE_ALL:
-        expected.attributes = filter_dict(a_attr, exclude_keys=ignored_attrs)
-        actual.attributes = filter_dict(b_attr, exclude_keys=ignored_attrs)
-    else:
+    if ignore_ann_group:
+        setattr(expected, "group", 0)
+        setattr(actual, "group", 0)
+
+    if ignored_attrs == None:
```
bonhunko:
```suggestion
    if ignored_attrs is None:
```
from pep8 style guide,
Comparisons to singletons like None should always be done with is or is not, never the equality operators.

bonhunko (approved):
Looks good to me, just left some minor style suggestion.

PR #1116, opened 2023-08-03
Enhance ClassificationValidator to cope with multilabel classification datasets with label_groups

wonjuleee:
### Checklist

### License

```python
#
```


src/datumaro/plugins/validators.py
vinnamkim, line 1: Would you please add a unit test for this change?

src/datumaro/plugins/validators.py
```diff
@@ -547,11 +547,24 @@ def compute_statistics(self, dataset):
...
+
         stats["items_with_multiple_labels"] = []
         for item_key, anns in filtered_anns:
-            ann_count = len(anns)
-            if ann_count > 1:
-                stats["items_with_multiple_labels"].append(item_key)
+            occupied_groups = set()
+            for ann in anns:
+                label_name = label_cat.items[ann.label].name
+                label_group = label_name_to_group.get(label_name, "")
```
vinnamkim: This will invoke L565 although `dm_label_groups` is an empty list since `label_group` will be `""` in that situation.

In addition, we might need to consider `group_type`. I feel that this logic should be effective only for `group_type == "exclusive"`
https://github.com/openvinotoolkit/datumaro/blob/04dff62070c4e51683e6133c88f0187e479b91b1/src/datumaro/components/annotation.py#L122-L126

Lastly, `group_type` seems selectable one, we can make it clearer by changing it from `str` to `IntEnum`.

src/datumaro/plugins/validators.py
wonjuleee: I implemented.

src/datumaro/plugins/validators.py
wonjuleee: Please see c2f6cc06f25d787b488970737e3732db5212ebf8.

src/datumaro/plugins/data_formats/datumaro/exporter.py
```diff
@@ -343,7 +343,7 @@ def _convert_label_categories(self, obj):
             converted["label_groups"].append(
                 {
                     "name": cast(label_group.name, str),
-                    "group_type": cast(label_group.group_type, str),
+                    "group_type": cast(label_group.group_type, int),
```
vinnamkim: For the backward compatibility of Datumaro data format, it seems we serialize it with string as same as before. If we change this, we might have to update the version of Datumaro data format. And, do we need to update the import part also?

src/datumaro/plugins/validators.py
```diff
@@ -547,11 +547,31 @@ def compute_statistics(self, dataset):
 
         stats, filtered_anns = self._compute_common_statistics(dataset)
 
+        label_cat = dataset.categories()[AnnotationType.label]
+        label_groups = label_cat.label_groups
+
+        label_name_to_group = {}
+        for label_group in label_groups:
+            for idx, label_name in enumerate(label_group.labels):
+                if label_group.group_type == 0:
```
vinnamkim: For better interpretability,
```suggestion
                if label_group.group_type == GroupType.EXCLUSIVE:
```

src/datumaro/plugins/validators.py
```diff
@@ -547,11 +547,31 @@ def compute_statistics(self, dataset):
...
-            ann_count = len(anns)
-            if ann_count > 1:
-                stats["items_with_multiple_labels"].append(item_key)
+            occupied_groups = set()
+            for ann in anns:
+                if ann.label in undefined_label_name:
+                    continue
+                label_name = label_cat[ann.label].name
+                label_group = label_name_to_group.get(label_name, "")
+                if label_group in occupied_groups:
```
vinnamkim: For better interpretability,
```suggestion
                label_group = label_name_to_group.get(label_name, None)
                if label_group is not None and label_group in occupied_groups:
```

src/datumaro/plugins/data_formats/datumaro/exporter.py
vinnamkim: For example,
```python
class GroupType(IntEnum):
    EXCLUSIVE = 0
    INCLUSIVE = 1
    RESTRICTED = 2

    def to_str(self) -> str:
          ...

    @classmethod
    def from_str(cls, text: str) -> GroupType:
          ...
```

wonjuleee: Thanks. I added at fe740c48f62a2808dc5157c756e23ac29cc28984

src/datumaro/plugins/validators.py
wonjuleee: Fixed. Thanks.

src/datumaro/plugins/validators.py
wonjuleee: I updated to consider about other `GroupType`s

vinnamkim (approved):
LGTM, but would you insert a new line at the end of files? You can enable this option in your VSCode to make it always on.

image

tests/assets/datumaro_dataset/legacy/annotations/test.json
```diff
@@ -73,4 +79,4 @@
             }
         }
     ]
-}
+}
```
vinnamkim: Please add a new line.

tests/assets/datumaro_dataset/legacy/annotations/train.json
```diff
@@ -99,4 +105,4 @@
             }
         }
     ]
-}
+}
```
vinnamkim: Please add a new line.

tests/assets/datumaro_dataset/legacy/annotations/validation.json
```diff
@@ -54,4 +60,4 @@
             }
         }
     ]
-}
+}
```
vinnamkim: Please add a new line.

wonjuleee:
> LGTM, but would you insert a new line at the end of files? You can enable this option in your VSCode to make it always on.
> 
> image

Thank you for letting me know this. I set up settings.json and added new lines in annotation files.

PR #1119, opened 2023-08-09
add boxes-to-polygons transform

jihyeonyi:
### Summary
Add bbox to polygon transform.

### Checklist

### License

```python
#
```

### DIFF
--- a/src/datumaro/plugins/transforms.py
+++ b/src/datumaro/plugins/transforms.py
@@ -268,14 +268,10 @@ class BoxesToMasks(ItemTransform, CliPlugin):
 
 class BoxesToPolygons(ItemTransform, CliPlugin):
     def transform_item(self, item):
-        annotations = []
-        for ann in item.annotations:
-            if ann.type == AnnotationType.bbox:
-                if not isinstance(item.media, Image):
-                    raise Exception("Image info is required for this transform")
-                annotations.append(self.convert_bbox(ann))
-            else:
-                annotations.append(ann)
+        annotations = [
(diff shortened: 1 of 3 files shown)


src/datumaro/plugins/transforms.py
```diff
@@ -266,6 +266,31 @@ def convert_bbox(bbox, img_h, img_w):
         )
 
 
+class BoxesToPolygons(ItemTransform, CliPlugin):
+    def transform_item(self, item):
+        annotations = []
+        for ann in item.annotations:
+            if ann.type == AnnotationType.bbox:
+                if not isinstance(item.media, Image):
+                    raise Exception("Image info is required for this transform")
```
vinnamkim: Is it still required?

src/datumaro/plugins/transforms.py
```diff
@@ -266,6 +266,31 @@ def convert_bbox(bbox, img_h, img_w):
...
+                annotations.append(self.convert_bbox(ann))
+            else:
+                annotations.append(ann)
```
vinnamkim:
```suggestion
        annotations = [
             self.convert_bbox(ann) if ann.type == AnnotationType.bbox else ann
             for ann in item.annotations
        ]
```

jihyeonyi: Oh, you're right. We don't need (w,h) of image for conversion, so, media type condition isn't needed.

vinnamkim (approved):
LGTM.

PR #1124, opened 2023-08-14
support video annotation

bonhunko:
### Summary

[CVS-116105](https://jira.devtools.intel.com/browse/CVS-116105)

Support the video annotation type for 'datumaro', 'datumaro_binary' format

### Checklist

### License

```python
#
```

### DIFF
--- a/src/datumaro/components/annotation.py
+++ b/src/datumaro/components/annotation.py
@@ -53,6 +53,7 @@ class AnnotationType(IntEnum):
 COORDINATE_ROUNDING_DIGITS = 2
 CHECK_POLYGON_EQ_EPSILONE = 1e-7
 NO_GROUP = 0
+NO_OBJECT_ID = -1
 
 
 @attrs(slots=True, kw_only=True, order=False)
@@ -82,6 +83,11 @@ class Annotation:
     # Annotations can be grouped, which means they describe parts of a
     # single object. The value of 0 means there is no group.
     group: int = field(default=NO_GROUP, validator=default_if_none(int))
+
+    # obeject identifier over the multiple items
+    # e.g.) in a video, person 'A' could be annotated on the multiple frame images
+    #   the user could assign >=0 value as id of person 'A'.
+    object_id: int = field(default=NO_OBJECT_ID, validator=default_if_none(int))
 
     _type = AnnotationType.unknown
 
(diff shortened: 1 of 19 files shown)


src/datumaro/components/annotation.py
```diff
@@ -83,6 +84,11 @@ class Annotation:
     # single object. The value of 0 means there is no group.
     group: int = field(default=NO_GROUP, validator=default_if_none(int))
 
+    # obeject identifier over the multiple items
+    # e.g.) in a video, person 'A' could be annotated on the multiple frame images
+    #   the user could assign >=0 value as id of person 'A'.
+    object_id: int = field(default=NO_OBJECT_ID, validator=default_if_none(int))
```
jihyeonyi, line 90: This can be a stupid question, but can we use 'id' but change its default value to -1, instead of using object_id?

src/datumaro/plugins/data_formats/datumaro/base.py
```diff
@@ -174,6 +178,21 @@ def _parse_item(self, item_desc: Dict) -> Optional[DatasetItem]:
...
+                if video_path not in self._videos:
+                    self._videos[video_path] = Video(video_path)
+                video = self._videos[video_path]
+
+                frame_index = video_frame_info.get("frame_index")
+
+                media = VideoFrame(video, frame_index)
+                if self.media_type == MediaElement:
+                    self.media_type = VideoFrame
+
```
jihyeonyi, line 193: Can you add some more tests to increase test coverage? (not necessarily limited to this line)

src/datumaro/components/annotation.py
jihyeonyi: I recognized the purpose of two variables are different. I'd like to withdraw my previous comment.

bonhunko: This is a good point. Wonju and I discuss what if we could use `id` to distinguish the objects over the frames.
But we conclude that `id` better be remains as it is since `id` has its own features.

src/datumaro/plugins/data_formats/datumaro/base.py
bonhunko: I added a unit test to archive 80>% coverage

jihyeonyi (approved):
Thanks for your effort.

wonjuleee (approved):

PR #1125, opened 2023-08-14
Enable Torch conversion from Dataset

wonjuleee:
### Checklist

### License

```python
#
```


src/datumaro/components/merge/extractor_merger.py
```diff
@@ -58,8 +58,8 @@ def get(self, id: str, subset: Optional[str] = None):
                 if item is not None:
                     return item
 
-        s = self._subset[subset]
-        return s.get(id)
```
vinnamkim, line 62: Why do we need this change? It will return `subset: str` rather than `None` if there is no item for the given `id`. I think that it should return `None` as before.

src/datumaro/plugins/framework_converter.py
```diff
@@ -0,0 +1,176 @@
...
+            target_transform: Optional[Callable] = None,
+        ):
+            self.dataset = dataset.get_subset(subset)
+            self.subset = subset
+            self.task = task
+            self.transform = transform
+            self.target_transform = target_transform
+            self._ids = []
+            for item in self.dataset:
+                self._ids.append(item.id)
```
vinnamkim:
```suggestion
            self._ids = [item.id for item in self.dataset]
```

src/datumaro/plugins/framework_converter.py
```diff
@@ -0,0 +1,176 @@
...
+
+            if len(image.shape) == 2:
+                image = np.expand_dims(image, axis=-1)
+
+            if self.task == "classification":
+                label = [
+                    ann.label
+                    for ann in dataitem.annotations
+                    if ann.type == TASK_ANN_TYPE[self.task]
+                ][0]
```
vinnamkim: Shouldn't we limit it from `classification` to `multi-class-classification` because we pick only one label. It is also preferred to add exception or an error log when there are multi labels.

src/datumaro/plugins/framework_converter.py
```diff
@@ -0,0 +1,176 @@
...
+                    ann.as_dict()
+                    for ann in dataitem.annotations
+                    if ann.type == TASK_ANN_TYPE[self.task]
+                ]
+            elif self.task == "semantic_segmentation":
+                label = [
+                    ann.image
+                    for ann in dataitem.annotations
+                    if ann.type == TASK_ANN_TYPE[self.task]
+                ][0]
```
vinnamkim: `Mask` is a binary mask with `label`, so that an image can multiple `Mask`s as far as I know.

https://github.com/openvinotoolkit/datumaro/blob/cf244a29199b9132f974e9f2d35bd497c555860c/src/datumaro/components/annotation.py#L356-L366

src/datumaro/plugins/framework_converter.py
```diff
@@ -0,0 +1,176 @@
...
+                        for ann in item.annotations
+                        if ann.type == TASK_ANN_TYPE[self.task]
+                    ][0]
+
+                yield image, label
+
+        def create(self) -> tf.data.Dataset:
+            tf_dataset = tf.data.Dataset.from_generator(
+                self.generator_wrapper, output_signature=self.output_signature
+            )
```
vinnamkim: I don't know how it works if
```python
self.create().shuffle()
```

I don't know exactly but I guess it might generate all the items and store them in the memory to shuffle.
Please see this https://medium.com/@acordier/tf-data-dataset-generators-with-parallelization-the-easy-way-b5c5f7d2a18.

We might create `tf.data.Dataset.from_tensors(item_ids)` and use `.map()` function to reconstruct the item from the `id in item_ids`.

src/datumaro/plugins/framework_converter.py
wonjuleee: fix aa2110a077f18bcaa22ff07fef62e225958e9c4f.

src/datumaro/plugins/framework_converter.py
wonjuleee: Thanks, I added log for warning this and a new multilabel classification task at aa2110a077f18bcaa22ff07fef62e225958e9c4f.

src/datumaro/plugins/framework_converter.py
wonjuleee: Thank you for pointing out this. I updated this by pixel-wise classification map at aa2110a077f18bcaa22ff07fef62e225958e9c4f.

src/datumaro/components/merge/extractor_merger.py
wonjuleee: When I added a test `test_can_convert_torch_framework_detection`, I got an error at https://github.com/openvinotoolkit/datumaro/blob/cf244a29199b9132f974e9f2d35bd497c555860c/src/datumaro/plugins/data_formats/coco/extractor_merger.py#L87 and this might be differ to the original `extractor_merger`'s one https://github.com/openvinotoolkit/datumaro/blob/cf244a29199b9132f974e9f2d35bd497c555860c/src/datumaro/components/merge/extractor_merger.py#L61 So I thought it was a type between "self._subsets" and "self._subset". On the other hands, most base class `_DatasetBase` has "self._subsets" https://github.com/openvinotoolkit/datumaro/blob/cf244a29199b9132f974e9f2d35bd497c555860c/src/datumaro/components/dataset_base.py#L232

src/datumaro/plugins/framework_converter.py
vinnamkim: If we follow this way, we can make a common base class for `torch` and `tf` such as
```python
class Base:
     def _gen_item(self, idx: int):
           ....
class Torch:
      def __getitem__(self, idx: int):
            return self._gen_item(idx)
class TF:
      def create(self):
               return tf.data.Dataset.from_tensors(item_ids).map(lambda idx: self._gen_item(idx))
```

src/datumaro/components/merge/extractor_merger.py
vinnamkim: Sorry I mistook it for Python's `dict.get()`. Resolve this conversation.

src/datumaro/plugins/framework_converter.py
wonjuleee: fixed at c9a2356f47ecc2117fd76ae8d62b4bc3f01da149. Thanks.

src/datumaro/plugins/framework_converter.py
```diff
@@ -0,0 +1,189 @@
...
+            label = [
+                ann.as_dict() for ann in item.annotations if ann.type == TASK_ANN_TYPE[self.task]
+            ]
+        elif self.task == "semantic_segmentation":
+            masks = [
+                ann.as_class_mask()
+                for ann in item.annotations
+                if ann.type == TASK_ANN_TYPE[self.task]
+            ]
+            label = np.sum(masks, axis=0, dtype=np.uint8)
```
vinnamkim: I'm worrying that this summation can make problems when there are some overlapped pixels between binary masks. Would you please consider using our `mask_tools.merge_masks()` util function?

https://github.com/openvinotoolkit/datumaro/blob/cf244a29199b9132f974e9f2d35bd497c555860c/src/datumaro/util/mask_tools.py#L378-L406

wonjuleee: fixed at cac9e9055749765f6ef306fed825c60a75fb80a2

vinnamkim (approved):
LGTM.

PR #1142, opened 2023-09-06
Add ConfigurableValidator

wonjuleee:
### Checklist

### License

```python
#
```

### DIFF
--- a/src/datumaro/plugins/configurable_validator.py
+++ b/src/datumaro/plugins/configurable_validator.py
@@ -1,8 +1,8 @@
 # Copyright (C) 2023 Intel Corporation
 #
 # SPDX-License-Identifier: MIT
-import logging as log
-from typing import List, Set
+from dataclasses import dataclass, field
+from typing import Dict, List, Optional, Set, Tuple
 
 import numpy as np
 
@@ -35,6 +35,22 @@ from datumaro.components.errors import (
 from datumaro.components.validator import Severity, TaskType, Validator
 
 
+@dataclass
+class StatsData:
+    categories: Dict[str, Dict[str, Set[str]]]
+    undefined_attribute: Set[Tuple[str, str, str]]  # item_id, item_subset, label_name, attr_name
+    undefined_label: Set[Tuple[str, str, str]]  # item_id, item_subset, label_name
+    missing_attribute: Set[Tuple[str, str, str, str]]  # item_id, item_subset, label_name, attr_name
+    missing_label: Set[Tuple[str, str]]  # item_id, item_subset
+    multiple_label: Optional[Set[Tuple[str, str]]] = field(default=None)  # item_id, item_subset
+    invalid_value: Optional[Set[Tuple[str, str]]] = field(
+        default=None
+    )  # item_id, item_subset, annotation_id
+    negative_length: Optional[Set[Tuple[str, str]]] = field(
+        default=None
+    )  # item_id, item_subset, annotation_id
+
+
 class _BaseAnnStats:
     ATTR_WARNINGS = {
         AttributeDefinedButNotFound,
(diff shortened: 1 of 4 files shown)


src/datumaro/plugins/configurable_validator.py
```diff
@@ -0,0 +1,732 @@
...
+            self.stats[label_name]["pts"] = []  # (id, subset, ann_id, w, h)
+
+        self.stats["invalid_value"] = set()
+        self.stats["negative_length"] = set()
+
+    def _update_item_type_stats(self, item: DatasetItem):
+        pass
+
+    def _update_ann_type_stats(self, item_key: tuple, annotation: Annotation):
+        if self.BBOX_WARNINGS & self.warnings:
```
vinnamkim: I hope that we build a collection of update functions when initializing this class instance according to the input configuration. This is for doing not to check its configuration every time during the dataset item loop. We can iterate on the collection of update functions.

src/datumaro/plugins/configurable_validator.py
```diff
@@ -0,0 +1,732 @@
...
+    def __init__(self, label_categories: LabelCategories, warnings: set):
+        super().__init__(label_categories=label_categories, warnings=warnings)
+
+        self.items = {}
+        for label_cat in self.label_categories.items:
+            label_name = label_cat.name
+            self.stats[label_name]["pts"] = []  # (id, subset, ann_id, w, h)
+
+        self.stats["invalid_value"] = set()
+        self.stats["negative_length"] = set()
```
vinnamkim: Could we have a typed schema, e.g. `@dataclass`, for the output data store? It currently uses a dictionary with string key and it is hard to see what is going on this class.

src/datumaro/plugins/configurable_validator.py
```diff
@@ -0,0 +1,732 @@
...
+
+    def compute_statistics(self, dataset):
+        if not self.init_flag:
+            self._label_categories = dataset.categories()[AnnotationType.label]
+            self._init_stats_collector(label_categories=self._label_categories)
+
+        for item in dataset:
+            for stats_collector in self.all_stats.values():
+                if not stats_collector:
+                    continue
```
vinnamkim: This is same for the comment above. Can we prepare the collection (list or tuple) of collectors in front of `compute_statistics()`? I don't see the reason why we keep `None` for the collector not used.

src/datumaro/plugins/configurable_validator.py
```diff
@@ -0,0 +1,732 @@
...
+        return {task: stats_collector.stats for task, stats_collector in self.all_stats.items()}
+
+    def generate_reports(self, task_stats):
+        reports = {}
+
+        for task, stats in task_stats.items():
+            reports[task] = []
+            if task == TaskType.classification:
+                if {MultiLabelAnnotations} & self.warnings:
+                    reports[task] += self._check_multiple_label(stats)
```
vinnamkim: Although we produce the reports as `dict`, is it possible to update each part of the dictionary in each collector's member function? I hope that we do not gather all the things getting together in this function.

src/datumaro/plugins/configurable_validator.py
wonjuleee: I have tried to register function for warning when initialization at cc10ff43ab54b743fc524f29ad1f4bc016eaa04b. Could you check this?

src/datumaro/plugins/configurable_validator.py
wonjuleee: I introduced StatsData at cc10ff43ab54b743fc524f29ad1f4bc016eaa04b.

src/datumaro/plugins/configurable_validator.py
```diff
@@ -0,0 +1,967 @@
...
+            AnnotationType.mask,
+            AnnotationType.polygon,
+            AnnotationType.ellipse,
+        ]:
+            return self.all_stats[TaskType.segmentation]
+        else:
+            return None
+
+    def compute_statistics(self, dataset):
+        if not self.init_flag:
```
vinnamkim: Do we need this flag? I guess that this class has configurations when it is created. With following these configuration and taking the `dataset` in this function argument, it produce statistics according to those two inputs. Therefore, we should initialize statistics carriers (self.all_stats) every time we take `dataset`, right?

src/datumaro/plugins/configurable_validator.py
```diff
@@ -0,0 +1,967 @@
...
+                if not stats_collector:
+                    continue
+                stats_collector.update_item(item)
+
+            item_key = (item.id, item.subset)
+            for annotation in item.annotations:
+                stats_collector = self._get_stats_collector(ann_type=annotation.type)
+                if not stats_collector:
+                    continue
+                stats_collector.update_ann(item_key, annotation)
```
vinnamkim: I haven't caught it up before, but it is very weird why there should exist the two paths between item update and annotation update at the top level although both can take `item` as their inputs. However, it requires many works to be fixed here, so that let's revisit afterwards.

src/datumaro/plugins/configurable_validator.py
```diff
@@ -0,0 +1,967 @@
...
+            return None
+
+    def compute_statistics(self, dataset):
+        if not self.init_flag:
+            self._init_stats_collector(label_categories=dataset.categories()[AnnotationType.label])
+
+        for item in dataset:
+            for stats_collector in self.all_stats.values():
+                if not stats_collector:
+                    continue
```
vinnamkim: Do we need this if statement? I think that L892 and `self._init_stats_collector()` make there is no `None` in `self.all_stats` without any exception. Similar to my above comment, wouldn't it be better to let `self._init_stats_collector()` return `all_stats` rather than storing it as the member variable? The `all_stats` are dependent to `dataset`, so that maintaining it as a member variable seems not good for understanding the data flow in a function.

src/datumaro/plugins/configurable_validator.py
wonjuleee: you meant that we have to compute statistics from scratch when dataset is passed everytime, is this correct?

src/datumaro/plugins/configurable_validator.py
wonjuleee: I believe, by introducing update_ann separately, we can reduce the annotation iterator. If we use only update_item, we have to run annotation iterator several times according to the desired task types. What do you think?

src/datumaro/plugins/configurable_validator.py
vinnamkim: Yes. However, if you imagine the functionality to update incrementally `all_stats` by calling `compute_statistics()` again and again for the sequence of `dataset`, I think it is better to introduce `all_stats` as the function argument (not member variables) for the more understandable data flow in this function.

src/datumaro/plugins/configurable_validator.py
vinnamkim: Yes, it is fair point since Python has severe overheads on calling iterations.

src/datumaro/plugins/configurable_validator.py
wonjuleee: I have removed init_flag following your opinion.

vinnamkim (approved):
LGTM.

PR #1145, opened 2023-09-11
Add multi-threading option to ModelTransform and SAMBboxToInstanceMask

vinnamkim:
### Summary

- Add multi-threading option (`num_workers > 0`) to `ModelTransform` and `SAMBboxToInstanceMask`.
- It is required if the model launcher can take multiple requests at the same time and have high throughput.

### How to test
Added some tests for this change.

### Checklist

### License

```python
#
```

### DIFF
--- a/src/datumaro/components/transformer.py
+++ b/src/datumaro/components/transformer.py
@@ -1,7 +1,8 @@
 # Copyright (C) 2019-2022 Intel Corporation
 #
 # SPDX-License-Identifier: MIT
-from typing import Generator, List, Optional
+from multiprocessing.pool import ThreadPool
+from typing import Generator, Iterator, List, Optional
 
 import numpy as np
 
@@ -10,6 +11,7 @@ from datumaro.components.cli_plugin import CliPlugin
 from datumaro.components.dataset_base import DatasetBase, DatasetItem, IDataset
 from datumaro.components.launcher import Launcher
 from datumaro.util import is_method_redefined, take_by
+from datumaro.util.multi_procs_util import consumer_generator
 
 
 class Transform(DatasetBase, CliPlugin):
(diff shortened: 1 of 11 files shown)


src/datumaro/components/transformer.py
```diff
@@ -71,35 +73,87 @@ def __iter__(self):
...
+
+        return [
+            self.wrap_item(
+                item,
+                annotations=item.annotations + annotations
+                if self._append_annotation
+                else annotations,
+            )
+            for item, annotations in zip(batch, inference)
+        ]
```
sooahleex, line 156: This function seems to be used in many places, so how about putting it in `datumato/util/__init__` or any `util`?

src/datumaro/plugins/sam_transforms/bbox_to_inst_mask.py
```diff
@@ -90,26 +92,35 @@ def __init__(
...
+            )
 
-                yield item.wrap(
+            items.append(
+                item.wrap(
                     annotations=self._convert_to_polygon(nested_masks[0])
                     if self._to_polygon
                     else nested_masks[0]
                 )
+            )
```
sooahleex: I think there are some unnecessary item wrapping.
```
batch_to_decode = [item.wrap(annotations=item.annotations + img_embed) for item, img_embed in zip(batch, img_embeds)]

nested_masks = self._sam_decoder_launcher.launch(batch_to_decode, stack=False)

items = []
for item, nested_mask in zip(batch, nested_masks):
    annotations = (self._convert_to_polygon(nested_masks[0]) if self._to_polygon else nested_masks[0])
    items.append(item.wrap(annotations=annotations))
```
I think it would be better to do item wrapping only once.

src/datumaro/components/transformer.py
```diff
@@ -71,35 +73,87 @@ def __iter__(self):
...
+                    future = pool.apply_async(
+                        func=self._process_batch,
+                        args=(batch,),
+                    )
+                    yield future
+
+            with consumer_generator(producer_generator=_producer_gen) as consumer_gen:
+                for future in consumer_gen():
+                    for item in future.get():
+                        yield item
```
wonjuleee, line 130: can we adopt this kind of function for other transforms too through `consumer_generator`?

src/datumaro/util/multi_procs_util.py
```diff
@@ -0,0 +1,91 @@
...
+                    if is_terminated:
+                        raise RuntimeError(
+                            "Item to enqueue is left. However, the main process is terminated."
+                        )
+
+    def _target(queue: Queue) -> None:
+        try:
+            _enqueue(ProducerMessage.START, queue)
+
+            for item in producer_generator():
```
sooahleex: It seems `producer_generator` is `Iterator`, so they could be use directly by `pruducer_generator`.
```suggestion
            for item in producer_generator:
```

src/datumaro/components/transformer.py
```diff
@@ -71,35 +73,87 @@ def __iter__(self):
...
+    def _iter_single_proc(self) -> Iterator[DatasetItem]:
+        for batch in take_by(self._extractor, self._batch_size):
+            for item in self._process_batch(batch=batch):
                 yield item
 
-    def _yield_item(
-        self, batch: List[DatasetItem], inference: List[List[Annotation]]
-    ) -> Generator[DatasetItem, None, None]:
-        for item, annotations in zip(batch, inference):
+    def _process_batch(
```
wonjuleee, line 137: Is there any special reason to rename `_yield_item` to `_process_batch`?

src/datumaro/components/transformer.py
vinnamkim: Yes, but this is a helper to limit the size of requests at the same time (`queue_size: int = 100`). The actual one does parallelization is `ThreadPool`.
https://github.com/openvinotoolkit/datumaro/blob/ff99cec1bbee8ffb4d66d54b67d8741a40c9297c/src/datumaro/components/transformer.py#L116-L117

The many other transforms are based on the Python implementation. We need to use `multiprocessing.pool.Pool` not `ThreadPool` since Python has Global Interpreter Lock (GIL), so that one Python code line can be executed at the same time under GIL. The `multiprocessing.pool.Pool` spawns multiple processes, not threads, we can avoid it.

src/datumaro/components/transformer.py
vinnamkim: Which one do you mean exactly? `def _process_batch(self)` is a member function so that it will be normally reused in its children.

src/datumaro/plugins/sam_transforms/bbox_to_inst_mask.py
vinnamkim: https://github.com/openvinotoolkit/datumaro/pull/1145/commits/865fb471695a9635ddd37b76383e1adc746aac1c decreases the number of calling `item.wrap` from two to one.

src/datumaro/util/multi_procs_util.py
vinnamkim: True, https://github.com/openvinotoolkit/datumaro/pull/1145/commits/865fb471695a9635ddd37b76383e1adc746aac1c fixes some wrong typings about it.

src/datumaro/components/transformer.py
vinnamkim: Before, we have two separated steps 1) do model inference and 2) wrap item from the inference output. However, we want to parallelize both sequential steps at once. `_process_batch` includes both steps, but `_yield_item` only includes 2).

sooahleex (approved):
Looks good to me.

PR #1168, opened 2023-10-06
Add SAMAutomaticMaskGeneration transform

vinnamkim:
### Summary

- Same as title
- Ticket no. 120781.

### How to test
Added unit tests for this change and added Jupyter notebook example which can validate this change.

### Checklist

### License

```python
#
```


src/datumaro/plugins/sam_transforms/automatic_mask_gen.py
```diff
@@ -0,0 +1,181 @@
...
+            batch=[item for item in batch if self._sam_encoder_launcher.type_check(item)]
+        )
+
+        items = []
+        for item, img_embed in zip(batch, img_embeds):
+            amg_masks: List[AMGMasks] = []
+
+            for i in range(0, len(self._points_grid), self.points_per_batch):
+                amg_points = [AMGPoints(points=self._points_grid[i : i + self.points_per_batch])]
+                item_to_decode = item.wrap(annotations=amg_points + img_embed)
```
wonjuleee, line 161: Is there no field to contain img_embed in item? It is okay for now, but it would be better to have img_embed in the future for data analysis. What do you think?

src/datumaro/plugins/sam_transforms/automatic_mask_gen.py
```diff
@@ -0,0 +1,181 @@
...
+
+                # Nested list of mask [[mask_0, ...]]
+                nested_masks: List[List[AMGMasks]] = self._sam_decoder_launcher.launch(
+                    [item_to_decode],
+                    stack=False,
+                )
```
wonjuleee, line 167: Can I ask why nested list is required? Is this related to `[item_to_decode]`? Finally, we only need to have only `nested_masks[0]`.

src/datumaro/plugins/sam_transforms/interpreters/sam_decoder_for_amg.py
```diff
@@ -0,0 +1,219 @@
...
+            )
+            masks = masks[keep_mask]
+            iou_preds = iou_preds[keep_mask]
+
+        binary_masks = masks > mask_threshold
+
+        if min_mask_region_area > 0:
+            keep_mask = binary_masks.sum(axis=(1, 2)) > min_mask_region_area
+            binary_masks = binary_masks[keep_mask]
+            iou_preds = iou_preds[keep_mask]
```
wonjuleee, line 112: These threshold (`pred_iou_threshold`, `stability_score_threshold`, and `min_mask_region_area`) looks like given by a user-end. How about enforcing them to be positive when creating `SAMAutomaticMaskGeneration` instead of these if conditions.

src/datumaro/plugins/sam_transforms/interpreters/sam_decoder_for_amg.py
```diff
@@ -0,0 +1,219 @@
...
+                    masks=masks,
+                    mask_threshold=mask_threshold,
+                    stability_score_offset=stability_score_offset,
+                )
+                > stability_score_thresh
+            )
+            masks = masks[keep_mask]
+            iou_preds = iou_preds[keep_mask]
+
+        binary_masks = masks > mask_threshold
```
wonjuleee, line 107: Minor comment: `mask_threshold` is fed into `self._calculate_stability_score(...)`, why don't you put `binary_masks = masks > mask_threshold` logic into `self._calculate_stability_score(...)`?

src/datumaro/plugins/sam_transforms/automatic_mask_gen.py
vinnamkim: Although we restrict it to give a single item, `Launcher.launch(x)` assumes `x` is a batch so that it requires `x` to be `list`. Thus, the output should be a nested list.

src/datumaro/plugins/sam_transforms/automatic_mask_gen.py
vinnamkim: Yes, unlike the 1:n relationship `annotations` field, we can add special fields to the dataset item that correspond 1:1 such as this feature embedding vector to it in the future.

src/datumaro/plugins/sam_transforms/interpreters/sam_decoder_for_amg.py
vinnamkim: Can I patch it in the following PR? This is because I'm hardly revisiting Arrow format for now so that checkout this branch and doing some works make me confusing.

src/datumaro/plugins/sam_transforms/interpreters/sam_decoder_for_amg.py
vinnamkim: The function `self._calculate_stability_score(...)` generates a `keep_mask`, which is a 1D boolean mask indicating what to retain after post-processing `masks` (an array of 2D images with a single floating-point number channel produced by the SAM decoder). A similar pattern of using `keep_mask` for post-processing is maintained in L91-93 and L110-112. So, the code flow seems unnatural if we let `self._calculate_stability_score(...)` to output `binary_masks`.

src/datumaro/plugins/sam_transforms/interpreters/sam_decoder_for_amg.py
wonjuleee: Sure, please.

src/datumaro/plugins/sam_transforms/interpreters/sam_decoder_for_amg.py
wonjuleee: Understood. Thanks.

wonjuleee (approved):
LGTM.

PR #1184, opened 2023-10-24
load CocoPageMapper once per process

jihyeonyi:
### Summary
When loading CocoPageMapper multiple times per interpreter or process, it would raise an ImportError as shown below.

image

So, I have modified it to import the module only once.

### Checklist

### License

```python
#
```

### DIFF
--- a/src/datumaro/plugins/data_formats/coco/page_mapper.py
+++ b/src/datumaro/plugins/data_formats/coco/page_mapper.py
@@ -5,7 +5,10 @@
 import logging as log
 from typing import Any, Dict, Iterator, List, Optional, Tuple
 
-from datumaro.rust_api import CocoPageMapper as CocoPageMapperImpl
+try:
+    from datumaro.rust_api import CocoPageMapper as CocoPageMapperImpl
+except ImportError:
+    pass
 
 __all__ = ["COCOPageMapper"]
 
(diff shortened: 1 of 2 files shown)


src/datumaro/plugins/data_formats/coco/page_mapper.py
```diff
@@ -5,7 +5,10 @@
 import logging as log
 from typing import Any, Dict, Iterator, List, Optional, Tuple
 
-from datumaro.rust_api import CocoPageMapper as CocoPageMapperImpl
+try:
+    from datumaro.rust_api import CocoPageMapper as CocoPageMapperImpl
+except ImportError:
+    pass
```
vinnamkim, line 11: What is the exact case requiring this fix? If any interpreter goes into L10-11, it means that it doesn't loads `CocoPageMapperImpl` just a once but does not load `CocoPageMapperImpl` totally.

jihyeonyi: The error occurs when importing a dataset multiple times in the streamlit GUI. 
I thought this problem can be reproduced easily in the normal interpreter, but it was not.
I guess I need to handle this problem in another way.
So, I'll cancel this change at the moment.

vinnamkim: What Python version is in your dev env? 3.12?

jihyeonyi: My Python version is 3.10.
Anyway, it's strange that I cannot reproduce this problem after reverting the code. 
I'll revisit this issue once it is reproduced. 
I might have made a mistake when dealing with the streamlit.

PR #1196, opened 2023-11-20
Upgrade OpenVINO

sovrasov:
### Checklist

### License

```python
#
```


wonjuleee (approved):

sovrasov:
@yunchu @wonjuleee in this PR, I relaxed OV requirement from == to >= to avoid cases when datumaro prevents other packages like OTX from upgrading the OV version. Do you foresee any issues if we use that strategy?
As an alternative, we can limit the upper version of OV to +2 versions from now, to enable reasonable space for the future upgrades, and avoid a complete versions mismatch in the future.

wonjuleee:
> @yunchu @wonjuleee in this PR, I relaxed OV requirement from == to >= to avoid cases when datumaro prevents other packages like OTX from upgrading the OV version. Do you foresee any issues if we use that strategy? As an alternative, we can limit the upper version of OV to +2 versions from now, to enable reasonable space for the future upgrades, and avoid a complete versions mismatch in the future.

Yes, it looks much better than fixing the version. Could you update `requirements-default.txt` too?

sovrasov:
> > @yunchu @wonjuleee in this PR, I relaxed OV requirement from == to >= to avoid cases when datumaro prevents other packages like OTX from upgrading the OV version. Do you foresee any issues if we use that strategy? As an alternative, we can limit the upper version of OV to +2 versions from now, to enable reasonable space for the future upgrades, and avoid a complete versions mismatch in the future.
> 
> Yes, it looks much better than fixing the version. Could you update `requirements-default.txt` too?

Done

sovrasov:
@wonjuleee  @yunchu apparently, unit tests on macos were killed by timeout, because they stall at some kind of download step
Is it a known issue? Unit testing on ubuntu (which probably download the same assets) was finished completed without errors
image

wonjuleee:
> @wonjuleee @yunchu apparently, unit tests on macos were killed by timeout, because they stall at some kind of download step Is it a known issue? Unit testing on ubuntu (which probably download the same assets) was finished completed without errors image

Hi @sovrasov, this is not expected and probably make us reluctant to install this version in Datumaro.. Have you analyzed this more? This is come from the OpenVINO problem?

sovrasov:
> > @wonjuleee @yunchu apparently, unit tests on macos were killed by timeout, because they stall at some kind of download step Is it a known issue? Unit testing on ubuntu (which probably download the same assets) was finished completed without errors image
> 
> Hi @sovrasov, this is not expected and probably make us reluctant to install this version in Datumaro.. Have you analyzed this more? This is come from the OpenVINO problem?

Unfortunately, this happens only on Mac platform, and can't be reproduced on Ubuntu. Do we have a Mac device to login and check?

yunchu:
@sovrasov now i'm looking into the failure of pr-test on the macos. it seemed to be stucked at the end of processing that download a model from the ov storage but not sure why it happened only on the macos as well as not happened on the develop with previous version of openvino (2023.1). 
finally, I bag your understanding that we don't have the macos machine that can be used for direct debugging.

yunchu:
@wonjuleee how about excluding the unittest "test_prune.py" from the macos target on this PR for further testing of the ov 2023.2 then i think we can address that issue separately after filing a bug.

yunchu:
@sovrasov there was misunderstanding about the stuck point in the past due to the log message flushing delays on the gh-actions job log display pane. I've checked that the testing was stucked in the call of "openvino.runtime.Core.read_model()" [here](https://github.com/openvinotoolkit/datumaro/blob/628287130cd9927135e0f9dc9360051abc78700f/src/datumaro/plugins/openvino_plugin/launcher.py#L202) by adding some dummy logging within the model downloading & infer flow.
this issue is only observed with the ov2023.2 on the macos-11.

sovrasov:
> @sovrasov there was misunderstanding about the stuck point in the past due to the log message flushing delays on the gh-actions job log display pane. I've checked that the testing was stucked in the call of "openvino.runtime.Core.read_model()" [here](https://github.com/openvinotoolkit/datumaro/blob/628287130cd9927135e0f9dc9360051abc78700f/src/datumaro/plugins/openvino_plugin/launcher.py#L202) by adding some dummy logging within the model downloading & infer flow. this issue is only observed with the ov2023.2 on the macos-11.

Thanks a lot for investigation! @wonjuleee since we don't have a mac device, can we just skip the problematic models? Apparently, the new OV has some bugs on Mac platform, but it'd be hard to make a reproducer for OV team given that we have only a gh actions executor.

yunchu:
FYI, @sovrasov @wonjuleee i've filed a issue #1219 for the hanging issue on the MacOS-11 with OpenVINO 2023.2 and updated TCs to be excluded from the run on macos target as workaround.

sovrasov:
@yunchu maybe we can move the builds to the next version of MacOS?
11.x was discontinued recently: https://endoflife.date/macos

sovrasov:
@vinnamkim please have a look as well

vinnamkim (approved):

PR #1205, opened 2023-11-27
Fix label compare of distance method

sooahleex:
### Summary

- Fix label comparison result of distance method
- Example: dataset1 included label ['a', 'b] and dataset2 included label ['background', 'a', 'b', 'c']
  - Expected Result :
    < c
    < background
  - Actual Result:
    a != background
    b != a
    < b
    < c
- Ticket no. 124879

### Checklist

### License

```python
#
```

### DIFF
--- a/src/datumaro/cli/util/compare.py
+++ b/src/datumaro/cli/util/compare.py
@@ -8,7 +8,6 @@ import os.path as osp
 import warnings
 from collections import Counter
 from enum import Enum, auto
-from itertools import zip_longest
 from typing import TYPE_CHECKING, Union
 
 import cv2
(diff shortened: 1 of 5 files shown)


src/datumaro/cli/util/compare.py
```diff
@@ -83,22 +82,24 @@ def save(self, a: IDataset, b: IDataset):
...
-        class_mismatch = [
-            (idx, a_cls, b_cls)
-            for idx, (a_cls, b_cls) in enumerate(zip_longest(a_classes, b_classes))
-            if getattr(a_cls, "name", None) != getattr(b_cls, "name", None)
+        a_classes = set(
+            a_cls.name for a_cls in a.categories().get(AnnotationType.label, LabelCategories())
+        )
+        b_classes = set(
+            b_cls.name for b_cls in b.categories().get(AnnotationType.label, LabelCategories())
+        )
```
vinnamkim: Would you please make it cleaner by introducing some interface in the `Category` class such as
```suggestion
        a_classes = set(a.categories().get_label_cat_names())
        b_classes = set(b.categories().get_label_cat_names())
```

src/datumaro/cli/util/compare.py
```diff
@@ -83,22 +82,24 @@ def save(self, a: IDataset, b: IDataset):
...
+
+        class_mismatch = [(diff, None) for diff in a_classes - b_classes] + [
```
vinnamkim: Is there any reason to keep `class_mismatch`? If not, we just do the following
```python
for idx, diff in enumerate(a_classes - b_classes):
     print(" #%s: > %s" % (idx, diff))
cnt = idx + 1
for idx, diff in enumerate(b_classes - a_classes):
     print(" #%s: < %s" % (idx + cnt, diff))
```

src/datumaro/cli/util/compare.py
jihyeonyi: F.Y.I., if you add the interface, it should be in the [`CategoriesInfo`](https://github.com/openvinotoolkit/datumaro/blob/44c1391abd7ef554b89615b422717833634174d7/src/datumaro/components/dataset_base.py#L58C1-L58C50), not in `Category`.
Anyway I don't think it is necessary to add this interface.
However it would be very useful if this is added, because we do this operation frequently.

src/datumaro/cli/util/compare.py
sooahleex: Fix this like
```python
for idx, diff in enumerate(a_classes - b_classes):
    print(" #%s: > %s" % (idx, diff))

for idx, diff in enumerate(b_classes - a_classes, start=len(a_classes)):
    print(" #%s: < %s" % (idx, diff))
```

src/datumaro/cli/util/compare.py
sooahleex: Fix this through define `get_label_cat_names` function in `Dataset`.

src/datumaro/cli/util/compare.py
vinnamkim: Are they different, `len(a_classes - b_classes)` and `len(a_classes)`?

sooahleex: Fix this in [b9538b0](https://github.com/openvinotoolkit/datumaro/pull/1205/commits/b9538b0125f1634c9f62c33b7e2b3a03fcac2f4b).

tests/integration/cli/test_compare.py
```diff
@@ -179,8 +181,9 @@ def test_can_compare_projects(self):  # just a smoke test
 
             self.assertNotEqual(0, os.listdir(osp.join(test_dir)))
 
+    @patch("sys.stdout", new_callable=io.StringIO)
     @mark_requirement(Requirements.DATUM_GENERAL_REQ)
-    def test_can_run_distance_diff(self):
+    def test_can_run_distance_diff(self, mock_stdout):
```
vinnamkim: Best practice for capturing `stdout` or `stderr` is using a special fixture provided by `pytest`. Please see https://pytest.org/en/7.4.x/how-to/capture-stdout-stderr.html#accessing-captured-output-from-a-test-function

vinnamkim (approved):
LGTM.

PR #1213, opened 2023-12-01
Add MMDetection COCO format importer

wonjuleee:
### Checklist

### License

```python
#
```


vinnamkim (approved):
LGTM. However, I have a feeling that we need to find a way to scale these COCO variants (like Roboflow) at the code level more efficiently. I guess that they have same file contents but different paths.

wonjuleee:
> LGTM. However, I have a feeling that we need to find a way to scale these COCO variants (like Roboflow) at the code level more efficiently. I guess that they have same file contents but different paths.

Thank you for the comment. I agree with you. We might be able to support similar directory structures through small modifications and relaxation. This is same for other famous public formats such as VOC and Cityscapes within mmdetection document https://mmdetection.readthedocs.io/en/latest/user_guides/dataset_prepare.html. Let me think about this more for support kaggle-like datasets in the same codebase.

PR #1223, opened 2023-12-13
[Bug] Fix Kinetics format to have media

wonjuleee:
### Checklist

### License

```python
#
```

### DIFF
--- a/src/datumaro/plugins/data_formats/kinetics.py
+++ b/src/datumaro/plugins/data_formats/kinetics.py
@@ -69,11 +69,9 @@ class KineticsBase(DatasetBase):
         self._subset_media_files[subset] = {}
         for root, _, files in os.walk(subset_path):
             for f in files:
-                file_path = osp.join(root, f)
-                key, file_extension = osp.splitext(osp.basename(file_path))
-                file_extension = file_extension.lstrip(".")
-                if file_extension in VIDEO_EXTENSIONS:
-                    self._subset_media_files[subset][key] = file_path
+                key, file_extension = osp.splitext(f)
+                if file_extension.lstrip(".") in VIDEO_EXTENSIONS:
+                    self._subset_media_files[subset][key] = osp.join(root, f)
 
         return self._subset_media_files[subset]
 
--- a/src/datumaro/plugins/data_formats/kinetics.py
+++ b/src/datumaro/plugins/data_formats/kinetics.py
@@ -65,11 +65,15 @@ class KineticsBase(DatasetBase):
             return self._subset_media_files[subset]
 
         subset_path = self._subset_path(subset)
-        self._subset_media_files[subset] = {
-            osp.splitext(osp.basename(f))[0]: osp.join(subset_path, f)
-            for f in os.listdir(subset_path)
-            if osp.splitext(osp.basename(f))[1] in VIDEO_EXTENSIONS
-        }
+
+        self._subset_media_files[subset] = {}
+        for root, _, files in os.walk(subset_path):
+            for f in files:
+                file_path = osp.join(root, f)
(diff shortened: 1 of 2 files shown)


src/datumaro/plugins/data_formats/kinetics.py
```diff
@@ -65,11 +65,15 @@ def _media_files(self, subset):
...
-            osp.splitext(osp.basename(f))[0]: osp.join(subset_path, f)
-            for f in os.listdir(subset_path)
-            if osp.splitext(osp.basename(f))[1] in VIDEO_EXTENSIONS
-        }
+
+        self._subset_media_files[subset] = {}
+        for root, _, files in os.walk(subset_path):
+            for f in files:
+                file_path = osp.join(root, f)
+                key, file_extension = osp.splitext(osp.basename(file_path))
```
vinnamkim: Is it just `osp.splitext(f)`?

src/datumaro/plugins/data_formats/kinetics.py
```diff
@@ -65,11 +65,15 @@ def _media_files(self, subset):
...
+                file_extension = file_extension.lstrip(".")
+                if file_extension in VIDEO_EXTENSIONS:
+                    self._subset_media_files[subset][key] = file_path
```
vinnamkim:
```suggestion
                key, file_extension = osp.splitext(f)
                if file_extension.lstrip(".") in VIDEO_EXTENSIONS:
                    self._subset_media_files[subset][key] = osp.join(root, f)
```

src/datumaro/plugins/data_formats/kinetics.py
jihyeonyi: Is it okay to use the base name(without extension) as key? I think it should be a relative path from `subset_path`, because there can be duplication.

src/datumaro/plugins/data_formats/kinetics.py
wonjuleee: Thank you. Fix at ebbdd433a23f4eb474ccddd9ef38540ab6c34f43.

src/datumaro/plugins/data_formats/kinetics.py
wonjuleee: fixed at ebbdd433a23f4eb474ccddd9ef38540ab6c34f43.

jihyeonyi: You used os.walk(), not os.listdir(). 
If there are files with same basename but different root folder, the value corresponding to the key will be overwritten.
I think this case is technically possible.
So, key should be relative path from subset_path, or some other distinguishable one.

wonjuleee: Sorry that I failed to understand the point. 
Kinetics directory structure is given by
```
subset
    category
        video_name_start_end.mp4
```
With this structure, I explore all files within a subset folder through `os.walk` and this brings us to `video_name_start_end` as a `key`. Are you worried the situation that there are duplicated `video_name_start_end` files between subsets, e.g., `train` and `val`?

jihyeonyi: You also use subset as key, I think that is distinguishable. 
Let's illustrate an example of a problematic scenario: when both 'category/video.mp4' and 'category/some_folder/video.mp4' exist together, or when 'category/video.mp4' and 'category/video.mkv' differ only in their file extensions.
If you're certain that such scenarios don't exist, there's no need to fix this.

wonjuleee: Thank you for your examples. I believe this fix is to enable to import Kinetics data and fortunately there is no mixed directory structure or video extensions within Kinetics.

jihyeonyi (approved):

PR #1224, opened 2023-12-13
Develop JsonSectionPageMapper in Rust API

vinnamkim:
### Summary

- Ticket no. 127135 and 127136.
- Develop `JsonSectionPageMapper` to construct page maps for top-level sections in a given JSON file.
- Enhance `DatumaroImporter.detect()`'s performance by replacing JSON file parsing logic with the `JsonSectionPageMapper`.

### How to test
Our existing test will validate its functionality. For the performance comparison, please see the following.

- Before
```python
from datumaro.rust_api import JsonSectionPageMapper
from time import time
import datumaro as dm

start = time()
format = dm.Dataset.detect("ws_test/coco/datumaro")
dt = 1000.0 * (time() - start)
print(f"Duration for detecting Datumaro data format: {dt:.1f}ms, format={format}")
```

```console
Duration for detecting Datumaro data format: 25784.5ms, format=datumaro
```

- After
```python
from datumaro.rust_api import JsonSectionPageMapper
from time import time
import datumaro as dm

start = time()
format = dm.Dataset.detect("ws_test/coco/datumaro")
dt = 1000.0 * (time() - start)
print(f"Duration for detecting Datumaro data format: {dt:.1f}ms, format={format}")
```

```console
Duration for detecting Datumaro data format: 17234.7ms, format=datumaro
```

It saves ~7 secs.

### Checklist

### License

```python
#
```

### DIFF
--- a/rust/src/json_section_page_mapper.rs
+++ b/rust/src/json_section_page_mapper.rs
@@ -1,0 +1,232 @@
+//  Copyright (C) 2023 Intel Corporation
+//
+//  SPDX-License-Identifier: MIT
+
+use crate::{
+    page_mapper::{JsonPageMapper, ParsedJsonSection},
+    utils::read_skipping_ws,
+};
+use pyo3::{prelude::*, types::PyDict};
+use std::{
+    collections::HashMap,
+    fs::File,
+    io::{self, BufReader, Read, Seek},
+    path::Path,
+};
+
+#[derive(Debug)]
+struct JsonSection {
+    key: String,
+    offset: usize,
+    size: usize,
+}
+
+fn handle_arr_or_dict(
+    mut stack: Vec<u8>,
+    mut reader: impl Read + Seek,
+    mut last_token: u8,
+) -> Result<(), io::Error> {
+    while stack.len() != 0 {
+        match read_skipping_ws(&mut reader) {
+            Ok(c) => match c {
+                b'{' | b'[' => {
+                    stack.push(c);
+                    last_token = c;
+                }
+                b'}' => {
+                    if last_token != b'{' {
+                        let cur_pos = reader.stream_position()?;
+                        let msg = format!("Last token in the stack is '{}', but the given token at offset={} is '}}'", last_token as char, cur_pos);
+                        return Err(io::Error::new(io::ErrorKind::InvalidData, msg));
+                    }
+                    stack.pop();
+                    if stack.len() != 0 {
+                        last_token = *stack
+                            .last()
+                            .ok_or(io::Error::new(io::ErrorKind::InvalidData, "stack is empty"))?;
+                    }
+                }
+                b']' => {
+                    if last_token != b'[' {
+                        let cur_pos = reader.stream_position()?;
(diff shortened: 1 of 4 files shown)


vinnamkim:
Unless we merge this PR, we cannot reach our first objective, memory bounded `detect()`. Please see the following memory profiling result with this PR:
after

This is because there are many data formats which read the whole JSON file at once during their `detect()`. This is one of them:
https://github.com/openvinotoolkit/datumaro/blob/4faaae505b849f3d280d13366fded37e9610afb6/src/datumaro/plugins/data_formats/segment_anything/importer.py#L17-L29

I'll fix the other data formats as well in the next PR.

rust/src/json_section_page_mapper.rs
```diff
@@ -0,0 +1,232 @@
...
+        Ok(dict.into_py(self_.py()))
+    }
+
+    fn __len__(&self) -> PyResult<usize> {
+        Ok(self.mapper.sections.len())
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
wonjuleee, line 204: shouldn't this be separated from src code? Is there any plan to introduce a test directory within rust?

src/datumaro/plugins/data_formats/datumaro/importer.py
```diff
@@ -28,9 +28,11 @@ def detect(
         with context.probe_text_file(
             annot_file,
             'must be a JSON object with "categories" ' 'and "items" keys',
-        ) as f:
-            contents = parse_json(f.read())
-            if not {"categories", "items"} <= contents.keys():
+        ):
+            fpath = osp.join(context.root_path, annot_file)
```
wonjuleee, line 32: Here, I personally love to get access to context.root_path directly, but this is recommended to avoid during detection. https://github.com/openvinotoolkit/datumaro/blob/4faaae505b849f3d280d13366fded37e9610afb6/src/datumaro/components/format_detection.py#L178-L179
Do you have any objection to use this during detection

rust/src/json_section_page_mapper.rs
vinnamkim: It is possible. I'll do in the future PR.

wonjuleee (approved):
It looks great to me in overall. Especially, it is happy to reduce the total time consumption during detection. Thank you!

src/datumaro/plugins/data_formats/datumaro/importer.py
vinnamkim: That guide is because the methods provided by `context` are using `self._root_path` internally. For example,
https://github.com/openvinotoolkit/datumaro/blob/4faaae505b849f3d280d13366fded37e9610afb6/src/datumaro/components/format_detection.py#L354-L357

However, `JsonSectionPageMapper` Rust API has no information for `root_path`. There is no choice but to inject the full path `osp.join(context.root_path, annot_file)` to access the file.

wonjuleee: Agree, how about other formats with pure Python codes? I just want to hear your opinion. 
Frankly speaking, I don't want to avoid the use of `context.root_path`.

Issue #1204, opened 2023-11-27
Statistics raise an error when there is a label annotation not in the category.


jihyeonyi:
I manually made a datumaro dataset which has 4 labels in the LabelCategories but has unknown label annotation(label=4).
When I tried to calculate annotation statistics, it raised an error like this:
![image](https://github.com/openvinotoolkit/datumaro/assets/75776702/9c396c30-ae67-4ef8-af70-fa7515b28c2b)

I think `compute_ann_statistics` and corresponding validator should be checked to fix this bug.

PR #1232, opened 2023-12-21
Handling undefined labels at the annotation statistics

jihyeonyi:
### Summary
Regarding the issue #1204 , modify `compute_ann_statistics` function to handle undefined labels.
And add corresponding unit test.

### Checklist

### License

```python
#
```


src/datumaro/components/operations.py
```diff
@@ -253,21 +258,26 @@ def get_label(ann):
     }
     label_stat = {
         "count": 0,
-        "distribution": {l.name: [0, 0] for l in labels.items},  # label -> (count, total%)
+        "distribution": defaultdict(lambda: [0, 0]),  # label -> (count, total%)
```
vinnamkim, line 267: If `[0,0]` does not change its length, please use `tuple` or `namedtuple`, Python list is less efficient for handling the fixed length data.

jihyeonyi: It make sense and I agree with you. However, I think we need to give users an appropriate notice before changing this, because someone may already use this output to his/her application.

vinnamkim (approved):
LGTM.

PR #1240, opened 2024-01-09
Add Kaggle data supports

wonjuleee:
### Checklist

### License

```python
#
```


src/datumaro/components/annotation.py
```diff
@@ -204,7 +204,8 @@ def add(
         attributes: Optional[Set[str]] = None,
     ) -> int:
         assert name
-        assert name not in self._indices, name
+        if name in self._indices:
```
vinnamkim: It seems to raise an error when an user is trying to add the duplicated label `name`. It will introduce a silent failure that an user cannot catch, so that can introduce an unintentional bug. What's the reason behind this change?

src/datumaro/plugins/data_formats/kaggle/base.py
```diff
@@ -0,0 +1,393 @@
...
+        for ind, row in df.iterrows():
+            if ind == 0:
+                continue
+
+            data_info = list(row)
+            media_name = data_info[indices["media"]]
+
+            if not media_name.lower().endswith(tuple(IMAGE_EXTENSIONS)):
+                for ext in IMAGE_EXTENSIONS:
+                    if osp.exists(os.path.join(self._path, media_name + ext)):
```
vinnamkim: Can we maintain this intermediate variable such as
`media_path = os.path.join(self._path, media_name + ext)`
It is being created three times in this code.

src/datumaro/plugins/data_formats/kaggle/base.py
```diff
@@ -0,0 +1,393 @@
...
+        label_cat = LabelCategories()
+
+        ids = []
+        items = []
+        with open(ann_file, "r", encoding="utf-8") as f:
+            for line in f:
+                line = line.split()
+                media_name = line[columns["media"]]
+
+                if not osp.exists(osp.join(self._path, media_name)):
```
vinnamkim: Why don't we append a file extension from `IMAGE_EXTENSIONS` this time?

src/datumaro/components/annotation.py
wonjuleee: Thank you for pointing out this. Most of Kaggle data doesn't have any additional file describing category information separately, and hence I should add labels when reading annotation file line by line (each line contains individual dataset item). I agree with the existence of salient failure. Let me add a warning message at here.

src/datumaro/plugins/data_formats/kaggle/base.py
wonjuleee: Will do. Thank you.

src/datumaro/plugins/data_formats/kaggle/base.py
wonjuleee: There is no case omitting an image extension in Kaggle data. But I think it would be good to automatic detection of image extension here too. I will update.

src/datumaro/components/annotation.py
vinnamkim: Although we need to expand the label category line by line during reading an annotation file, is it possible to keep the previous behavior? For example,

```python
label_category: LabelCategories

for line in file.read_lines():
     label_name: str = get_label(line)
     _, label = label_category.find(label_name)
     if label is None:
          label_cateogry.add(label_name)
```
It is reluctant to me changing drastically from the very strict check (assert) to the loose check (warning).

src/datumaro/plugins/data_formats/kaggle/base.py
wonjuleee: fixed at 0b13690c5f9a0ad5bcf55e3d6b3baf17ed84c7eb

src/datumaro/plugins/data_formats/kaggle/base.py
wonjuleee: fixed at 0b13690c5f9a0ad5bcf55e3d6b3baf17ed84c7eb

src/datumaro/components/annotation.py
wonjuleee: I modified this without change annotation at 0b13690c5f9a0ad5bcf55e3d6b3baf17ed84c7eb.
It may be the same approach with your idea.

vinnamkim (approved):
LGTM.

jihyeonyi (approved):
LGTM

Issue #1214, opened 2023-12-01
Import only import one source


AbdelsalamHaa:
System: 
Windows 11 with WSL 
python 3.8 

##Issue: 
importing the dataset imports only one source of the project, although it might have many sources. 

## Should be 
The import function should import the full project not only one source 

details 
here is my tree/config.yml file 
        `build_targets:
          project:
            parents: []
            stages:
            - hash: ''
              kind: ''
              name: root
              params: {}
              type: project
          task_113:
            parents: []
            stages:
            - hash: ''
              kind: ''
              name: root
              params: {}
              type: source
            - hash: ''
              kind: rename
              name: stage-1
              params:
                regex: '|frame_|task_113_'
              type: transform
            - hash: ''
              kind: random_split
              name: stage-2
              params:
                seed: null
                splits:
                - - train
                  - 0.7
                - - test
                  - 0.3
              type: transform
          task_117:
            parents: []
            stages:
            - hash: ''
              kind: ''
              name: root
              params: {}
              type: source
            - hash: ''
              kind: rename
              name: stage-1
              params:
                regex: '|frame_|task_117_'
              type: transform
            - hash: ''
              kind: random_split
              name: stage-2
              params:
                seed: null
                splits:
                - - train
                  - 0.7
                - - test
                  - 0.3
              type: transform
        format_version: 2
        sources:
          task_113:
            format: yolo
            hash: ''
            options: {}
            path: ''
            url: /home/iradar/aiv_pipeline/datasets/tuna/task_113
          task_117:
            format: yolo
            hash: ''
            options: {}
            path: ''
    url: /home/iradar/aiv_pipeline/datasets/tuna/task_117`

but only task 117 is imported 
 
My code to import
    `dataset_path = '/home/iradar/aiv_pipeline/datasets/tuna'

    dataset = Dataset.import_from(dataset_path, 'yolo') `

AbdelsalamHaa:
@vinnamkim 
I just would like to add,  when I used version 1.0.0, the import function worked well and everything seemed fine.

vinnamkim:
Hi @AbdelsalamHaa,
Sorry for the late response. It takes time to find the root cause for this. Until we solve this thing, you can merge your datasets manually by code. For example,

```python
from datumaro import Dataset, HLOps

dataset_1 = Dataset.import_from("/home/iradar/aiv_pipeline/datasets/tuna/task_113", 'yolo')
[...]

PR #1243, opened 2024-01-15
Fix a bug in the previous behavior when importing nested datasets in the project

vinnamkim:
### Summary

- Ticket no. 127589
- Fix a bug found in #1214
- In the previous Datumaro version, we could import the nested datasets in the given path. For example, if we `Dataset.import_from(./some_project)` for the following directory structure
``` console
./some_project/
├── dataset_1
└── dataset_2
```
, the imported dataset include `dataset_1` and `dataset_2`.
- This is common pattern for the Datumaro project, so that we have to fix this.

### How to test
Added an integration test for this scenario.

### Checklist

### License

```python
#
```

### DIFF
--- a/tests/integration/cli/test_yolo_format.py
+++ b/tests/integration/cli/test_yolo_format.py
@@ -215,7 +215,7 @@ class YoloIntegrationScenariosTest:
     def fxt_yolo_dir(self, request) -> str:
         return get_test_asset_path("yolo_dataset", request.param)
 
-    @mark_requirement(Requirements.DATUM_BUG_1204)
+    @mark_requirement(Requirements.DATUM_BUG_1214)
     def test_can_import_nested_datasets_in_project(self, fxt_yolo_dir, test_dir, helper_tc):
         run(helper_tc, "project", "create", "-o", test_dir)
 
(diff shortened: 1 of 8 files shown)


tests/integration/cli/test_yolo_format.py
```diff
@@ -202,3 +207,53 @@ def test_can_delete_labels_from_yolo_dataset(self):
...
             compare_datasets(self, target_dataset, parsed_dataset)
+
+
+# TODO(vinnamki): Migrate above test cases to the pytest framework below
+class YoloIntegrationScenariosTest:
+    @pytest.fixture(params=["annotations", "labels", "strict"])
+    def fxt_yolo_dir(self, request) -> str:
+        return get_test_asset_path("yolo_dataset", request.param)
+
+    @mark_requirement(Requirements.DATUM_BUG_1204)
```
wonjuleee: maybe `DATUM_BUG_1214`?

vinnamkim: Fixed in e0d266b75c384afef62dd08a61d8cbac7a0af8d6

wonjuleee (approved):
In overall, it looks good to me, but I have a one quick question. Is this fix available only for coco and yolo formats?

vinnamkim:
> In overall, it looks good to me, but I have a one quick question. Is this fix available only for coco and yolo formats?

As we discussed, it is available only for the yolo formats. This is because importing datasets from sub-directories at once is not commonly available for all data formats.

PR #1247, opened 2024-01-22
Implement dm.Dataset index access

itrushkin:
Issue No: 126681
### Summary

This PR introduces dataset random index access for the easy conversion from `dm.Dataset` to `torch.Dataset`. The implementation requires an additional list property for `DatasetItemStorage` class, which would cost additional O(n) memory. It doesn't store items but references to them (subset and ID). 

Concern:
I wonder whether it is possible to store and manage this list for indexing, removal, addition, and iteration to reduce memory consumption. I tried to combine the following changes:
1) The `_traversal_order` dictionary seems unnecessary as all data is already stored and managed in the `data` property via `.put()` and `.remove()` methods. `__len__`, `__iter__`, and other methods can rely solely on `data`. Optionally, an `_order` property could enable O(1) `__len__` implementation.
2) Instead of overwriting `data` values with `None` to mark removal, deleting the entire dictionary entry accurately reflects the dataset state and eliminates ambiguity.

As a result, tests set `test_can_create_patch*` started to fail. It appears that items were not actually removed from the dataset, but merely marked as such. This suggests either the cache initialization or element remove operation needs adjusting for this scenario. Resolving this with assistance could significantly improve `dm.Dataset` interaction.

### Checklist

### License

```python
#
```

### DIFF
--- a/src/datumaro/components/dataset.py
+++ b/src/datumaro/components/dataset.py
@@ -932,6 +932,9 @@ class Dataset(IDataset):
         """
         return deepcopy(self)
 
+    def __getitem__(self, idx: int) -> DatasetItem:
+        return self._data[idx]
+
 
 class StreamDataset(Dataset):
     _stream = True
--- a/src/datumaro/components/dataset_item_storage.py
+++ b/src/datumaro/components/dataset_item_storage.py
@@ -106,7 +106,7 @@ class DatasetItemStorage:
         copied._order = copy(self._order)
         copied.data = copy(self.data)
         return copied
-    
+
     def __getitem__(self, idx: int) -> DatasetItem:
         _id, subset = self._order[idx]
         item = self.data[subset][_id]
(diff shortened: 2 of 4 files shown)


vinnamkim:
Hi @itrushkin,

Could we add a test for this feature about the scenarios where the size of dataset increases or decreases? For example,

a) Scenario with `dm.Dataset.put()` and `dm.Dataset.remove()`.

1. Construct an arbitrary `dm.Dataset`
2. Get an item from the dataset with index
3. Increase the dataset size with `dm.Dataset.put` or decrease it with `dm.Dataset.put` couple of times.
4. Get an item from the dataset with index and check whether it behaves correctly

b) Scenario with transforms

1. Construct an arbitrary `dm.Dataset`
2. Get an item from the dataset with index
3. Increase or decrease the dataset size with Datumaro transforms (I guess you might use tile/untile transforms for both https://github.com/openvinotoolkit/datumaro/tree/develop/src/datumaro/plugins/tiling).
4. Get an item from the dataset with index and check whether it behaves correctly

vinnamkim (approved):
LGTM.

PR #1246, opened 2024-01-19
Enable image backend and color channel format to be selectable

vinnamkim:
### Summary

- Previously we enabled the image decoding context managing only for both 3-channel colors or unchanged. However, we should extend it to image backend; and RGB or BGR color channel format selection for OTX use cases.

### How to test
I updated the existing unit tests to cover this change as well.

### Checklist

- [x] I have added unit tests to cover my changes.​
- [ ] I have added integration tests to cover my changes.​
- [x] I have added the description of my changes into [CHANGELOG](https://github.com/openvinotoolkit/datumaro/blob/develop/CHANGELOG.md).​
- [ ] I have updated the [documentation](https://github.com/openvinotoolkit/datumaro/tree/develop/docs) accordingly

### License

- [x] I submit _my code changes_ under the same [MIT License](https://github.com/openvinotoolkit/datumaro/blob/develop/LICENSE) that covers the project.
  Feel free to contact the maintainers if that's a concern.
- [x] I have updated the license header for each file (see an example below).

```python
# Copyright (C) 2023 Intel Corporation
#
# SPDX-License-Identifier: MIT
```

PR #1259, opened 2024-02-04
Prioritize using imagesize library to get image size for ImageFromFile

imyhxy:
### Summary

Accelerate loading of image file-based datasets.

I found that printing out the YOLO dataset information for the first time was slow. After some digging I found that `datamaro` was reading the entire dataset through to get the size of each image.

```python
ds = Dataset.import_from("/yolo-ultralytics", "yolo")
print(ds)  # <-- wait a long time
```

```python
    # from class Image
    @property
    def size(self) -> Optional[Tuple[int, int]]:
        """Returns (H, W)"""

        if self._size is None:
            try:
                data = self.data  # <-- load the whole media into memory
            except _image_loading_errors:
                return None
            if data is not None:
                self._size = tuple(map(int, data.shape[:2]))
        return self._size
```

Interactive encoding with datasets on HDD is slow. So I added an override `size()` property in the `ImageFromFile` class which first tries to get the image size using `PIL`. The `PIL` library is about 8 times faster than `OpenCV` in getting the image size.

All dataset classes that use the `size` property of `ImageFromFile` can benefit from this modification.

### Checklist

### License

```python
#
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,4 +2,3 @@
 -r requirements-default.txt
 
 opencv-python-headless>=4.1.0.25
-imagesize>=1.4.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,3 +2,4 @@
 -r requirements-default.txt
 
 opencv-python-headless>=4.1.0.25
+imagesize>=1.4.1
--- a/src/datumaro/components/media.py
+++ b/src/datumaro/components/media.py
@@ -337,14 +337,18 @@ class ImageFromFile(FromFileMixin, Image):
         return data
 
     @property
+import imagesize
+
+...
+
     def size(self) -> Optional[Tuple[int, int]]:
         """Returns (H, W)"""
 
         if self._size is None:
-            if _HAS_PIL:
-                w, h = PILImage.open(self.path).size
-                self._size = (h, w)
-            else:
+            try:
+                width, height = imagesize.get(self.path)
+                self._size = (height, width)
+            except Exception:
                 _ = super().size
         return self._size
 
--- a/src/datumaro/components/media.py
+++ b/src/datumaro/components/media.py
@@ -49,6 +49,12 @@ else:
 
     pd = lazy_import("pandas")
 
+try:
+    from PIL import Image as PILImage
+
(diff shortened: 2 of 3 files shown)


src/datumaro/components/media.py
```diff
@@ -330,6 +336,18 @@ def data(self) -> Optional[np.ndarray]:
...
+    def size(self) -> Optional[Tuple[int, int]]:
+        """Returns (H, W)"""
+
+        if self._size is None:
+            if _HAS_PIL:
+                w, h = PILImage.open(self.path).size
+                self._size = (h, w)
+            else:
+                _ = super().size
+        return self._size
```
vinnamkim: Hi @imyhxy,

Thanks for your good suggestion. However, in this PR https://github.com/openvinotoolkit/datumaro/pull/1246, we changed Datumaro image backend as selectable by the context variables. In addition, it would be good to extract a image size from the file header using the external library, e.g. https://pypi.org/project/imagesize,

```suggestion
import imagesize

...

    def size(self) -> Optional[Tuple[int, int]]:
        """Returns (H, W)"""

        if self._size is None:
            try:
                width, height = imagesize.get(self.path)
                self._size = (height, width)
            except Exception:
                _ = super().size
        return self._size
```

Would you please update your PR in this way? You might have to add that library to our `requirements.txt`.

imyhxy:
@vinnamkim 
It would be great if adding new dependencies was acceptable. I actually considered several approaches [here](https://stackoverflow.com/questions/15800704/get-image-size-without-loading-image-into-memory) and settled on `PIL` for compatibility. But yes, `imagesize` is a little faster.

`imagesize` vs `cv2`: 0.3s vs 8.1s on my benchmark dataset.

requirements.txt
```diff
@@ -2,3 +2,4 @@
 -r requirements-default.txt
 
 opencv-python-headless>=4.1.0.25
+imagesize>=1.4.1
```
vinnamkim: Sorry for my confusion. You have to move this to `requirements-core.txt`.

vinnamkim (approved):
LGTM

PR #1261, opened 2024-02-06
Boost segmentation import performance

wonjuleee:
### Summary

When analyzing the import performance for `cityscapes` and `kaggle_image_mask`, I have checked that the most bottleneck is `np.unique` for parsing the unique class indices within each mask.
Analysis before PR:
image

Analysis after PR:
image

Instead of parsing unique class indices within a mask, I have changed to use all class indices in a dataset.
As a result, the performance is 5 times faster.

### Checklist

### License

```python
#
```


jihyeonyi:
Are you going to fix the unit-test errors? It seems that the errors are not directly related to this PR. Your code change looks good to me.

wonjuleee:
> Are you going to fix the unit-test errors? It seems that the errors are not directly related to this PR. Your code change looks good to me.

The unit tests are broken because of checking `is_crowd` logic in https://github.com/openvinotoolkit/datumaro/blob/76fc9416e902ccd4e8bb364ed0d1cc2caef47785/src/datumaro/plugins/data_formats/cityscapes.py#L276-L283.

From my investigation, this is not defined in the official cityscapes document https://www.cityscapes-dataset.com/dataset-overview/#class-definitions, while this is for support CVAT cityscapes format as described in https://opencv.github.io/cvat/docs/manual/advanced/formats/format-cityscapes/. 
So, we need to decide whether drop this logic in Datumaro or not.
What do you think @openvinotoolkit/datumaro-maintainers?

src/datumaro/plugins/anchor_generator.py
```diff
@@ -183,7 +183,7 @@ def get_anchors(
                 x_center + 0.5 * ws,
                 y_center + 0.5 * hs,
             ]
-            base_anchors = torch.stack(base_anchors, dim=-1)
+            base_anchors = torch.stack(base_anchors, dim=-1).to(self.device)
```
jihyeonyi: Is this related to the purpose of this PR?

jihyeonyi:
> > Are you going to fix the unit-test errors? It seems that the errors are not directly related to this PR. Your code change looks good to me.
> 
> The unit tests are broken because of checking `is_crowd` logic in
> 
> https://github.com/openvinotoolkit/datumaro/blob/76fc9416e902ccd4e8bb364ed0d1cc2caef47785/src/datumaro/plugins/data_formats/cityscapes.py#L276-L283
> 
> .
> From my investigation, this is not defined in the official cityscapes document https://www.cityscapes-dataset.com/dataset-overview/#class-definitions, while this is for support CVAT cityscapes format as described in https://opencv.github.io/cvat/docs/manual/advanced/formats/format-cityscapes/. So, we need to decide whether drop this logic in Datumaro or not. What do you think @openvinotoolkit/datumaro-maintainers?

How about marking skip or xfail for this test and investigate this further later?
It is because we could miss something.

src/datumaro/plugins/anchor_generator.py
wonjuleee: Nope, this is to fix anchor generator.. I will make a separate PR for fixing this. Thank you for pointing out this.

wonjuleee:
> > > Are you going to fix the unit-test errors? It seems that the errors are not directly related to this PR. Your code change looks good to me.
> > 
> > 
> > The unit tests are broken because of checking `is_crowd` logic in
> > https://github.com/openvinotoolkit/datumaro/blob/76fc9416e902ccd4e8bb364ed0d1cc2caef47785/src/datumaro/plugins/data_formats/cityscapes.py#L276-L283
> > 
> > .
> > From my investigation, this is not defined in the official cityscapes document https://www.cityscapes-dataset.com/dataset-overview/#class-definitions, while this is for support CVAT cityscapes format as described in https://opencv.github.io/cvat/docs/manual/advanced/formats/format-cityscapes/. So, we need to decide whether drop this logic in Datumaro or not. What do you think @openvinotoolkit/datumaro-maintainers?
> 
> How about marking skip or xfail for this test and investigate this further later? It is because we could miss something.

I have fixed unit tests within cityscapes. Please review again :)

jihyeonyi (approved):
LGTM

PR #1263, opened 2024-02-06
Fix RISE algorithm for explain function

wonjuleee:
### Checklist

### License

```python
#
```


src/datumaro/components/algorithms/rise.py
```diff
@@ -34,186 +25,104 @@ class RISE:
...
-        mhmaps = _flatmatvec(heatmaps)
-        mhmaps /= _expand(counts * self.prob + eps)
-        mhmaps -= _expand(np.min(mhmaps, axis=1))
-        mhmaps /= _expand(np.max(mhmaps, axis=1) + eps)
-        return np.reshape(mhmaps, heatmaps.shape)
+    def normalize_saliency(self, saliency):
+        normalized_saliency = np.empty_like(saliency)
+        for idx, sal in enumerate(saliency):
+            normalized_saliency[idx, ...] = (sal - np.min(sal)) / (np.max(sal) - np.min(sal))
+        return normalized_saliency
```
sooahleex, line 44:
```suggestion
        return (saliency - np.min(saliency)) / (np.max(saliency) - np.min(saliency))
```
I think this could be replaced like this through vectorized operation.

src/datumaro/plugins/openvino_plugin/samples/googlenet-v4-tf_interp.py
```diff
@@ -29,11 +31,20 @@ def preprocess(self, inp: DatasetItem) -> Tuple[LauncherInputType, PrepInfo]:
         return img, None
 
     def postprocess(self, pred: ModelPred, info: PrepInfo) -> List[Annotation]:
+        logit = pred.get(self.LOGIT_KEY)
+        if logit is None:
+            raise DatumaroError(f'"{self.LOGIT_KEY}" key should exist in the model prediction.')
```
sooahleex, line 36:
```suggestion
        logit = pred.get(self.LOGIT_KEY, [])
        if not logit:
            raise DatumaroError(f'"{self.LOGIT_KEY}" key should exist in the model prediction.')
```
It seems better to use default values for safer.

src/datumaro/plugins/openvino_plugin/samples/googlenet-v4-tf_interp.py
```diff
@@ -29,11 +31,20 @@ def preprocess(self, inp: DatasetItem) -> Tuple[LauncherInputType, PrepInfo]:
...
+
         feature_vector = pred.get(self.FEAT_KEY)
         if feature_vector is None:
             raise DatumaroError(f'"{self.FEAT_KEY}" key should exist in the model prediction.')
```
sooahleex, line 40:
```suggestion
        feature_vector = pred.get(self.FEAT_KEY, [])
        if not feature_vector:
            raise DatumaroError(f'"{self.FEAT_KEY}" key should exist in the model prediction.')
```

sooahleex (approved):
Looks good to me. I left some minor comments.

src/datumaro/components/algorithms/rise.py
wonjuleee: thank you for the feedback, but the original code is to normalize each saliency map per class_idx, while your suggestion is to normalize each saliency map from global max and min. So I expect the results will be different.

src/datumaro/plugins/openvino_plugin/samples/googlenet-v4-tf_interp.py
wonjuleee: thank you. let me update this later.

PR #1282, opened 2024-02-29
Update dockerfile to resolve remained pin-deps issue

yunchu:
### Summary

Resolve issue - https://github.com/openvinotoolkit/datumaro/security/code-scanning/13

### Checklist

### License

```python
#
```


wonjuleee (commented):
@vinnamkim, could you review this?

docker/segment-anything/Dockerfile.exporter
```diff
@@ -9,13 +9,21 @@ ARG no_proxy
...
 
 WORKDIR /ws
+
 RUN mkdir -p /ws/output
-RUN pip install -e git+https://github.com/facebookresearch/segment-anything/@6fdee8f2727f4506cfbbe553e23b895e27956588#egg=segment_anything
+# install segment-anything with pinned hash
+RUN pip install --no-deps -e git+https://github.com/facebookresearch/segment-anything/@6fdee8f2727f4506cfbbe553e23b895e27956588#egg=segment_anything
 WORKDIR /ws/src/segment-anything
+# generate requirements file for sagment-anything from setup.py with hashes
+RUN pip-compile --extra all,dev --generate-hashes --output-file=/tmp/requirements/requirements.txt
```
vinnamkim: Why are we doing this if we are going to create a pinned dependency list at runtime?

yunchu: @vinnamkim can you elaborate about the "runtime"? does it mean the docker image build time?

According to [this security risk reporting](https://github.com/openvinotoolkit/datumaro/security/code-scanning/13) by OSSF scorecard, we need to remedy the issue reported on `segment-anything` installation step in the `Dockerfile.exporter`. pip provides an option '--require-hashes' to the installation command for the secure installation but it cannot be used with editable mode that is used for installation of `segment-anything` module to the docker image.
The only way to resolve this issue is adding `--no-deps` to the `segment-anything` installation command with editable option now. if using `--no-deps` option, it will not install dependencies so i've added `pip-compile` stage to generate pinned dependencies for the `segment-anything` separately.

vinnamkim: > The only way to resolve this issue is adding --no-deps to the segment-anything installation command with editable option now. if using --no-deps option, it will not install dependencies so i've added pip-compile stage to generate pinned dependencies for the segment-anything separately.

AFAIK, the reason why we are doing this is "to pin" dependencies. However, you execute `pip-compile` at runtime (building docker image). It means that the `requirements.txt` file is generated online (yes it includes the hash numbers but they are actually generated online). This seems that you are not actually "pinning" dependencies.

yunchu: @vinnamkim I thought those are the indirect dependencies to the docker image so i've pinned only the `segment-anything` with git commit hash. but as you pointed out, it's better to pinning all dependencies through pre-generated requirements file. I will upload new patch soon. thank you.

vinnamkim (approved):

PR #1327, opened 2024-03-11
Fix ContextVar into DatumaroApp

sooahleex:
### Summary

- `LookupError: <ContextVar name='IMAGE_BACKEND' at 0x7f528dd63240>` occurred in `load_image` function.
- Update `ImageBackend` in default page of streamlit
- Set `Image_Backend` in each component

### How to test

- Add `init_func` in each apptest of unit test

### Checklist

### License

```python
#
```

### DIFF
--- a/gui/datumaro_gui/components/multiple/streamlit_app.py
+++ b/gui/datumaro_gui/components/multiple/streamlit_app.py
@@ -18,6 +18,7 @@ from datumaro_gui.utils.dataset.state import (
     save_dataset,
 )
 from datumaro_gui.utils.drawing.css import box_style, custom_css
+from datumaro_gui.utils.page import check_image_backend
 from datumaro_gui.utils.readme import github_pypi_desc
 from streamlit import session_state as state
 
@@ -25,6 +26,7 @@ from . import tabs
 
 
 def main():
+    check_image_backend(st.session_state.get("IMAGE_BACKEND"))
     log.info("__multiple__.main is called")
     st.write(github_pypi_desc)
     st.markdown(custom_css, unsafe_allow_html=True)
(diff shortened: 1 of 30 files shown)


gui/datumaro_gui/components/multiple/streamlit_app.py
```diff
@@ -18,13 +18,15 @@
...
 )
 from datumaro_gui.utils.drawing.css import box_style, custom_css
+from datumaro_gui.utils.page import check_image_backend
 from datumaro_gui.utils.readme import github_pypi_desc
 from streamlit import session_state as state
 
 from . import tabs
 
 
 def main():
```
vinnamkim, line 28:
```suggestion
# Copy context vars somewhere close to initialization
from contextvars import copy_context
st.session_state["context_vars"] = copy_context()

...

from functools import wraps

def update_context_vars(func):
    @wraps(func):
    def wrapped(*args, **kwargs):
        if ctx := st.session_state.get("context_vars", None):
            for val, value in ctx.items():
                val.set(value)
        else:
            msg = "You should copy context variables to streamlit session_state first."
            logger.error(msg)
        return func(*args, **kwargs)
        
    return wrapped

@update_context_vars
def main():
```
Revise it to cover all context variables and consider to make it as a function decorator for a cleaner code pattern.

yunchu:
@sooahleex please re-target to the releases/2.0.0 for this PR if this change is required for the next release.

gui/datumaro_gui/components/multiple/streamlit_app.py
sooahleex: I will cover this in following PR.

wonjuleee (approved):

PR #1360, opened 2024-03-19
Enable coco format to import bbox annotations

wonjuleee:
### Checklist

### License

```python
#
```

### DIFF
--- a/src/datumaro/plugins/data_formats/coco/base.py
+++ b/src/datumaro/plugins/data_formats/coco/base.py
@@ -29,7 +29,6 @@ from datumaro.components.errors import (
     DatasetImportError,
     InvalidAnnotationError,
     InvalidFieldTypeError,
-    MissingFieldError,
     UndeclaredLabelError,
 )
 from datumaro.components.importer import ImportContext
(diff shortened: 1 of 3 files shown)


src/datumaro/plugins/data_formats/coco/base.py
```diff
@@ -450,7 +449,8 @@ def _parse_field(
     ) -> Any:
         value = ann.get(key, NOTSET)
         if value is NOTSET:
-            raise MissingFieldError(key)
+            log.warning(f"field '{key}' is not existed in the annotation file.")
+            return None
```
vinnamkim: I think that just removing this error raise is not appropriate. It can be a problem for those mandatory fields

https://github.com/openvinotoolkit/datumaro/blob/183a69e1ac8310ba9d8c4fec5f22067328fa1fa6/src/datumaro/plugins/data_formats/coco/base.py#L465
https://github.com/openvinotoolkit/datumaro/blob/183a69e1ac8310ba9d8c4fec5f22067328fa1fa6/src/datumaro/plugins/data_formats/coco/base.py#L480

They are not only mandatory, but also have no handling logic if `None` is given. This can introduce a bug which is not easily understandable if the malformed JSON input is given.

wonjuleee: Instead, I have updated by making only "segmentation" optional. Could you review again?

vinnamkim (approved):

PR #1362, opened 2024-03-20
Add DatasetImportError for yolo variants

wonjuleee:
### Summary

Datumaro is able to import a dataset from extractor directly as described in https://github.com/openvinotoolkit/datumaro/blob/f9366173a0a5ba6fe479703800a2f3a0caf15530/src/datumaro/components/dataset.py#L822
However, in this case, we have to be careful for importing a dataset.
For instance, Yolo extractors, i.e., `YoloLooseBase`, `YoloUltralyticsBase`, `RoboflowYoloBase`, and `RoboflowYoloObbBase`, require to have `urls` for specifying subset and all annotation files accordingly.

In this PR, I have added `DatasetImportError` when importing a Yolo dataset without `urls`.

### Checklist

### License

```python
#
```

### DIFF
--- a/src/datumaro/plugins/data_formats/yolo/base.py
+++ b/src/datumaro/plugins/data_formats/yolo/base.py
@@ -316,6 +316,9 @@ class YoloLooseBase(SubsetBase):
         if not osp.isdir(config_path):
             raise DatasetImportError(f"{config_path} should be a directory.")
 
+        if not urls:
+            raise DatasetImportError(f"urls should be specified for {self.__class__.__name__}.")
+
         rootpath = self._get_rootpath(config_path)
 
         self._image_info = YoloStrictBase.parse_image_info(rootpath, image_info)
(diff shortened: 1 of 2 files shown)


src/datumaro/plugins/data_formats/yolo/base.py
```diff
@@ -316,6 +316,9 @@ def __init__(
         if not osp.isdir(config_path):
             raise DatasetImportError(f"{config_path} should be a directory.")
 
+        if not urls:
+            raise DatasetImportError(f"urls should be specified for {self.__class__.__name__}.")
```
vinnamkim: Could it be more descriptive such as
```suggestion
            raise DatasetImportError(
                f"`urls` should be specified for {self.__class__.__name__}, "
                f"if you want to import a dataset with using this {self.__class__.__name__} directly. "
                "In most case, it happens by giving an incorrect format name to the import interface. "
                'Please consider to import your dataset with this format name, "yolo", such as `Dataset.import_from(..., format="yolo")`'
            )
```

vinnamkim (approved):

PR #1409, opened 2024-04-04
Add ignore_index argument to Mask.as_class_mask() and Mask.as_instance_mask()

vinnamkim:
### Summary

- Ticket no. 137625
- Same as title.

### How to test
Added an unit test for this change.

### Checklist

### License

```python
#
```

### DIFF
--- a/src/datumaro/components/annotation.py
+++ b/src/datumaro/components/annotation.py
@@ -370,23 +370,54 @@ class Mask(Annotation):
             image = image()
         return image
 
-    def as_class_mask(self, label_id: Optional[int] = None) -> IndexMaskImage:
-        """
-        Produces a class index mask. Mask label id can be changed.
+    def as_class_mask(
+        self,
+        label_id: Optional[int] = None,
+        ignore_index: int = 0,
+        dtype: Optional[np.dtype] = None,
+    ) -> IndexMaskImage:
+        """Produces a class index mask based on the binary mask.
+
+        Args:
+            label_id: Scalar value to represent the class index of the mask.
+                If not specified, `self.label` will be used. Defaults to None.
+            ignore_index: Scalar value to fill in the zeros in the binary mask.
+                Defaults to 0.
+            dtype: Data type for the resulting mask. If not specified,
+                it will be inferred from the provided `label_id`. Defaults to None.
+
+        Returns:
+            IndexMaskImage: Class index mask generated from the binary mask.
         """
         if label_id is None:
             label_id = self.label
         from datumaro.util.mask_tools import make_index_mask
 
-        return make_index_mask(self.image, label_id)
+        return make_index_mask(self.image, index=label_id, ignore_index=ignore_index, dtype=dtype)
 
-    def as_instance_mask(self, instance_id: int) -> IndexMaskImage:
-        """
-        Produces a instance index mask.
+    def as_instance_mask(
+        self,
+        instance_id: int,
+        ignore_index: int = 0,
+        dtype: Optional[np.dtype] = None,
+    ) -> IndexMaskImage:
+        """Produces an instance index mask based on the binary mask.
+
+        Args:
+            instance_id: Scalar value to represent the instance id.
+            ignore_index: Scalar value to fill in the zeros in the binary mask.
+                Defaults to 0.
+            dtype: Data type for the resulting mask. If not specified,
+                it will be inferred from the provided `label_id`. Defaults to None.
+
+        Returns:
+            IndexMaskImage: Instance index mask generated from the binary mask.
         """
         from datumaro.util.mask_tools import make_index_mask
 
-        return make_index_mask(self.image, instance_id)
+        return make_index_mask(
+            self.image, index=instance_id, ignore_index=ignore_index, dtype=dtype
+        )
 
     def get_area(self) -> int:
         return np.count_nonzero(self.image)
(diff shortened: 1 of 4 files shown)


src/datumaro/components/annotation.py
```diff
@@ -370,23 +370,54 @@ def image(self) -> BinaryMaskImage:
...
+    ) -> IndexMaskImage:
+        """Produces a class index mask based on the binary mask.
+
+        Args:
+            label_id: Scalar value to represent the class index of the mask.
+                If not specified, `self.label` will be used. Defaults to None.
+            ignore_index: Scalar value to fill in the zeros in the binary mask.
+                Defaults to 0.
+            dtype: Data type for the resulting mask. If not specified,
+                it will be inferred from the provided `label_id`. Defaults to None.
```
wonjuleee: This description looks like insufficient for me. Does this mean `np.uint8` would be default as `label_id` is integer generally?

src/datumaro/util/mask_tools.py
```diff
@@ -124,8 +124,54 @@ def remap_mask(mask, map_fn):
...
+        >>> binary_mask = np.eye(2, dtype=np.bool_)
+        >>> index_mask = make_index_mask(binary_mask, index=10, ignore_index=255, dtype=np.uint8)
+        >>> print(index_mask)
+        array([[ 10, 255],
+               [255,  10]], dtype=uint8)
+    """
+    if dtype is None:
+        dtype = np.min_scalar_type(index)
+        if dtype != np.min_scalar_type(ignore_index):
+            raise ValueError()
```
wonjuleee: Looks good overall, but can we provide more messages for this value error?

src/datumaro/util/mask_tools.py
```diff
@@ -124,8 +124,54 @@ def remap_mask(mask, map_fn):
...
+
+    flipped_zero_np_scalar = ~np.full(tuple(), fill_value=0, dtype=dtype)
+
+    # NOTE: This dispatching rule is required for a performance boost
+    if ignore_index == flipped_zero_np_scalar:
+        flipped_index = ~np.full(tuple(), fill_value=index, dtype=dtype)
+        return ~(binary_mask * flipped_index)
```
wonjuleee, line 174: Sorry, I was failed to understand this logic correctly. When `ignore_index` is 255, this returns `binary_mask * np.full(tuple(), fill_value=index, dtype=dtype)`. Is this right?
There is two negative `~` operations. Should we have this?
We have to explanation what `ignore_index==255` stands for.

src/datumaro/components/annotation.py
vinnamkim: https://github.com/openvinotoolkit/datumaro/pull/1409/commits/53b3a5aa02d3e1062a00e68e2e8058de18c0a62d improved it

src/datumaro/util/mask_tools.py
vinnamkim: https://github.com/openvinotoolkit/datumaro/pull/1409/commits/53b3a5aa02d3e1062a00e68e2e8058de18c0a62d did it

src/datumaro/util/mask_tools.py
vinnamkim: Please see this example
https://github.com/openvinotoolkit/datumaro/blob/53b3a5aa02d3e1062a00e68e2e8058de18c0a62d/src/datumaro/util/mask_tools.py#L152-L157

We want to convert a binary mask to an index mask by filling `1`s with `index` and `0`s with `ignore_index`. This function would be a simple function like
```python
return np.where(binary_mask, mask, ignore_index)
```
, but this function has if-else branching in the loop. Thus, it is hard to be accelerated by hardware.

`~` is a bit-wise flipping operator. Therefore, this logic is doing (`~255 == 0`)
```python
[~10, 0]
[0, ~10]

=>

[10, 255]
[255, 10]
```

Anyway, https://github.com/openvinotoolkit/datumaro/pull/1409/commits/53b3a5aa02d3e1062a00e68e2e8058de18c0a62d improved it further to cover the general cases.

vinnamkim:
```python
def kernel4(binary_mask, index, ignore_index=0, dtype=None):
    if dtype is None:
        dtype = np.min_scalar_type(index)
        if dtype != np.min_scalar_type(ignore_index):
            raise ValueError()

    mask = binary_mask * np.array([index], dtype=dtype)

    if ignore_index == 0:
        return mask

    return np.where(binary_mask, mask, ignore_index)

from timeit import timeit

binary_mask = np.random.randint(0, 2, size=[1024, 1024], dtype=np.bool_)

print("Kernel4 time: ", timeit(lambda: kernel4(binary_mask, index=10, ignore_index=230), number=200))
print("make_index_mask time: ", timeit(lambda: make_index_mask(binary_mask, index=10, ignore_index=230), number=200))
```
There was a huge performance gap with/without this optimization in my desktop
```console
Kernel4 time:  0.8781529648695141
make_index_mask time:  0.03690833714790642
```

wonjuleee (approved):
Could you update changelog to add this in 2.0 (will 1.6.0)?

vinnamkim:
> Could you update changelog to add this in 2.0 (will 1.6.0)?

Can we do the update after merging https://github.com/openvinotoolkit/datumaro/pull/1419#pullrequestreview-1985669024 and branching out `1.6.0`?

PR #1422, opened 2024-04-09
Add task type information when importing

wonjuleee:
### Checklist

### License

```python
#
```

### DIFF
--- a/src/datumaro/components/dataset.py
+++ b/src/datumaro/components/dataset.py
@@ -57,6 +57,7 @@ from datumaro.components.launcher import Launcher
 from datumaro.components.media import Image, MediaElement
 from datumaro.components.merge import DEFAULT_MERGE_POLICY
 from datumaro.components.progress_reporting import NullProgressReporter, ProgressReporter
+from datumaro.components.task import TaskType
 from datumaro.components.transformer import ItemTransform, ModelTransform, Transform
 from datumaro.util.log_utils import logging_disabled
 from datumaro.util.meta_file_util import load_hash_key
@@ -160,6 +161,7 @@ class Dataset(IDataset):
         *,
         env: Optional[Environment] = None,
         media_type: Type[MediaElement] = Image,
+        task_type: Optional[TaskType] = None,
     ) -> Dataset:
         """
         Creates a new dataset from an iterable object producing dataset items -
@@ -254,6 +256,7 @@ class Dataset(IDataset):
         infos: Optional[DatasetInfo] = None,
         categories: Optional[CategoriesInfo] = None,
         media_type: Optional[Type[MediaElement]] = None,
+        task_type: Optional[TaskType] = None,
         env: Optional[Environment] = None,
     ) -> None:
         super().__init__()
--- a/src/datumaro/components/dataset.py
+++ b/src/datumaro/components/dataset.py
@@ -164,7 +164,7 @@ class Dataset(IDataset):
         *,
         env: Optional[Environment] = None,
         media_type: Type[MediaElement] = Image,
-        task_type: Optional[TaskType] = None,
+        task_type: Optional[TaskType] = TaskType.unlabeled,
     ) -> Dataset:
         """
         Creates a new dataset from an iterable object producing dataset items -
--- a/src/datumaro/components/dataset.py
+++ b/src/datumaro/components/dataset.py
@@ -884,7 +884,6 @@ class Dataset(IDataset):
                     extractors.append(
                         env.make_extractor(src_conf.format, src_conf.url, **extractor_kwargs)
                     )
-
             dataset = (
                 cls(source=extractor_merger(extractors), env=env)
                 if extractor_merger is not None
--- a/src/datumaro/components/dataset.py
+++ b/src/datumaro/components/dataset.py
@@ -270,7 +270,7 @@ class Dataset(IDataset):
 
         self.eager = None
         self._data = DatasetStorage(
-            source,
+            source=source,
             infos=infos,
             categories=categories,
             media_type=media_type,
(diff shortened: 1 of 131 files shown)


src/datumaro/components/task.py
```diff
@@ -0,0 +1,91 @@
...
+    detection_landmark = 5
+    detection_rotated = 6
+    detection_3d = 7
+    segmentation_semantic = 8
+    segmentation_instance = 9
+    segmentation_panoptic = 10
+    caption = 11
+    super_resolution = 12
+    depth_estimation = 13
+    mixed = 14
```
jihyeonyi, line 28: Is there a way for users to know what tasks are possible when TaskType is mixed?

src/datumaro/components/dataset.py
vinnamkim, line 1: I could see there are many changes in `plugins/data_format`. However, I'd rather revert them and let a set of annotation types existing in the dataset to be managed by `DatasetStorage` (`Dataset`'s dataset item container) or `StreamDatasetStorage` (a correspondent to `StreamDataset`). This is because 

1. This implementation makes a hidden constraint that every dataset extractor (`DatasetBase`) should implement an annotation type gatherer such as
https://github.com/openvinotoolkit/datumaro/blob/f97820b6e6d6de107af8561c829db4321e7b7f70/src/datumaro/plugins/data_formats/ade20k2017.py#L72
2. This implementation is not aligned with our dataset transform logics. It currently compute `task_type` at `DatasetBase`. Let's assume that some `DatasetBase` decides that a given dataset has two annotation types, `Label` and `Bbox`. However, if an arbitrary dataset transform is applied on top of it and it drops every `Bbox`, we must re-compute a set of annotation types existed in the dataset after transformation. This should be done by `DatasetStorage` or `StreamDatasetStorage`.

Following this idea, it would be
```python
class DatasetStorage:
    def __init__(self):
         ...
         self._set_of_ann_types: set | None = None

...

    @property
    def set_of_ann_types(self):
        if self._set_of_ann_types is None:
             self._set_of_ann_types = set()
             # If reset or not computed, run its iterator to compute
             for item in self:
                 for ann in item.annotations:
                     self._set_of_ann_types.add(ann.type)
        return self._set_of_ann_types

    @property
    def task_type(self):
        return infer_task_type_from_set_of_ann_types(self.set_of_ann_types)

...

    def _iter_init_cache_unchecked(self) -> Iterable[DatasetItem]:
        # Merges the source, source transforms and patch, caches the result
        # and provides an iterator for the resulting item sequence.
        ...

        # Reset if there is a possible change
        self._set_of_ann_types = None
```

wonjuleee: Thank you for the good idea. When I approached with this way, it needs to have a single iteration of whole dataset for obtaining available task information. So, I have turned to obtain available task during importing. What do you think?

jihyeonyi:
I have a question. I'm curious why datasets are designed to have a single task_type. For example, if a dataset has both label and bbox annotations, it can be used for both classification and detection tasks. (And even for anomaly cls/det. tasks if the labels are anomalous and normal). However, according to your implementation, it seems like the task_type becomes detection.

wonjuleee:
> I have a question. I'm curious why datasets are designed to have a single task_type. For example, if a dataset has both label and bbox annotations, it can be used for both classification and detection tasks. (And even for anomaly cls/det. tasks if the labels are anomalous and normal). However, according to your implementation, it seems like the task_type becomes detection.

Hi @jihyeonyi, thank you for the question. We are able to identify the mapping between annotation types and tasks in `task.py` and there is some included or excluded relationships between them. So it is preferable to provide new transformations for changing/filtering dataset items per each desired task types. Let me add more explanations later.

src/datumaro/components/task.py
wonjuleee: `Mixed` task can be transformed to any task types. The reason why we are providing `mixed` is because Datumaro format can have any `AnnotationType` when importing.

src/datumaro/components/dataset.py
wonjuleee: Please see 3d30c1d91e458a72eaf4b9771a0aa73964c2ce7a

src/datumaro/components/dataset_storage.py
```diff
@@ -420,8 +471,13 @@ def put(self, item: DatasetItem) -> None:
 
         if is_new and not self.is_cache_initialized():
             self._length = None
+            self._set_of_ann_types = set()
         if self._length is not None:
             self._length += is_new
+            # # when adding a new item, task_type will be updated automatically
+            # for ann in item.annotations:
+            #     self._set_of_ann_types.add(ann.type)
+            # self._task_type = TaskAnnotationMapping().get_task(self._set_of_ann_types)
```
jihyeonyi: I think this can be deleted.

src/datumaro/components/dataset_storage.py
```diff
@@ -643,7 +705,15 @@ def stacked_transform(self) -> IDataset:
         return transform
 
     def __iter__(self) -> Iterator[DatasetItem]:
-        yield from self.stacked_transform
+        # yield from self.stacked_transform
```
jihyeonyi: This line can be deleted.

src/datumaro/components/dataset_storage.py
wonjuleee: fixed at 0a313ca5d4b9e08208b20a226840d026456c1655

src/datumaro/components/dataset_storage.py
wonjuleee: fixed at 0a313ca5d4b9e08208b20a226840d026456c1655

jihyeonyi (approved):

PR #1442, opened 2024-04-16
Fix ambiguous coco format detector

wonjuleee:
### Summary

Before, all the coco datasets such as panoptic, stuff, caption are detected as just `coco`.
With this PR, a specific coco task is detected.
For instance,
```code
formats = dm.Environment().detect_dataset(path="./tests/assets/coco_dataset/coco_instances")
```
returns `coco_instances`.

If the dataset directory contains multiple task types of COCO, it looks like
```code
formats = dm.Environment().detect_dataset(path="./tests/assets/coco_dataset/coco")
```
returns `coco`.

So we can choose one task for importing operation.
We here remain the original `coco` to import all annotation files like below.
image

*note*. Previously, our coco importer can load any annotation filename like `{unknowntask}_{subsetname}`, but with this PR, it is restricted to import `{instances,panoptic,caption,labels,...}_{subsetname}`.

### Checklist

### License

```python
#
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   (<https://github.com/openvinotoolkit/datumaro/pull/1422>)
 
 ### Enhancements
+- Fix ambiguous COCO format detector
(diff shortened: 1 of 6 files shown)


sooahleex (approved):
Looks good to me.

PR #1471, opened 2024-04-23
Get target information for tabular dataset

sooahleex:
### Summary

- Set `target` as `dictionary` to get input and output column. ex. `{"input": "question", "output":["rating", "sentiment"]}`
- If `target` is `None`, bring all columns
- Set input of target column as media and output of it as annotations
  - For example, if we set target as `{"input": "length(m)", "output": ["breed_category", "pet_category"]}`
    image
    And each item for train and test subset is like below.
    image
    Test file did not have column for `breed_category` and `pet_category`, so test subset did not have any annotations. 

- Set `CategoricalDtype` for column which have dtype as object but that could be used as label. For this, we should define the threshold.

### How to test

- Fix target for unit test and cli test

### Checklist

### License

```python
#
```

### DIFF
--- a/src/datumaro/plugins/data_formats/tabular.py
+++ b/src/datumaro/plugins/data_formats/tabular.py
@@ -64,7 +64,7 @@ class TabularDataBase(DatasetBase):
     def _parse(
         self,
         paths: List[str],
-        target: Optional[Union[str, List[str]]] = None,
+        target: Optional[Dict[str, List[str]]] = None,
         dtype: Optional[Dict[str, Type[TableDtype]]] = None,
     ) -> Tuple[List[DatasetItem], Dict[AnnotationType, Categories]]:
         """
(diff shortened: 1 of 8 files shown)


src/datumaro/plugins/data_formats/tabular.py
```diff
@@ -84,45 +86,56 @@ def _parse(
         items: List[DatasetItem] = []
         categories: TabularCategories = TabularCategories()
 
+        if target is not None:
+            target = eval(",".join(target)) if isinstance(target, list) else target
+            if target["input"] is None or target["output"] is None:
+                raise TypeError('Target should have both "input" and "output"')
```
wonjuleee, line 91: maybe it is more good for considering `target={}`. In this case, media=TableRow(all_data) and annotations=[]. What do you think?

sooahleex: I set `media=TableRow(all_data)` and `annotations = []` when `target = {}`. Fixed in [0c84d9e](https://github.com/openvinotoolkit/datumaro/pull/1471/commits/0c84d9e5673c1832d67092b827ba71507b551152)

wonjuleee: Thank you

wonjuleee (approved):
LGTM

PR #1480, opened 2024-04-26
Add ExtractedMask and update importers who can use it to use it

vinnamkim:
### Summary

- Ticket no. 138711
- Add `ExtractedMask` which is a dedicated mask annotation class to use a single index mask source. The index mask is a integer 2D array and its pixel can indicate a label id (class) or instance id.
- The advantage of introducing this class is that we can minimize a memory burden by sharing a single source explicitly. In addition, we can use a computational burden for the following scenario:

| | Create a semantic segmentation map by combining Datumaro Mask annotations |
| :-: | :-: |
| Setup | Assume there is a Datumaro dataset item and it has multiple Datumaro Mask. We can merge them into one index mask to construct a semantic segmentation map. |
|  Before | 1) For each Datumaro Mask annotation, create a binary masks, 2) Merge binary masks into one index mask |
| After |  1) Check all Mask annotations are `ExtractedMask` and it share the same index mask, 2) Just extract the source index mask |

It can save huge computations required for 2D array manipulations: creating a binary mask and merging binary masks.

### Checklist

### License

```python
#
```


src/datumaro/components/annotation.py
```diff
@@ -501,6 +503,41 @@ def __eq__(self, other):
...
+        index_mask: Integer 2D Numpy array. Its pixel can indicate a label id (class) or an instance id.
+        index: Integer value to extract a binary mask from the given index mask.
+
+    Examples:
+        >>> import numpy as np
+        >>> from datumaro.components.annotation import ExtractedMask
+        >>>
+        >>> index_mask = np.random.randint(low=1, high=3, size=(10, 10), dtype=np.uint8)
+        >>> mask1 = ExtractedMask(index_mask=index_mask, index=1, label=1) #
+        >>> mask2 = ExtractedMask(index_mask=index_mask, index=2, label=2) #
```
wonjuleee: What is the difference between `index` and `label` here?

src/datumaro/components/annotation.py
```diff
@@ -501,6 +503,41 @@ def __eq__(self, other):
...
+        >>> np.unique(mask1.image).tolist()  # `image` property create a binary mask
+        np.array([0, 1])
+        >>> mask1.index_mask == mask2.index_mask  # They share the same source
+        True
+    """
+
+    index_mask: Union[IndexMaskImage, IndexMaskImageCallable] = field()
```
wonjuleee, line 534: This looks like `instance_mask` for me. How about using `instance_mask` instead?

src/datumaro/plugins/data_formats/ade20k2017.py
```diff
@@ -108,9 +107,10 @@ def _load_items(self, subset):
                     attributes = {k: True for k in v["attributes"]}
 
                     item_annotations.append(
-                        Mask(
+                        ExtractedMask(
+                            index_mask=mask,
+                            index=instance_id,
```
wonjuleee, line 112: If so, there might be the copied `index_mask`s for all annotations in an item. For instance, if an image is composed of 100-class indices, we should cope an index mask for 100 times. Don't we have any memory leakage problem?

src/datumaro/components/annotation.py
vinnamkim: As described here
https://github.com/openvinotoolkit/datumaro/blob/e2e945e5d25d3e1437c8a42f756eab14bd144d25/src/datumaro/components/annotation.py#L513-L515
`index` can denote an instance id or a label id. It depends on the dataset. Anyway, I enhanced the example docstring in https://github.com/openvinotoolkit/datumaro/pull/1480/commits/af03f2989f5f983d6adab947280631814be4fccc to make it more descriptive.

src/datumaro/components/annotation.py
vinnamkim: Likewise https://github.com/openvinotoolkit/datumaro/pull/1480#discussion_r1582527296, `index_mask` is 2d integer map which pixel can indicate an instance id or label id. It is available to represent one of either.

src/datumaro/plugins/data_formats/ade20k2017.py
vinnamkim: No, we do not deep-copy it. `index_mask` is a reference (pointer) of where the mask data (2d array) exists.  We do this to maintain a single source "explicitly" if multiple binary masks (e.g., `datumaro.Mask`) are created from the single source.

The advantage of explicitly maintaining `index_mask` reference is described in the PR description above:
> | | |
> | :-: | :-: |
> | After | 1) Check all Mask annotations are ExtractedMask and it share the same index mask, 2) Just extract the source index mask |

If we want to merge multiple `ExtractedMask` from the same source into one, we can simply bring the `index_mask` rather than merging each binary mask step by step.

src/datumaro/components/annotation.py
wonjuleee: I am still confused. The updated example have the same index and label. In my understanding, mask1 is a binary mask with 0 and 1, where all 1s indicate the `index` indeed. This looks fine. But `index` might be `label` for all cases. Is there any case that `index` is not equal to `label`?

src/datumaro/components/annotation.py
wonjuleee: So I am saying about the rename of `index_mask` to `instance_mask` for keeping the consistencies between Datumaro. Please correct me if I was wrong.

src/datumaro/plugins/data_formats/ade20k2017.py
wonjuleee: Got it. If so, all masks from an image should have the same index. It might be less possible, but you considered the case that an image has multiple class index masks (for instance, if an image can have multiple label schema for the segmentation). Is this true?

src/datumaro/plugins/data_formats/ade20k2020.py
```diff
@@ -111,10 +111,11 @@ def _load_items(self, subset):
                 for class_idx, label_name in classes:
                     label_id = labels.find(label_name)[0]
                     item_annotations.append(
-                        Mask(
+                        ExtractedMask(
+                            index_mask=mask,
+                            index=class_idx,
```
wonjuleee, line 116: shouldn't this be `index=label_id`? `class_idx` can be different to `label_id` because of `label_categories` generation in Datumaro. It looks like the potential bug of Datumaro.

src/datumaro/plugins/data_formats/icdar/base.py
```diff
@@ -248,18 +248,19 @@ def _load_segmentation_items(self):
...
+                classes = np.unique(np_index_mask)
                 for label_id in classes:
                     if label_id == 0:
                         continue
                     i = int(label_id)
                     annotations.append(
-                        Mask(
+                        ExtractedMask(
+                            index_mask=index_mask,
+                            index=label_id,
```
wonjuleee, line 262: We may miss the `label=label_id` here.

vinnamkim: It was intended. There was no `label` assignment previously. You can see L259, `i = int(label_id)` and L263 `group=groups[i]`. The original implementation just left a `group` assignment here.

src/datumaro/plugins/data_formats/ade20k2020.py
vinnamkim: This should be 100% equivalent to the previous implementation (`mask.lazy_extract(class_idx)`). The previous implementation extract a binary mask with `mask.lazy_extract(class_idx)`. Then it can use the binary mask with its `label=label_id`. This is exactly same as the current implementation. It just replaces `mask.lazy_extract(class_idx)` with `ExtractedMask(index_mask=mask, index=class_idx, ...)`. The only difference is that we explicitly maintain the source reference `index_mask` as `ExtractedMask.index_mask`.

src/datumaro/components/annotation.py
vinnamkim: I thought that there is two mask types broadly accepted in the ML community: 1) instance mask and 2) class mask. The instance mask represents an unique identifier for object. For example, there are two apples, but they can have two instance mask values such as 0 for the first apple and 1 for the second apple. On the other hand, class mask represents a semantic label. In the previously example, both apples will have 1 class mask values (0 for background and 1 for apple).

However, this time, I introduced a third terminology which can express both. It is possible that we can design `ExtractedMask` to specify the source mask one of either such as
```python
class ExtractedMask:
     source: np.ndarray
     source_type: Literal["instance", "class"]
```
, and it was my first thought but I changed this design to the current one. The reason why I changed my mind can be found in this comment, https://github.com/openvinotoolkit/datumaro/pull/1480/files#r1582581086. This data format tries to import something that is not clear whether its source is instance mask or class mask. I thought that it can be happened by any chance, so I'd like to introduce more general abstraction for this by introducing `index_mask` and `index`.

src/datumaro/components/annotation.py
vinnamkim: `mask` is a binary mask and its 1s actually indicate `label=1` (not `index=1`) in this example. `index` is just used to extract a binary mask from `index_mask`. It is possible that `index` can be different with `label` by dataset transformation. For example, if the label remapping transform merges `label=1, index=1` (1: apple) and `label=2, index=2` (2: orange) to `label=3` (3: fruits). Then, apple masks will have `index=1, label=3` and orange masks will have `index=2, label=3`.

src/datumaro/plugins/data_formats/ade20k2017.py
vinnamkim: I guess I cannot understand your comment well. Can we proceed this conversation in Teams?

wonjuleee (approved):

PR #1483, opened 2024-04-29
Update stability tests

yunchu:
### Summary

Update weekly workflow to run fuzzing with different extra installation option instead of pytest-loop.

### Checklist

### License

```python
#
```

### DIFF
--- a/.github/workflows/weekly_check.yml
+++ b/.github/workflows/weekly_check.yml
@@ -19,18 +19,17 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ['ubuntu-20.04', 'windows-2022']
-        python-version: ['3.10']
+        python-version: ['3.9', '3.10', '3.11']
         extra: ['default', 'tf', 'tfds', 'torch']
         include:
+          - python-version: "3.9"
+            tox-env-py: "39"
           - python-version: "3.10"
             tox-env-py: "310"
-          - os: "ubuntu-20.04"
-            tox-env-os: "lin"
-          - os: "windows-2022"
-            tox-env-os: "win"
-    name: stability test (${{ matrix.os }}, Python ${{ matrix.python-version }}, Extra ${{ matrix.extra }})
-    runs-on: ${{ matrix.os }}
+          - python-version: "3.11"
+            tox-env-py: "311"
+    name: stability test (Python ${{ matrix.python-version }}, Extra ${{ matrix.extra }})
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
       - name: Installing python
@@ -46,7 +45,7 @@ jobs:
       - name: Upload test results
         uses: actions/upload-artifact@v4
         with:
-          name: stability-${{ matrix.tox-env-os }}-${{ matrix.extra }}
+          name: stability-${{ matrix.python-version }}-${{ matrix.extra }}
           path: .tox/stability-${{ matrix.extra }}/log
   call-notify-to-teams:
     needs: [weekly_stability_test]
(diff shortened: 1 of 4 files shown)


.github/workflows/weekly_check.yml
```diff
@@ -19,17 +19,17 @@ jobs:
...
-          - os: "ubuntu-20.04"
-            tox-env-os: "lin"
-          - os: "windows-2022"
-            tox-env-os: "win"
-    name: stability test (${{ matrix.os }}, Python ${{ matrix.python-version }})
-    runs-on: ${{ matrix.os }}
+          - python-version: "3.11"
+            tox-env-py: "311"
+    name: stability test (Python ${{ matrix.python-version }}, Extra ${{ matrix.extra }})
+    runs-on: ubuntu-latest
```
wonjuleee, line 32: can I ask you why testing on different `os` is dropped?

tox.ini
```diff
@@ -76,29 +76,30 @@ commands =
...
 allowlist_externals =
     bash
 commands_pre =
     bash -c 'cargo -V; echo "cargo (rust) version checking exit code = $?"'
-    python -m pip uninstall pytest-stress -y
 commands =
-    python -m pytest -v --csv={toxworkdir}/results-{envname}.csv -x {posargs:--loop 5}
-
+    ; coverage erase
+    python tests/fuzzing/cli_fuzzing.py {posargs:-dict=tests/fuzzing/assets/cli_operations.dict -artifact_prefix={toxworkdir}/ -print_final_stats=1 -atheris_runs=500000}
```
wonjuleee, line 89: what is the difference between fuzzing and stability tests?

.github/workflows/weekly_check.yml
yunchu: we're using fuzzing tool `atheris` but they does not support windows.

tox.ini
yunchu: Stability tests repeatedly runs all tests 5 times sequentially and the results were used for the stress test report which is a requirement to the release but it was replaced with the fuzzing after we implemented it.
Fuzzing or fuzz testing is the technic to verify stability or robustness of interfaces such as CLI or API that exposed from the project and it works by passing randomly generated data to the CLI or API as the arguments.
The expectation of this changes would be that datumaro CLI provides proper help messages or usage without having any uncoughted exceptions on the different environments that setup by using different extra options while installing datumaro.

.github/workflows/weekly_check.yml
wonjuleee: Is this okay to skip tests for windows?

tox.ini
wonjuleee: Is there any background of having `atheris`?

wonjuleee: Previously, I meant that fuzzing requires 500K runs, while stability runs the same test code only for 5K times.
Is stability for testing extra dependencies? Could you let me know the test schedule for them (daily, weekly, manual, etc.)?

.github/workflows/weekly_check.yml
yunchu: current fuzzing implementation is very primitive and works only with datumaro CLI. but I can say that this changes would cover small portion of stability tests that required to this product. it's better to work as well for the windows but it's not supported by a tool of fuzzing we're using.

tox.ini
yunchu: fuzzing will be performed once per a release (not scheduled, on demand base) but I want to verify basic stability of the datumaro CLI on all available installation environments of datumaro those can be created through the different extra options. To do that, i don't think that needs to attempt too much trials and it's good to run with weekly schedule.

wonjuleee (approved):

PR #1485, opened 2024-04-30
Split video directory by subset in datumaro format. 

jihyeonyi:
### Summary

* ticket No.: 137978

According to the [datumaro documentation](https://openvinotoolkit.github.io/datumaro/latest/docs/data-formats/formats/datumaro.html), video directory is not separated by subset. 
image

So a video will be overwritten if it has a duplicated filename in other subset. 

Following example has two subsets, `train` and `val`, and both subsets have Video_0.avi, ..., Video_9.avi. 
So the videos in the train subset is overwritten by the videos in the val subset.
image

So I'd like to split the video directory by subset to avoid this situation. 

### Checklist

### License

```python
#
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add ExtractedMask and update importers who can use it to use it
   (<https://github.com/openvinotoolkit/datumaro/pull/1480>)
 
+### Bug fixes
+- Split the video directory into subsets to avoid overwriting
+  (<https://github.com/openvinotoolkit/datumaro/pull/1485>)
+
 ## May 2024 Release 1.6.1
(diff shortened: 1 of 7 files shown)


wonjuleee (approved):
LGTM, but is there any modification required for datumaro importer for videos?
One more question, considering the fact that a video could be huge, this change needs more storages (2 twice at the worst case). Do you have any reason to save duplicated videos for each subset?

jihyeonyi:
> LGTM, but is there any modification required for datumaro importer for videos? One more question, considering the fact that a video could be huge, this change needs more storages (2 twice at the worst case). Do you have any reason to save duplicated videos for each subset?

I have modified the import, too. Basic assumption is that a video isn't shared by subsets. So there shouldn't be duplicated videos. However, if this assumption is too string, than we need another workaround to save videos.

PR #1498, opened 2024-05-13
Add TabularValidator

sooahleex:
### Summary

- Add TabularValidator
- Validate annotations based on dataset after `AstypeAnnotations`
- Validation result consists of 
  - `defined_labels`
  - `empty_labels` : which label is not set for item(= nan label)
  - `defined_captions`
  - `empty_captions` : which caption is not set for item (= nan caption)
  - `redundancies`
  - `total_ann_count`
  - `items_missing_annotations` : if item does not any annotations
  - `items_broken_annotations`
- Anomaly type consists of
  - `MissingAnnotation` : if item did not have any caption or label
  - `BrokenAnnotations`
  - `FewSamplesInLabel`
  - `ImbalancedLabels`
  - `EmptyLabel` : If value is Nan for a label
  - `FewSamplesInCaption`
  - `RedundanciesInCaption`
  - `EmptyCaption` : If value is Nan for a caption
  - `ImbalancedCaptions` : There is an imbalance in the caption distribution
  - `ImbalancedDistInCaptions` : Values are not evenly distributed for a caption only if caption is number
  - `FarFromCaptionMean` : An annotation has ann too small or large value than average for a caption only if caption is number
- Updating ConfigurableValidator will be covered following PR

### How to test

- Add unit test for TabularValidator

### Checklist

### License

```python
#
```


src/datumaro/plugins/validators.py
```diff
@@ -1196,3 +1208,451 @@ def generate_reports(self, stats):
...
+        label_columns = [
+            cat
+            for cat, type_ in dataset._tabular_cat_types.items()
+            if isinstance(type_, CategoricalDtype)
+        ]
+        redundancies_in_caption_dist = {
+            cat: deepcopy(redundancies_template)
+            for cat, type_ in dataset._tabular_cat_types.items()
+            if type_ is str
+        }
```
vinnamkim: `TabularValidator` seems completely new one and no dependency on existing one. Can we make this first step cleaner? Please introduce static data objects rather than using Python dictionary for internal valuables such as
```suggestion
@dataclasses
class TabularValidationStats:
      total_ann_count: int
      ...
      
      def to_dict(self):
           return {
                "total_ann_count": self.total_ann_count,
                ...
           }

...

        # No dictionary anymore
        # stats = {...}
        stats = TabularValidationStats()
```

sooahleex: Fix in [fed5e27](https://github.com/openvinotoolkit/datumaro/pull/1498/commits/fed5e272400bcd3d4cf336035f58571caaffdcc0). If there is any problem, please let me know.

vinnamkim (approved):

PR #1503, opened 2024-05-21
Enhance DatasetItem annotations for semantic segmentation model training use case

vinnamkim:
### Summary

- Ticket no. 141410
- Datumaro has been maintaining every mask as a binary mask representation. This is because of flexibility on manipulating dataset level management (e.g., label remapping). However, this design can bring performance degradation in case of model training use case. For example, semantic segmentation model training uses 2d mask filled in an integer value for each pixel. However, during importing a dataset, Datumaro should convert it to a collection of binary masks. It makes impossible to use the integer mask directly, so that it requires a 2d complexity computation cost to merge the imported binary masks into an integer mask.
- To resolve this performance degradation, this PR introduces a new interface on `dataset_item.annotations` by introducing `Annotation(list[Annotation])` class. This class extends the pure Python list to equip more utility functions. At the same time, the utility function we introduce this time is `dataset_item.annotations.get_semantic_seg_mask()`. This function bypasses binary mask conversion to construct an integer mask used for semantic segmentation model training.

### How to test
- Added unit tests to cover this change (mainly in `tests/unit/test_annotation.py`).
- This is a performance test on the semantic segmentation model training use case. Throughput improvement on getting a semantic segmentation mask is ~3x (31.38 it/s -> 99.31 it/s).
image

### Checklist

### License

```python
#
```


src/datumaro/plugins/data_formats/widerface.py
```diff
@@ -157,7 +157,7 @@ def _load_items(self, path):
                                     attributes[attr] = bbox_list[i]
                             i += 1
 
-                    annotations.append(
```
wonjuleee, line 160: Why only Widerface is affected by this change?

src/datumaro/plugins/data_formats/widerface.py
```diff
@@ -157,7 +157,7 @@ def _load_items(self, path):
...
+                    items[item_id].annotations.append(
```
wonjuleee, line 160: Why only Widerface is affected by this change?

wonjuleee (approved):

src/datumaro/plugins/data_formats/widerface.py
vinnamkim: At this line, `annotations: List` is created
https://github.com/openvinotoolkit/datumaro/blob/072c8a80398291a139f1f2739ae7dcd88b52ca4b/src/datumaro/plugins/data_formats/widerface.py#L111
Then, this `annotations` is used to construct `DatasetItem` here
https://github.com/openvinotoolkit/datumaro/blob/072c8a80398291a139f1f2739ae7dcd88b52ca4b/src/datumaro/plugins/data_formats/widerface.py#L123-L128
However, in the following lines, `Bbox` is pushed into `annotations`, not `items[item_id].annotations`,
https://github.com/openvinotoolkit/datumaro/blob/072c8a80398291a139f1f2739ae7dcd88b52ca4b/src/datumaro/plugins/data_formats/widerface.py#L160-L169

Previously, it should work because `id(annotations) == id(items[item_id].annotations)` (same Python object). However, after this PR change, 
```python
annotations: List
items[item_id].annotations: Annotations
=>
id(annotations) != id(items[item_id].annotations)
```
, so that `widerface.py#L160-L169` above will be not working.

PR #1513, opened 2024-05-29
Add correct functionality for tabular data type

sooahleex:
### Summary

- Modify `correct` functionality in transform for tabular dataset
- Correct dataset based on `TabularValidator` reports
- Method
  - Remove items
  - Data Imputation
    - Handle Missing Value : Replace with a representative value (like SimpleImputer)
      - Real(continuous) variables
        - Symmetric Distribution : `mean`
        - Asymmetric Distribution : `median`
      - Integer(continuous) or categorical value : `most_frequent`
    - Handle Outliers for numeric : Cap outliers with the upper bound of IQR method
  - Remove Redundancies
    - Remove HTML Tags
    - Remove URLs
    - Remove Emojis
    - Convert to lowercase
    - Remove Special Characters and Punctuation
    - Remove Extra Whitespace
    - Remove Stopwords
- Add new anomlay type `OutlierInCaption` to update reports

### How to test

- Add unit test for correct and update existed tests

### Checklist

- [X] I have added unit tests to cover my changes.​
- [ ] I have added integration tests to cover my changes.​
- [X] I have added the description of my changes into [CHANGELOG](https://github.com/openvinotoolkit/datumaro/blob/develop/CHANGELOG.md).​
- [ ] I have updated the [documentation](https://github.com/openvinotoolkit/datumaro/tree/develop/docs) accordingly

### License

- [ ] I submit _my code changes_ under the same [MIT License](https://github.com/openvinotoolkit/datumaro/blob/develop/LICENSE) that covers the project.
  Feel free to contact the maintainers if that's a concern.
- [ ] I have updated the license header for each file (see an example below).

```python
# Copyright (C) 2024 Intel Corporation
#
# SPDX-License-Identifier: MIT
```

PR #1520, opened 2024-06-04
Add Clean Transform for tabular data type

sooahleex:
### Summary

- This PR should be merged with #1513
- Add Clean Transform to refine dataset without any reports

### How to test

- Add unit test for Clean Transform

### Checklist

### License

```python
#
```

### DIFF
--- a/src/datumaro/plugins/transforms.py
+++ b/src/datumaro/plugins/transforms.py
@@ -1387,9 +1387,12 @@ class Correct(Transform, CliPlugin):
 
         self._outlier_captions = defaultdict(list)
         self._outlier_value = {}
+<<<<<<< HEAD
 
         self._far_from_mean_caption = defaultdict(list)
         self._far_from_mean_value = {}
+=======
+>>>>>>> c35a9fc9e (Edit correct function for tabular dataset)
 
         self._analyze_reports(report=self._reports)
 
@@ -1477,6 +1480,7 @@ class Correct(Transform, CliPlugin):
                 caption = rep["description"].split("'")[1]
                 self._empty_captions[(rep["item_id"], rep["subset"])].append(caption)
 
+<<<<<<< HEAD
             if rep["anomaly_type"] == OutlierInCaption.__name__:
                 desc = rep["description"].split("'")
                 caption = desc[1]
@@ -1492,6 +1496,22 @@ class Correct(Transform, CliPlugin):
                 upper_bound = float(desc[11])
                 self._far_from_mean_caption[(rep["item_id"], rep["subset"])].append(caption)
                 self._far_from_mean_value[caption] = (lower_bound, upper_bound)
+=======
+            if rep["anomaly_type"] == "FewSamplesInCaption":
+                caption = rep["description"].split("'")[1]  # TODO
+
+            if rep["anomaly_type"] in [
+                "OutlierInCaption",
+                "ImbalancedDistInCaption",
+            ]:  # FarFromCaptionMean
+                desc = rep["description"].split("'")
+                caption = desc[1]
+                caption_type = self._extractor._tabular_cat_types[caption]
+                lower_bound = caption_type(desc[5])
+                upper_bound = caption_type(desc[7])
+                self._outlier_captions[(rep["item_id"], rep["subset"])].append(caption)
+                self._outlier_value[caption] = (lower_bound, upper_bound)
+>>>>>>> c35a9fc9e (Edit correct function for tabular dataset)
 
     def remove_unnecessary_char(self, annotations, item_id):
         if self._table is None:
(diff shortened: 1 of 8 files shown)


tests/unit/test_transforms.py
```diff
@@ -1426,3 +1429,143 @@ def test_transform_annotation_type_label_with_nan(self):
...
+            media_type=TableRow,
+        )
+
+    @mark_requirement(Requirements.DATUM_GENERAL_REQ)
+    def test_remove_unneccessary_char(self):
+        example_text = "This is a test 😊! Check out https://example.com for more <b>details</b> about this text. Enjoy!!!"
+        cleaned_text = "test check details text enjoy"
+
+        with self.subTest("with None"):
+            self.assertIsNone(transforms.Clean.remove_unneccessary_char(None))
```
wonjuleee, line 1529: Sorry that I am not sure about the way to use `transforms` in Datumaro.
As far as I know, Datumaro mostly uses `transforms` like `dataset.transforms("clean", xxx)`.
Could you clarify this by providing the way to use clean transform in a general way or finding some examples like yours (transforms.Clean.xxx).

src/datumaro/plugins/transforms.py
```diff
@@ -1837,3 +1845,129 @@ def transform_item(self, item: DatasetItem):
...
+
+    def transform_item(self, item):
+        if not isinstance(item.media, TableRow):
+            raise DatumaroError(
+                "Item %s: image info is required for this " "transform" % (item.id,)
+            )
+
+        refined_media = self.refine_tabular_media(item) if item.media.has_data else None
+        refined_annotations = [
+            ann.wrap(values=refined_media.data) if isinstance(ann, Tabular) else ann
```
wonjuleee: Does this mean that the annotation will be same to refined_media.data? If so, you are thinking about performing `clean` prior to `correct` and `validate`?

src/datumaro/plugins/transforms.py
```diff
@@ -1837,3 +1845,129 @@ def transform_item(self, item: DatasetItem):
...
+            stop_words = set(stopwords.words("english"))  # TODO
+
+        text = re.sub(r"<.*?>", "", text)
+        text = re.sub(r"http\S+|www\S+|https\S+", "", text, flags=re.MULTILINE)
+        text = emoji_pattern.sub(r"", text)
+        text = text.lower()
+        text = re.sub(r"[^A-Za-z\s]+", "", text)
+        text = re.sub(r"\s+", " ", text).strip()
+        text = " ".join([word for word in text.split() if word not in stop_words])
+        return text
```
wonjuleee, line 1885: Could you please describe what each line is for in comments?

src/datumaro/plugins/transforms.py
sooahleex: The approach was planned and implemented as follows:
1. Validate -> Correct: Modify only annotations based on reports, while maintaining media information unchanged.
2. Clean: Modify both media and annotations of dataset items without relying on reports.
If there are any issues with the above approach, please let me know, and I will make the necessary adjustments.

src/datumaro/plugins/transforms.py
sooahleex: Fix in [cca8b29](https://github.com/openvinotoolkit/datumaro/pull/1520/commits/cca8b29229e1b86bec4745125d20ca3bd03fad61).

tests/unit/test_transforms.py
sooahleex: I fixed this in [cca8b29](https://github.com/openvinotoolkit/datumaro/pull/1520/commits/cca8b29229e1b86bec4745125d20ca3bd03fad61) for other cases. 
In this scenario, I've utilized the aforementioned approach to employ `transform` for testing the static functions of the class. If there are alternative methods to test these functions, please inform me, and I'll make the necessary updates.
Additionally, other tests for 'transform' adopt the usage pattern of `transform.Clean(dataset),` and thus, I've followed suit. Although there are variations in results between these methods, such as obtaining a class or dataset, it doesn't affect the functionality testing of the dataset. This is because we can employ `transform_item` while comparing datasets through iterators. 
Moreover, I've utilized the approach of `dataset.transforms("clean")` for real-case scenarios and restricted the use of `transform.Clean(dataset)` solely for testing, aligning with other tests. If any adjustments are needed, please advise accordingly

src/datumaro/plugins/transforms.py
wonjuleee: Could you double check whether it is okay to `ann.wrap(values=refined_media.data)`?

tests/unit/test_transforms.py
wonjuleee: Could you please provide an example of performing `remove_unneccessary_char` in `dataset.transforms("clean")` or `transform.Clean(dataset)` form?

src/datumaro/plugins/transforms.py
sooahleex: Fix in [59e8ede](https://github.com/openvinotoolkit/datumaro/pull/1520/commits/59e8edeb2a6fde966f5b666dbcf53d3667d93616). I updated values of the annotation with target of dataset.

wonjuleee (approved):
LGTM

PR #1528, opened 2024-06-12
Add hierarchical ImageNet-like dataset format

itrushkin:
### Summary
CVS-143460

### Checklist

### License

```python
#
```


src/datumaro/components/importer.py
```diff
@@ -197,7 +196,7 @@ def _change_context_root_path(context: FormatDetectionContext, path: str):
                 )
 
             for sub_dir in os.listdir(path):
-                if sub_dir.lower() in SUBSET_NAME_BLACKLIST:
+                if sub_dir.lower() not in ("train", "test", "val"):
```
vinnamkim: Why this change is needed?

src/datumaro/plugins/data_formats/imagenet.py
vinnamkim, line 1: It is not desired to change the existing interface for consistency between versions. Is it impossible to add this new feature in additive manner?

vinnamkim (commented):
Please make Windows CI green

src/datumaro/plugins/data_formats/imagenet.py
itrushkin: The current `image_dir` importer implementation detects images at any subfolder level, leading to ambiguity with the newly added format. To resolve this, we have two options: either change the requirements for one of the formats (I did this) or introduce a new detection confidence level between `EXTREME_LOW` for `image_dir` and `LOW` for `imagenet`.

src/datumaro/components/importer.py
itrushkin: This is necessary to distinguish between two importers with unlimited hierarchy levels: `nested_imagenet` and `nested_imagenet_with_subset_dirs`. Without this distinction, we cannot determine whether a folder is a subset directory or a top-level class.

src/datumaro/plugins/data_formats/imagenet.py
vinnamkim: Yea sorry I misunderstood that `format=imagenet` is removed and there remains `format=nested_imagenet` only.

src/datumaro/components/importer.py
vinnamkim, line 199: I meant what is the reason that we should remove `SUBSET_NAME_BLACKLIST`. Is it impossible such as

```suggestion
from datumaro.util.definitions import SUBSET_NAME_BLACKLIST, SUBSET_NAME_WHITELIST
# SUBSET_NAME_WHITELIST = ("train", "test", "val")

                if sub_dir.lower() in SUBSET_NAME_BLACKLIST or sub_dir.lower() not in SUBSET_NAME_WHITELIST:
```

p.s. I saw `("train", "test", "val")` is used other place so it seems possible to share it by declaring it at `datumaro.util.definitions`

vinnamkim: I meant what is the reason that we should remove `SUBSET_NAME_BLACKLIST`. Is it impossible such as

```suggestion
from datumaro.util.definitions import SUBSET_NAME_BLACKLIST, SUBSET_NAME_WHITELIST
# SUBSET_NAME_WHITELIST = ("train", "test", "val")

                if sub_dir.lower() in SUBSET_NAME_BLACKLIST or sub_dir.lower() not in SUBSET_NAME_WHITELIST:
```

p.s. I saw `("train", "test", "val")` is used other place so it seems possible to share it by declaring it at `datumaro.util.definitions`

itrushkin: If the whitelist condition is not fulfilled, the `sub_dir` is considered to belong to the blacklist, and vice versa. If the directory name is not on either list, it is treated as if it belongs to the blacklist and is not processed. This logically equals to process only whitelist condition.

vinnamkim: Ah ok then can you move it to `datumaro.util.definitions` to be shared in other source files such as `src/datumaro/plugins/data_formats/imagenet.py`

vinnamkim (approved):

PR #1533, opened 2024-06-17
Apply clean transform of updated annotations only for tabular annotation type

sooahleex:
### Summary

- Fix to update annotation for `Tabular` and `Caption` annotation type during `Clean` transform
- Add comment for `Clean`
- Fix typo of function name

### How to test

- Update unit test

### Checklist

### License

```python
#
```


src/datumaro/plugins/transforms.py
```diff
@@ -1967,12 +1981,22 @@ def transform_item(self, item):
...
+                    key: refined_media.data[key] for key in item.annotations[0].values.keys()
+                }  # only for tabular
+                ann = ann.wrap(values=annotation_values)
+            elif isinstance(ann, Caption):
+                value = [
+                    f"{key}{sep_token}{refined_media.data[key]}"
+                    for key in refined_media.data.keys()
+                    if ann.caption.startswith(key)
+                ]
+                ann = ann.wrap(caption=value[0])
```
vinnamkim, line 2003: Is it guaranteed that `value` has always one element?

src/datumaro/plugins/transforms.py
```diff
@@ -1967,12 +1981,22 @@ def transform_item(self, item):
...
-        }  # only for tabular
-        refined_annotations = [
-            ann.wrap(values=annotation_values) if isinstance(ann, Tabular) else ann
-            for ann in item.annotations
-        ]
+        refined_annotations = []
+        for ann in item.annotations:
+            if isinstance(ann, Tabular):
+                annotation_values = {
+                    key: refined_media.data[key] for key in item.annotations[0].values.keys()
```
vinnamkim, line 1994: What's the reason that we do not care `item.annotations[i].values` for `i=1,...` not just `i=0`?

sooahleex: If a dataset item contains a tabular annotation type, it will contain only one annotation of tabular for each item. This item will not contain any other annotations. I think this part is a bit hard-coded, so if there is another way to check the annotation type, please let me know.

src/datumaro/plugins/transforms.py
sooahleex: As we get value through the above lines, caption would have only one value.

vinnamkim: `value` comes from this for-loop `for key in refined_media.data.keys()`. How did you guarantee it is singular?

src/datumaro/plugins/transforms.py
```diff
@@ -1967,12 +1981,22 @@ def transform_item(self, item):
...
-            key: refined_media.data[key] for key in item.annotations[0].values.keys()
-        }  # only for tabular
-        refined_annotations = [
-            ann.wrap(values=annotation_values) if isinstance(ann, Tabular) else ann
-            for ann in item.annotations
-        ]
+        refined_annotations = []
+        for ann in item.annotations:
+            if isinstance(ann, Tabular):
+                annotation_values = {
```
vinnamkim: You might be able to put in a kind of checker here such as
```suggestion
            if isinstance(ann, Tabular):
                if len(item.annotations) != 1:
                    msg = ...
                    raise ValueError(msg)
                annotation_values = {
```

src/datumaro/plugins/transforms.py
sooahleex: So far, the expectation is to do `AstypeAnnotations` on a tabular dataset before the `Clean` transformation. So if the annotation type is `Caption`, then it should have only one of them. We've refined the tabular annotation, so if "Title" is "It's really good!", we've transformed it to "Title:It's really good! We have not yet covered other normal image datasets that are not refined tabular datasets. So if the annotation type is caption, it should be singular. To cover other normal image datasets, I will come back to this.

src/datumaro/plugins/transforms.py
sooahleex: I added this through [afd56c5](https://github.com/openvinotoolkit/datumaro/pull/1533/commits/afd56c5ceaf32e6d778d05753e1b4ea27d7943be). Thank you for your suggestion.

src/datumaro/plugins/transforms.py
vinnamkim: That seems an implicit condition in Datumaro. What I want is to make everything explicit if possible. Relying on the implicit condition can lead to a bug. If someone develops something that can changes the implicit condition, he must never be noticed what he is doing can affect to the behavior of this code. It leads to a bad result in software maintenance. Add something like https://github.com/openvinotoolkit/datumaro/pull/1533#discussion_r1651887322

sooahleex: That's why I added https://github.com/sooahleex/datumaro/blob/f6d5c3902d59e7f2d84ad8ef74edf3dfad2c9da6/src/datumaro/plugins/transforms.py#L1979-L1982. If this is not enough to show the implicit condition, please let me know. I will consider it.

vinnamkim (approved):

PR #1548, opened 2024-06-27
verify w and h input multiplication overflow to rleEncode()

yunchu:
### Summary

To apply same changes of #1544 to 1.8.0

### Checklist

### License

```python
#
```

### DIFF
--- a/src/datumaro/capi/pybind.cpp
+++ b/src/datumaro/capi/pybind.cpp
@@ -50,7 +50,12 @@ struct RLE
 
 RLE rleEncode(const byte *M, siz h, siz w)
 {
+
     siz j, k, a = w * h;
+
+    if (a / w != h)
+        throw std::runtime_error("result overflow. input size is too big.");
+
     uint c;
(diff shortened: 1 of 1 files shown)


src/datumaro/capi/pybind.cpp
```diff
@@ -50,7 +50,12 @@ struct RLE
 
 RLE rleEncode(const byte *M, siz h, siz w)
 {
+
     siz j, k, a = w * h;
+
+    if (a / w != h)
```
jihyeonyi, line 56: Shouldn't we also check for cases where 'w' is 0? And if `a/w` results in a floating-point value, we shouldn't compare it using `!=`.

yunchu: @jihyeonyi  w and h would be passed from the size of numpy 2d array so i think it would be verified while initializing 2d array. and this function will not be exposed directly to the python.

jihyeonyi (approved):

PR #1571, opened 2024-07-23
Make ImageNet data format streamable

itrushkin:
Geti requires all data formats to be streamable, i. e. generate dataset items on demand.
This PR allows streaming for ImageNet format.

### DIFF
--- a/tests/unit/test_imagenet_format.py
+++ b/tests/unit/test_imagenet_format.py
@@ -7,7 +7,7 @@ import numpy as np
 import pytest
 
 from datumaro.components.annotation import AnnotationType, Label, LabelCategories
-from datumaro.components.dataset import Dataset
+from datumaro.components.dataset import Dataset, StreamDataset
 from datumaro.components.dataset_base import DatasetItem
 from datumaro.components.environment import Environment
 from datumaro.components.media import Image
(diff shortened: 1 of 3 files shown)


sooahleex (approved):
Looks good to me, Is there any need to add a test for this?

itrushkin:
While testing the `imagenet_with_subset_dirs` format, I found that it is still not streamable. This format usually has multiple sources and they need to be merged. Merge operation always creates non-streamable dataset. I will resolve it in the current PR.

tests/unit/test_imagenet_format.py
```diff
@@ -212,37 +211,40 @@ def _create_expected_dataset(self):
...
+        assert dataset.is_stream == is_stream
 
-        compare_datasets(self, expected_dataset, dataset, require_media=True)
+        compare_datasets(helper_tc, expected_dataset, dataset, require_media=True)
 
     @mark_requirement(Requirements.DATUM_GENERAL_REQ)
     def test_can_detect_imagenet(self):
         detected_formats = Environment().detect_dataset(self.DUMMY_DATASET_DIR)
-        self.assertEqual([self.IMPORTER_NAME], detected_formats)
+        assert [self.IMPORTER_NAME] == detected_formats
```
sooahleex, line 225: Overall changes look good to me. Is there any reason to use `assert` directly instead of using `self.asserEqual` of `unittest`?

sooahleex (approved):

tests/unit/test_imagenet_format.py
itrushkin: The class is not a subclass of TestCase anymore, so it does not have `assertEqual` method

PR #1572, opened 2024-07-24
Change common semantic segmentation dataset detection rule

itrushkin:
Since Datumaro 1.8, the following folder structure:
```
# Common semantic segmentation dataset example from OTX tests
.
└── supervised
    ├── test
    │   ├── dataset_meta.json
    │   ├── images
    │   │   ├── 0001.png
    │   │   └── 0002.png
    │   └── masks
    │       ├── 0001.png
    │       └── 0002.png
    ├── train
    │   ├── dataset_meta.json
    │   ├── images
    │   │   ├── 0001.png
    │   │   ├── 0002.png
    │   │   └── 0003.png
    │   └── masks
    │       ├── 0001.png
    │       ├── 0002.png
    │       └── 0003.png
    └── val
        ├── dataset_meta.json
        ├── images
        │   ├── 0001.png
        │   └── 0002.png
        └── masks
            ├── 0001.png
            └── 0002.png
```
is detected as both `common_semantic_segmentation` and `common_semantic_segmentation_with_subset_dirs`. Since they are detected at different levels, the current implementation could not set the higher priority to the second one.

This PR updates detection rules for the `common_semantic_segmentation` format, searching for requirements in the root directory only.


jihyeonyi (approved):
LGTM. 
However, I think related test should be modified, but I guess there's no test affected by this modification. Could you add a related test later?

jihyeonyi (approved):
LGTM.

PR #1582, opened 2024-08-12
Fix bug where custom environment plugins were lost on dataset merge

williamcorsel:
### Summary

I ran into an issue where plugins loaded into the `Environment` are not retained after merging. Turns out a generator is being passed to [Environment.merge](https://github.com/openvinotoolkit/datumaro/blob/30b1add52d6fe458dba5e32e1f63ffeea999ad7b/src/datumaro/components/environment.py#L271) instead of the expected Sequence. This meant that custom plugins were never registered into the merged environment.

### How to test

Reproducible example:

```python
from datumaro.components.environment import Environment
from datumaro.components.exporter import Exporter
from datumaro.components.hl_ops import HLOps
from datumaro.components.media import Image


class MyPlugin(Exporter):
    pass


environment = Environment()
environment.exporters.register('my_plugin', MyPlugin)

dataset1 = Dataset(media_type=Image, env=environment)
datasets = [dataset1, dataset1.clone()]
merged = HLOps.merge(*datasets)

assert 'my_plugin' in dataset1.env.exporters  # Passes
assert 'my_plugin' in merged.env.exporters  # Fails

```

### Checklist

### License

```python
#
```

### DIFF
--- a/src/datumaro/components/hl_ops/__init__.py
+++ b/src/datumaro/components/hl_ops/__init__.py
@@ -282,13 +282,14 @@ class HLOps:
         merger = get_merger(merge_policy, **kwargs)
         merged = merger(*datasets)
         env = Environment.merge(
-            dataset.env
-            for dataset in datasets
-            if hasattr(
-                dataset, "env"
-            )  # TODO: Sometimes, there is dataset which is not exactly "Dataset",
-            # e.g., VocClassificationBase. this should be fixed and every object from
-            # Dataset.import_from should have "Dataset" type.
+            [
+                dataset.env
+                for dataset in datasets
+                if hasattr(dataset, "env")
+                # TODO: Sometimes, there is dataset which is not exactly "Dataset",
+                # e.g., VocClassificationBase. this should be fixed and every object from
+                # Dataset.import_from should have "Dataset" type.
+            ]
         )
         if report_path:
             merger.save_merge_report(report_path)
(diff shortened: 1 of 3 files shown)


src/datumaro/components/hl_ops/__init__.py
```diff
@@ -282,13 +282,14 @@ def merge(
...
-            # e.g., VocClassificationBase. this should be fixed and every object from
-            # Dataset.import_from should have "Dataset" type.
+            [
+                dataset.env
+                for dataset in datasets
+                if hasattr(dataset, "env")
+                # TODO: Sometimes, there is dataset which is not exactly "Dataset",
+                # e.g., VocClassificationBase. this should be fixed and every object from
+                # Dataset.import_from should have "Dataset" type.
+            ]
```
sooahleex, line 292: I think those does not have to convert to `list` type. Do you have any reason to set those as list?

sooahleex (commented):
Thank you for your interest to our project! Could you add an unit test for this?

src/datumaro/components/hl_ops/__init__.py
williamcorsel: Besides the fact that the type hint of the merge function defines a `Sequence`, the real issue is that [here](https://github.com/williamcorsel/datumaro/blob/57f176f149c86e8061ae94b789c2d4c4724298b1/src/datumaro/components/environment.py#L272) we loop over all items to check if they're default. If you supply a generator it means that the next loop [here](https://github.com/williamcorsel/datumaro/blob/57f176f149c86e8061ae94b789c2d4c4724298b1/src/datumaro/components/environment.py#L280) is always skipped.

sooahleex (approved):
LGTM. Thanks for your interest into our projects.

PR #1594, opened 2024-08-30
Add PseudoLabeling transform for unlabeled dataset

sooahleex:
### Summary

- Add `PseudoLabeling` transform for unlabeled dataset
- Use `Explorer` of Datumaro to make pseudo label for each datasetitem

### How to test

- Add unit tests

### Checklist

### License

```python
#
```

### DIFF
--- a/src/datumaro/plugins/transforms.py
+++ b/src/datumaro/plugins/transforms.py
@@ -22,6 +22,7 @@ import pycocotools.mask as mask_utils
 from pandas.api.types import CategoricalDtype
 
 import datumaro.util.mask_tools as mask_tools
+from datumaro.components.algorithms.hash_key_inference.explorer import Explorer
 from datumaro.components.algorithms.hash_key_inference.hashkey_util import calculate_hamming
 from datumaro.components.annotation import (
     AnnotationType,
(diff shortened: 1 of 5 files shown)


src/datumaro/plugins/transforms.py
```diff
@@ -2004,3 +2007,64 @@ def transform_item(self, item):
...
+    def transform_item(self, item: DatasetItem):
+        hashkey_ = np.unpackbits(self._explorer._get_hash_key_from_item_query(item).hash_key)
+        logits = calculate_hamming(hashkey_, self._label_hashkeys)
+        inverse_distances = 1.0 / (logits + 1e-6)
+        probs = inverse_distances / np.sum(inverse_distances)
+        ind = np.argsort(probs)[::-1]
+
+        pseudo = np.array(self._labels)[ind][0]
+        pseudo_annotation = [Label(label=self._label_indices[pseudo])]
+        return self.wrap_item(item, annotations=pseudo_annotation)
```
jihyeonyi, line 2070: It seems that the original annotations are replaced with the pseudo annotation. Is it intended? Then how about add this notice to the document?

sooahleex: In this case, I assumed the case that each item does not have any annotation. so the original annotations is not existed. If my assumption is not proper, please let me know.

jihyeonyi: Oh, I see. Then, we can enhance this feature when the assumption is changed.

jihyeonyi (approved):

PR #1695, opened 2025-01-16
Add position information to PointsCategories

djdameln:
### Summary

This PR adds optional position information as meta-info for (key)point annotations. The position information can be used to define one or multiple default relative positions of a set of keypoints, for example a set of default human poses (standing, running etc) for human keypoint detection datasets. This type of meta-information about points data is useful in annotation tools.

### Changes

- Adds `positions` attribute to `PointsCategories.Category` class. Similar to the `Points` annotation type, the positions are represented as a list of floats in the format `[x1, y1, x2, y2, ..., xn, yn]`. The positions are optional and default to an empty list.
- Adds a validator that validates the position coordinates that are passed to the `Category` constructor. Also validates if the number of position coordinates is the same as the number of labels.
- Adds a normalization function that normalizes the positions to the [0, 1] range (preserving aspect ratio). The normalization function is called by the validator to ensure that position information in the `PointsCategories` instance is always normalized.
- Adds logic to Datumaro format importer and exporter, to ensure that position information gets stored in the annotations json, and is also loaded correctly during import. The importer can handle datasets where position information is missing, so backward compatibility is preserved.
- Adds unit tests for normalization, validation and import/export functionality described above.

### How to test

Add some position information to the `PointsCategories` instance of a keypoint detection dataset, then try export and importing the dataset.

### Checklist

### License

```python
#
```

### DIFF
--- a/src/datumaro/components/annotation.py
+++ b/src/datumaro/components/annotation.py
@@ -1601,7 +1601,7 @@ class PointsCategories(Categories):
         joints: Set[Tuple[int, int]] = field(factory=set, validator=default_if_none(set))
 
         # Set of default x, y coordinates of the points
-        positions: List[float] = field(factory=list)
+        positions: List[float] = field(default=[])
 
         @positions.validator
         def positions_validator(
@@ -1613,17 +1613,15 @@ class PointsCategories(Categories):
             To be used as an attrs validator for the positions field of PointsCategories.Category.
 
             Args:
-                attribute (attr.Attribute[list]): The attribute being validated (unused).
+                attribute (attr.Attribute[list]): The attribute being validated.
                 positions (list[float]): A list of point positions.
 
             Raises:
                 ValueError: If the provided value cannot be converted to a list of floats,
-                    or if the number of positions is not even.
+                    or if the number of positions is invalid.
             """
-            del attribute
-            print(len(self.labels))
-            if positions is None or positions == []:
-                self.positions = []
+            if positions is None:
+                positions = []
             else:
                 # convert to a list of floats
                 try:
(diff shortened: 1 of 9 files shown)


src/datumaro/components/annotation.py
```diff
@@ -1598,14 +1600,49 @@ class Category:
...
+            """
+            if positions is None or positions == []:
+                self.positions = []
+            else:
+                # convert to a list of tuples
+                try:
+                    positions = list(map(float, positions))
+                except (TypeError, ValueError):
+                    raise ValueError(
+                        f"Cannot convert {attribute.name} to list of floats. Check your input data."
```
A-Artemis: What is attribute referring to here? 
Could you please add the params and return to the doc string, and type hinting please.

src/datumaro/plugins/data_formats/datumaro/base.py
```diff
@@ -115,7 +115,12 @@ def _load_categories(parsed) -> Dict:
         if parsed_points_cat:
             point_categories = PointsCategories()
             for item in parsed_points_cat["items"]:
-                point_categories.add(int(item["label_id"]), item["labels"], joints=item["joints"])
+                point_categories.add(
+                    int(item["label_id"]),
+                    item["labels"],
+                    joints=item["joints"],
+                    positions=item.get("positions"),
```
sovrasov, line 122: `item.get("...")` looks a bit inconsistent vs `item["..."]`. Is there a specific reason to mix different styles of element access?

src/datumaro/components/annotation.py
djdameln: `attribute` is a mandatory argument for this validator method, that gets passed to the method internally by the `attr` library. It contains some meta information about the to be validated attribute (in this case `positions`), such as name, expected type and default value. I added type annotations and some refactoring to make it more clear.

src/datumaro/plugins/data_formats/datumaro/base.py
djdameln: This is to ensure backward compatibility. Datasets that were created with older versions of Datumaro may not have the "positions" key, so retrieving them directly using `item["positions"]` would result in an error. By soft-getting the positions value, we ensure that we can safely load these older datasets.

sovrasov (approved):

djdameln:
Closing this PR, as changes will be added to release branch instead, see https://github.com/openvinotoolkit/datumaro/pull/1702

PR #1701, opened 2025-01-23
Add Python 3.12 to the CI and update Ubuntu

gdlg:
### Summary

This PR adds support for Python 3.12 in the project configuration and CI. I have left the min Python version unchanged (3.9).

### Checklist

### License

```python
#
```


tests/unit/test_framework_converter.py
```diff
@@ -479,20 +473,21 @@ def test_can_convert_torch_framework_detection(self):
                 assert torch_ann["iscrowd"] == dm_ann["attributes"]["is_crowd"]
 
     @pytest.mark.skipif(not TORCH_AVAILABLE, reason="PyTorch is not installed")
-    def test_can_convert_torch_framework_tabular_label(self, fxt_tabular_label_dataset):
+    def test_can_convert_torch_framework_tabular_label(
+        self, fxt_tabular_label_dataset, fxt_text_example
+    ):
         class IMDBDataset(Dataset):
-            def __init__(self, data_iter, vocab, transform=None):
+            def __init__(self, data_iter, tokenizer, transform=None):
```
djdameln: A bit nitpicky, but now that the content of text item is changed, `IMDBDataset` may not be the best name for this dataset anymore. Can you change it to something more meaningful, e.g. `DummyTabularDataset`?

src/datumaro/plugins/framework_converter.py
djdameln, line 1: Somewhat out of scope of this PR, but my concern with this implementation is that it may be unclear to users when to pass just the tokenizer, and when to pass both a tokenizer and a vocab. I think a proper docstring would address this.

Adding the complete dosctring is out of scope, but maybe you could add a brief docstring, or comment, explaining that the vocab arg is not needed when the tokenizer already returns the token ids instead of the raw tokens, and that this depends on which NLP framework is used. Just a brief 1-2 sentence would be sufficient. It's mainly to help whoever adds the full docstring for this class in the future.

djdameln (changes requested):
Thanks a lot for the PR. I only have some minor comments.

src/datumaro/plugins/framework_converter.py
gdlg: Done but I think that the construction of those classes needs to be refactored to be simplified. The parameters of the `to_framework` function depends on which framework is used which makes it difficult to document.

djdameln: Yes, I fully agree. We'll address this at a later point.

djdameln (approved):

PR #1776, opened 2025-07-02
Add deprecation notice to features that will soon be removed

### LINKED ISSUE #1775: Feature deprecation
We would like to deprecate the following features which are seeing little or no use:
* Crypter
* Advanced dataset comparison based on similarity search with OpenVINO
* Synthetic dataset generation
* Data exploration
* BBox to mask using SAM
* Tiling
* Dataset versioning with DVCS
* Telemetry
* Anchor generation
* Missing annotation detection
* Model inference explanation
* Near-duplicate removal
* Pruning
* Pseudo-labels

This will help us reduce the number of dependencies of Datumaro and simplify the code base.

We are still planning to support some of the features within Intel’s vision ecosystem. For instance, tiling is supported by https://github.com/open-edge-platform/training_extensions . We are considering creating a new library in the future for some advanced features such as anchor generation, NDR, noisy-labels, pruning and pseudo-labels which will use https://github.com/open-edge-platform/model_api instead of re-implementing inference inside Datumaro.

For this ticket, use the same approach as Anomalib for deprecation (https://github.com/open-edge-platform/anomalib/blob/main/src/anomalib/utils/deprecation.py). Update the changelog accordingly. We plan to deprecate those features in v1.11 and remove them in v1.12.

AlbertvanHouten:
### Summary
This pull request introduces a new `@deprecated` decorator to mark classes as deprecated and applies it across multiple files to notify users of upcoming removals. Additionally, it includes a unit test to verify the functionality of the decorator. 

Resolves #1775

### Checklist

- [x] I have added the description of my changes into [CHANGELOG](https://github.com/open-edge-platform/datumaro/blob/develop/CHANGELOG.md).​

### License

- [x] I submit _my code changes_ under the same [MIT License](https://github.com/open-edge-platform/datumaro/blob/develop/LICENSE) that covers the project.

```python
#
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Mark pyemd as optional since it does not support Python 3.12
   (<https://github.com/open-edge-platform/datumaro/pull/1770>)
 
+### Deprecations
+- Added deprecation notices to features that will soon be removed
+  (<https://github.com/open-edge-platform/datumaro/pull/1776>)
+
 ## Q1 2025 Release 1.10.0
 
 ### New features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   (<https://github.com/open-edge-platform/datumaro/pull/1770>)
 
 ### Deprecations
-- Added deprecation notices to features that will soon be removed
+- Added deprecation notices to the following features that will soon be removed:
+  - Crypter
+  - Advanced dataset comparison based on similarity search with OpenVINO
+  - Synthetic dataset generation
+  - Data exploration
+  - BBox to mask using SAM
+  - Tiling
+  - Dataset versioning with DVCS
+  - Telemetry
+  - Anchor generation
+  - Missing annotation detection
+  - Model inference explanation
+  - Near-duplicate removal
+  - Pruning
+  - Pseudo-labels
   (<https://github.com/open-edge-platform/datumaro/pull/1776>)
 
 ## Q1 2025 Release 1.10.0
--- a/src/datumaro/components/hl_ops/__init__.py
+++ b/src/datumaro/components/hl_ops/__init__.py
@@ -26,6 +26,7 @@ from datumaro.components.merge import DEFAULT_MERGE_POLICY, get_merger
 from datumaro.components.transformer import ModelTransform, Transform
 from datumaro.components.validator import TaskType, Validator
 from datumaro.util import parse_str_enum_value
+from datumaro.util.deprecation import deprecated
 from datumaro.util.scope import on_error_do, scoped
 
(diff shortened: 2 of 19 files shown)


tests/unit/test_deprecation.py
```diff
@@ -0,0 +1,28 @@
+# Copyright (C) 2022-2025 Intel Corporation
+# LIMITED EDGE SOFTWARE DISTRIBUTION LICENSE
```
gdlg: Update to use Datumaro’s copyright notices.

src/datumaro/util/deprecation.py
```diff
@@ -0,0 +1,24 @@
...
+        original_init = cls.__init__
+
+        @wraps(original_init)
+        def new_init(self, *args, **kwargs):
+            warnings.warn(
+                f"The {cls.__name__} class will be deprecated in version 1.11 "
+                f"and will be removed in version 1.12.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
```
gdlg, line 22: Use the same approach as in Anomalib, taking the version numbers as parameters instead of hard-coding it. We can probably just reuse the code from Anomalib as-is instead of changing it unless there’s a need to do it differently.

CHANGELOG.md
```diff
@@ -29,6 +29,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Mark pyemd as optional since it does not support Python 3.12
   (<https://github.com/open-edge-platform/datumaro/pull/1770>)
 
+### Deprecations
+- Added deprecation notices to features that will soon be removed
+  (<https://github.com/open-edge-platform/datumaro/pull/1776>)
```
gdlg, line 48: Explicitly list all the deprecations in the changelog. One line per deprecation.

src/datumaro/components/hl_ops/__init__.py
```diff
@@ -35,6 +36,7 @@
 __all__ = ["HLOps"]
 
 
+@deprecated()
```
gdlg: Let’s keep this class as it has useful operations like filter and export; though, we may refactor it in the future. We can deprecate the run_model function.

src/datumaro/util/deprecation.py
```diff
@@ -0,0 +1,24 @@
+import warnings
```
gdlg, line 5: Add copyright notice

gdlg (commented):
Thanks Albert, just left a few comments

src/datumaro/util/deprecation.py
AlbertvanHouten: I've deliberately written the code to not be reusable, as this file should be deleted in 1.12. I don't think we should make deprecation utilities part of datumaro.

gdlg: It’s very likely that we will need it in the future so I think that we should keep it.

AlbertvanHouten: Alright, fixed.

gdlg (approved):
LGTM, thanks Albert!

PR #1792, opened 2025-07-14
Deprecate project and add info to cli commands that will be deprecated

AlbertvanHouten:
### Checklist

### License

```python
#
```

### DIFF
--- a/src/datumaro/cli/commands/__init__.py
+++ b/src/datumaro/cli/commands/__init__.py
@@ -28,6 +28,8 @@ __all__ = [
     "get_project_commands",
 ]
 
+deprecated = "[DEPRECATED, will be removed in 1.12]"
+
 
 def get_non_project_commands():
     return [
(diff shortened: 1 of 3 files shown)


src/datumaro/cli/commands/__init__.py
```diff
@@ -28,6 +28,8 @@
     "get_project_commands",
 ]
 
+deprecated = "[DEPRECATED, will be removed in 1.12]"
+
```
gdlg, line 32: We should also mark the project commands as deprecated.

gdlg (commented):
Thanks Albert, just one comment

src/datumaro/cli/commands/__init__.py
AlbertvanHouten: Fixed.

gdlg (approved):
LGTM, thanks Albert!

PR #1804, opened 2025-07-25
Add support for Python 3.13 and MacOS

### LINKED ISSUE #1755: `pip install datumaro` does not work without rust installation
**Describe the bug**

Datumaro cannot be installed without rust. I think this is not ideal. A user should be able to install and use datumaro as a standalone python package. If rust provided additional features, this could be handled properly. 

**Steps to Reproduce**

1. Use a machine where `rust` is not installed
2. Install datumaro via `pip install datumaro`
3. Installation will fail complaining that rust is not installed on the system.

**Environment:**

- OS: Ubuntu 24.04, MacOS
- Datumaro version: 1.10, 1.9.1
- Python version:

gdlg:
### Summary

This PR adds support for Python 3.13 and MacOS. No changes to Datumaro itself but I had to disable the tensorflow tests for Python 3.13 since Tensorflow doesn’t suport 3.13 yet.

In theory, the tests already had support for running without Tensorflow and PyTorch; however, in practice, I found a few issues especially with parameterised test. So I have addressed those issues.

I have disabled the test test_can_release_resources_on_checkout on Windows because it is failing with Python 3.13 and the tested feature is already deprecated.

Resolves #1755

### Checklist

### License

```python
#
```

### DIFF
--- a/.github/workflows/health_check.yml
+++ b/.github/workflows/health_check.yml
@@ -27,6 +27,8 @@ jobs:
             tox-env-py: "311"
           - python-version: "3.12"
             tox-env-py: "312"
+          - python-version: "3.13"
+            tox-env-py: "313"
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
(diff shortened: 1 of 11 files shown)


samet-akcay (approved):
thanks for the effort!

Issue #1789, opened 2025-07-10
Backward compatibility between the old and new datasets


gdlg:
Add functions to convert back and forth between the current Datumaro dataset and the new experimental one. Functions must be able to convert between datasets and also between DatasetItem and Sample.

- [x] Conversion to the new Dataset class
- [ ] Conversion from the new Dataset class

PR #1810, opened 2025-08-06
Implement conversion to the experimental Dataset format.

gdlg:
### Summary

This PR implements the conversion from the legacy to the experimental Dataset class. I will implement the conversion back to the legacy class in a separate PR.

Misc fixes:
* Also implements `__len__`, `__delitem__` and `__iter__` in the Dataset class.
* Fix bug when fetching ann_types() before the cache is initialised.

The conversion works in two steps: the first step analyse the existing dataset and generate the schema for the new dataset. The second step actually converts the data.

I have defined MediaConverter and AnnotationConverter base class which can be extended to support new media/annotation types. This PR implements the conversion logic but the conversion for specific media/annotation type will be implemented later. 

Part of #1789

### Checklist

- [x] I have added the description of my changes into [CHANGELOG](https://github.com/open-edge-platform/datumaro/blob/develop/CHANGELOG.md).​

### License

```python
#
```

### DIFF
--- a/tests/unit/experimental/test_dataset.py
+++ b/tests/unit/experimental/test_dataset.py
@@ -322,3 +322,110 @@ def test_dataset_lazy_converters_property():
     assert isinstance(lazy_converters, list)
     # Initially should be empty
     assert len(lazy_converters) == 0
+
+
+def test_dataset_len():
+    """Test __len__ method returns correct dataset size."""
+
+    class TestSample(Sample):
+        image: np.ndarray[Any, Any] = image_field(dtype=pl.UInt8, format="RGB")
+        bbox: np.ndarray[Any, Any] = bbox_field(dtype=pl.Float32, normalize=False)
+        image_info: ImageInfo = image_info_field()
+
+    dataset = Dataset(TestSample)
+
+    # Initially empty
+    assert len(dataset) == 0
+
+    # Add one sample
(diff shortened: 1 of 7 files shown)


tests/unit/experimental/test_dataset.py
```diff
@@ -322,3 +322,110 @@ class TestSample(Sample):
...
+    assert len(dataset) == 2
+
+    # Add third sample
+    sample3 = TestSample(
+        image=np.array([[[128, 64, 192]], [[96, 160, 32]]], dtype=np.uint8),
+        bbox=np.array([[0.9, 0.8, 0.7, 0.6]], dtype=np.float32),
+        image_info=ImageInfo(width=1, height=2),
+    )
+    dataset.append(sample3)
+    assert len(dataset) == 3
```
AlbertvanHouten: Adding this third sample seems redundant after having already tested two appends. It would make more sense to remove one here and test if the len still works properly.

AlbertvanHouten (approved):

tests/unit/experimental/test_dataset.py
gdlg: Done, I have also implemented `__delitem__` for that.

AlbertvanHouten (approved):

Issue #1785, opened 2025-07-10
New Datumaro dataset structures


gdlg:
This issue aggregates all the tasks related to refactoring of the core dataset structure of Datumaro.

Datumaro has its own annotation format implemented using Python classes. While this format is quite flexible, it requires conversion when integrating with external tools/libraries. Those conversions can be expensive and cumbersome. Processing inside Datumaro is quite slow because those data structures are inefficient compared to Numpy arrays. Preliminary benchmarks show that we can optimise some operations by 3x–10x using arrays instead of the current data structure.

We would like to replace this annotation format with one based on data frames which simplifies the integration with other libraries (e.g. PyTorch) and reduce operation overhead by allowing zero-copy conversions to Numpy arrays or PyTorch tensors. The new format is intended to be flexible, allowing for different data representations, while also enabling conversions between those representations. For instance, the new format must allow easy conversion between normalised and denormalised coordinates, or between masks and polygon representations for segmentation.

Such refactoring will introduce breaking changes to Datumaro but those changes will be primarily on the parts of the Datumaro API used for direct access handling of annotations and dataset items. High-level functionalities such as import/export, data format conversions, filtering, etc will be preserved (though, the API may still have smaller breaking changes).

PR #1823, opened 2025-08-19
Add categories to the experimental dataset class.

gdlg:
### Summary

This PR adds the notion of categories to the new experimental dataset class. The approach is the same as the existing dataset class. Categories are tied to specific schema attributes.

Part of #1785

### Checklist

### License

```python
#
```

### DIFF
--- a/src/datumaro/experimental/categories.py
+++ b/src/datumaro/experimental/categories.py
@@ -1,0 +1,298 @@
+# Copyright (C) 2025 Intel Corporation
+#
+# SPDX-License-Identifier: MIT
+
+"""
+Categories definitions for the experimental dataset system.
+
+This module provides category management functionality using standard dataclasses
+instead of attrs, taking inspiration from the original Categories implementation.
+"""
+
+from __future__ import annotations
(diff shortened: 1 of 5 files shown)


src/datumaro/experimental/categories.py
```diff
@@ -0,0 +1,298 @@
...
+from enum import IntEnum
+from typing import Dict, Iterator, List, NamedTuple, Optional, Tuple, Union
+
+
+class GroupType(IntEnum):
+    """Types of label groups for organizing labels."""
+
+    EXCLUSIVE = 0
+    INCLUSIVE = 1
+    RESTRICTED = 2
```
AlbertvanHouten, line 24: What is a restricted group type? Would be good to have a description for these options.

gdlg: The problem is that this is undocumented in the original file. I would suggest to leave it for now and revisit if we still need it after having progressed on the OTX integration for classification.

AlbertvanHouten (approved):

PR #1834, opened 2025-08-20
Additional fields and type inferring for union

AlbertvanHouten:
### Summary

This pull request introduces robust support for Python `Union` types in the experimental Datumaro type registry and dataset schema inference. It enables seamless conversion between multiple candidate types (including both `typing.Union` and modern `A | B` syntax), with fallback logic and comprehensive test coverage. The changes also improve image type conversion and schema inference for datasets, making the system more flexible and reliable.

### Type registry and conversion improvements

* Added full support for `Union` types in the type registry: both `typing.Union` and Python 3.10+ `A | B` syntax are now handled, with fallback to subsequent types if the first conversion fails. This includes updated logic in `from_polars_data` and new tests for ordering, error handling, and fallback behavior. [[1]](diffhunk://#diff-e324261812079d99ca2989612441e5df1dd15dabde37fb2e5e8c0c1b639dac0dR122-R154) [[2]](diffhunk://#diff-e324261812079d99ca2989612441e5df1dd15dabde37fb2e5e8c0c1b639dac0dR170-R269) [[3]](diffhunk://#diff-30f23b2869128577a39c918ed25c78229a30cb96578c33728d45e5ebce740ac2R1-R162)
* Added comprehensive tests for type registry conversions, including basic types, union types, error cases, ordering, and converter functionality for numpy and torch tensors.

### Dataset and schema inference enhancements

* Improved schema inference in `Dataset` to resolve string annotations to actual type objects, supporting cases where `from __future__ import annotations` is used, and added correct handling for `Union` types to preserve the original annotation. [[1]](diffhunk://#diff-4ac196ddc4dc8e6d33daf684ded18886ff8774fadb8b6cbd4bfa88ca424bb34fR65-R80) [[2]](diffhunk://#diff-4ac196ddc4dc8e6d33daf684ded18886ff8774fadb8b6cbd4bfa88ca424bb34fR94-R110)
* Updated type variable definitions and method signatures in `dataset.py` for clarity and correctness, and removed unnecessary imports. [[1]](diffhunk://#diff-4ac196ddc4dc8e6d33daf684ded18886ff8774fadb8b6cbd4bfa88ca424bb34fR19-R25) [[2]](diffhunk://#diff-4ac196ddc4dc8e6d33daf684ded18886ff8774fadb8b6cbd4bfa88ca424bb34fL105-R128) [[3]](diffhunk://#diff-4ac196ddc4dc8e6d33daf684ded18886ff8774fadb8b6cbd4bfa88ca424bb34fL134-R157)

### API and import improvements

* Updated the experimental module’s public API to expose new converters, dataset classes, fields, schema types, and registry functions.

### Test coverage

* Added targeted tests for union type handling in dataset samples, ensuring both modern and legacy union syntax are supported.

These changes significantly improve the flexibility and reliability of type conversion and schema inference in Datumaro’s experimental pipeline.

### Checklist

### License

```python
#
```


src/datumaro/experimental/dataset.py
```diff
@@ -76,15 +101,23 @@ def infer_schema(cls) -> Schema:
...
+                final_type = annotation
+            else:
+                final_type = type_origin if type_origin is not None else annotation
             attributes[name] = AttributeInfo(type=final_type, annotation=field_annotation)
         return Schema(attributes=attributes)
 
 
-DType = TypeVar("DType", bound=Sample, default=Sample)
-DTargetType = TypeVar("DTargetType", bound=Sample, default=Sample)
+DType = TypeVar("DType", bound=Sample)
```
AlbertvanHouten, line 107: This change is necessary for proper type hints in IDEs for Sample subclasses

src/datumaro/experimental/converters.py
```diff
@@ -343,3 +343,71 @@ def op(x: pl.Expr, y: pl.Expr) -> pl.Expr:
...
+@converter
+class ImageTypeConverter(Converter):
+    """
+    Converter that transforms images between different types (numpy arrays, PIL Images, torch tensors).
+    This converter enables switching between different image representations while preserving
+    the image content. It supports conversions between:
+    - numpy.ndarray
+    - PIL.Image.Image
+    - torch.Tensor
+    """
```
gdlg: This class does quite make sense to me because when we store the data in the Polars dataframe as a Polars list, so Polars does not see the Numpy/PIL/Torch tensors. Because of this, the conversion function can just convert between those data types without having to changing the data.

The implementation of this function seems to just rename the columns. Currently, we are lacking support for renaming in the Dataset.convert_to_schema. We are probably lacking support for converting data type between Numpy, PyTorch, etc as well. However, I think that we can implement it in a generic way rather than tying it specifically to images.

For the conversion between Polars and PyTorch/Numpy/etc, I believe your changes in type_registry.py should be sufficient.

That being said, it would be useful to have a function for converting between the different image dtypes (uint8, uint16, float, etc), formats (RGB, BGR), and data normalisation. So I would suggest to keep this class but refocus it on those conversions (we can keep as empty placeholder for now if you don’t want to work on those in this PR).

src/datumaro/experimental/dataset.py
gdlg: What error do you get? Which IDE? As far as I remember, the default is needed in MyPy for the case where we pass a schema to the constructor instead of the dtype.

When passing a schema, the DType is not explicitly passed to the constructor (since there is no dtype) and in this case, we want to default to a `Dataset[Sample]`. Otherwise Mypy and Pylance deduce the type as `Dataset[Any]` and start complaining.

If there’s no way around from an IDE standpoint, we might need to rework the constructor to remove the overload and instead use a classmethod.

src/datumaro/experimental/dataset.py
```diff
@@ -61,8 +62,22 @@ def infer_schema(cls) -> Schema:
         Raises:
             TypeError: If attributes don't have proper Field annotations
         """
+        import sys
```
gdlg: Move the import to the top of the file.

gdlg (commented):
Thanks Albert! First round of review, I still need to review the part on the type union.

src/datumaro/experimental/dataset.py
AlbertvanHouten: image

In the above image, it doesn't properly deduce that Sample is a ClassificationSample if we set the default to Sample for DType. This is in Pycharm.

src/datumaro/experimental/dataset.py
```diff
@@ -76,15 +92,22 @@ def infer_schema(cls) -> Schema:
                 raise TypeError(f"Attribute '{name}' must have a Field annotation.")
 
             # Extract base class from generic types like MyClass[A, B, C] -> MyClass
+            # Special handling for Union types - preserve the Union instance itself
```
gdlg: This comment is redundant. Let’s remove it.

src/datumaro/experimental/dataset.py
```diff
@@ -76,15 +92,22 @@ def infer_schema(cls) -> Schema:
...
+            else:
+                final_type = type_origin if type_origin is not None else annotation
             attributes[name] = AttributeInfo(type=final_type, annotation=field_annotation)
         return Schema(attributes=attributes)
 
 
-DType = TypeVar("DType", bound=Sample, default=Sample)
-DTargetType = TypeVar("DTargetType", bound=Sample, default=Sample)
+DType = TypeVar("DType")
+DTargetType = TypeVar("DTargetType")
```
gdlg: Add a comment in the code explaining why removing them. Mention the issue with Dataset(schema) too.

src/datumaro/experimental/dataset.py
```diff
@@ -281,10 +304,6 @@ def convert_to_schema(
         Returns:
             A new Dataset instance with the converted schema
         """
-        # Import the converter implementations to register them
-        # ruff: noqa: F401
-        import datumaro.experimental.converters  # pyright: ignore [reportUnusedImport, reportMissingImports]
```
gdlg, line 286: This line is important. Without it, the converters won’t be registered. So we need to keep it.

src/datumaro/experimental/fields.py
```diff
@@ -21,6 +21,13 @@
 T = TypeVar("T")
 
 
+@dataclass
+class Label:
+    """Minimal Label annotation class."""
+
+    label: Any  # Can be int, str, or other types for flexibility
```
gdlg: This class is not needed. When creating a sample, the user can specify whatever type they want. E.g.
```
class MySample(Sample):
    label: int = label_field()
```

In this example, we don’t use a label class. The data is directly converted to int.

src/datumaro/experimental/fields.py
```diff
@@ -246,3 +253,88 @@ def image_path_field(semantic: Semantic = Semantic.Default) -> Any:
...
+            return {name: self.dtype}
+
+    def to_polars(self, name: str, value: Any) -> dict[str, pl.Series]:
+        """Convert label(s) to Polars format, handling both single and multi-label cases."""
+        if value is None:
+            return {name: pl.Series(name, [None], dtype=self.dtype)}
+
+        # Handle single Label object
+        if hasattr(value, "label"):
+            return {name: pl.Series(name, [value.label], dtype=self.dtype)}
```
gdlg: This case is not needed. Apart from special fields like ImageInfo, we want to avoid structs as much as possible.

src/datumaro/experimental/fields.py
```diff
@@ -246,3 +253,88 @@ def image_path_field(semantic: Semantic = Semantic.Default) -> Any:
...
+
+        # Handle numpy array or list (multi-label)
+        elif isinstance(value, (np.ndarray, list)):
```
gdlg: Use `self.multi_label` to determine the case instead of checking the input type.

The problem with checking the input type is that it restricts the input type to a list of known types preventing extensions to new types (like PyTorch Tensors or TensorFlow). So this code has to remain independent of the input type.

Instead use the functions from type_registry.py to do the conversions. Have a look at the implementation of TensorField for example.

In this example, you convert numpy arrays to a Python list. The problem is that Python lists are not the most efficient and most conversions will result in a copy. Instead convert the input type to Numpy because most conversions to Numpy are zero-copy. For instance, PyTorch to Numpy has little overhead while PyTorch to a list is expensive. See TensorField for example.

src/datumaro/experimental/legacy.py
```diff
@@ -185,7 +185,7 @@ def analyze_legacy_dataset(legacy_dataset: LegacyDataset) -> AnalysisResult:
     ann_types = legacy_dataset.ann_types()
 
     attributes: dict[str, AttributeInfo] = {}
-    media_converter: ForwardMediaConverter | None = None
+    media_converter: ForwardMediaConverter | None
```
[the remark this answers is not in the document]
gdlg: I believe some linters don’t like when `= None` is not used.

src/datumaro/experimental/type_registry.py
```diff
@@ -118,9 +119,39 @@ def from_polars_data(polars_data: Any, target_type: type) -> Any:
...
+    except Exception:
+        pass
+
+    if is_union and union_args:
+        # Try each type in the union until one succeeds
+        for union_type in union_args:
+            if union_type in _from_polars_converters:
+                try:
+                    return _from_polars_converters[union_type](polars_data)
+                except Exception:
```
gdlg: Check for KeyError but any other exception should be raised to the user.

src/datumaro/experimental/type_registry.py
```diff
@@ -136,3 +167,103 @@ def from_polars_data(polars_data: Any, target_type: type) -> Any:
...
     pass
+
+
+# Register PIL Image converters if available
+try:
+    from PIL import Image
+
+    register_numpy_converter(Image.Image, lambda x: np.array(x))
+    register_from_polars_converter(
+        Image.Image, lambda x: Image.fromarray(np.array(x).astype(np.uint8))
```
gdlg: Do we have to convert to uint8? Does PIL support other types?

src/datumaro/experimental/legacy.py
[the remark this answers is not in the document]
AlbertvanHouten: Settings this to None will result in an unused variable assignment warning, since it is overwritten in the try/except block at line 192.

src/datumaro/experimental/type_registry.py
AlbertvanHouten: I believe PIL only supports uint8 for RGB images. In that case, astype will not do anything since it's already the right type.

src/datumaro/experimental/fields.py
```diff
@@ -246,3 +246,86 @@ def image_path_field(semantic: Semantic = Semantic.Default) -> Any:
...
+        """Convert label(s) to Polars format, handling both single and multi-label cases."""
+        if value is None:
+            return {name: pl.Series(name, [None], dtype=self.dtype)}
+
+        # Handle numpy array or list (multi-label)
+        elif self.multi_label:
+            if isinstance(value, np.ndarray):
+                value_list = value.tolist()
+            elif isinstance(value, list):
+                value_list = value
```
gdlg: Convert to Numpy instead of to a list. You can use the `to_numpy` function from type_registry.py

src/datumaro/experimental/fields.py
```diff
@@ -246,3 +246,86 @@ def image_path_field(semantic: Semantic = Semantic.Default) -> Any:
...
+            else:
+                raise TypeError(f"Expected value to be a list or np.ndarray, got {type(value)}")
+            return {name: pl.Series(name, [value_list], dtype=pl.List(self.dtype))}
+
+        # Handle single integer value
+        elif isinstance(value, (int, np.integer)):
```
gdlg:
```suggestion
        else:
```

In this case, you can assume that we are working with a single label. No need to check the input type.

src/datumaro/experimental/fields.py
```diff
@@ -246,3 +246,86 @@ def image_path_field(semantic: Semantic = Semantic.Default) -> Any:
...
+    def from_polars(self, name: str, row_index: int, df: pl.DataFrame, target_type: type[T]) -> T:
+        """Reconstruct label(s) from Polars data."""
+        data = df[name][row_index]
+
+        # If data is a list, it's multi-label
+        if isinstance(data, list):
+            if target_type == np.ndarray or target_type is np.ndarray:
+                return np.array(data, dtype=np.int64)
+            elif target_type is list:
+                return data
```
gdlg: Same here, use `from_polars_data` from type_registry.py

gdlg (approved):
LGTM, thanks Albert!

PR #1840, opened 2025-08-22
Update the documentation to remove deprecated features

gdlg:
### Summary

This PR updates the documentation to remove deprecated features. I have fixed the documentation of the download command which was outdated. I have also added a new "kaggle" extras dependency group to install the correct dependencies for downloading from Kaggle. I also fixed the description of some commands and well as some error messages.

I also disabled the git-check pre-commit because it conflicts with .rst files and I couldn’t find a way to ignore those; we can re-enable it if we find a better way.

### Checklist

### License

```python
#
```


docs/source/docs/command-reference/stats.md
```diff
@@ -1,40 +1,40 @@
...
+- `--ann-stats` (bool) - Compute annotation statistics (default: True)
+- `-h, --help` - Print the help message and exit
 
-Example:
-- Compute project statistics
+## Example
+- Compute dataset statistics
   ```console
-  datum stats -p <path/to/project/>
+  datumaro stats /path/to/dataset:coco
```
AlbertvanHouten: The previous files in this PR all use `datum` as the command name. Why is this one different?

gdlg: Good catch, I have battled pretty hard with Copilot to avoid it doing this change. The issue is that the Python package is called `datumaro` and when calling the CLI help using Python (i.e. `python -m datumaro convert --help`), the first line says `Usage: datumaro convert ...` which is really confusing.

I am almost thinking of renaming the CLI to `datumaro` instead of `datum` to resolve the confusion. Anyway, for now, I have reverted it.

AlbertvanHouten (approved):

PR #1866, opened 2025-09-03
Use uv for dependency management

AlbertvanHouten:
This pull request modernizes the build and dependency management for the project by migrating from `setup.py` and requirements files to a fully declarative `pyproject.toml` configuration and switching the recommended workflow to use the `uv` tool for dependency management and testing. It also updates Python version support to require 3.10+ (3.9 EOLs in October and was causing some problems). The C++ compiler is now also optional, as a fallback was added to use python implementation in case it is not installed. 

### Checklist

- [ ] I have added the description of my changes into [CHANGELOG](https://github.com/open-edge-platform/datumaro/blob/develop/CHANGELOG.md).


.github/workflows/security-scan.yml
```diff
@@ -78,16 +78,11 @@ jobs:
         with:
           python-version: "3.11"
 
-      - name: Install dependencies
-        run: python -m pip install pip-tools
-
-      - name: Freeze dependencies
+      - name: Prepare dependency files for scanning
```
AlexanderBarabanov: Please remove this step and previous `Set up Python` as uv.lock will be available for Trivy scan after repo checkout step - no additional steps are required.

src/datumaro/util/mask_tools.py
```diff
@@ -221,7 +220,12 @@ def lazy_mask(path, inverse_colormap=None):
 
 
 def mask_to_rle(binary_mask):
-    return encode(binary_mask)
```
gdlg, line 224: I am concerned about unintentional consequences. If for some reason, we accidentally fail to load the _capi module and execute the encode() function, we may run a slower implementation without realising it. So I think that we should revert this change.

Instead we can create another PR to migrate to use Rust instead of C for the implementation.

pyproject.toml
```diff
@@ -2,14 +2,135 @@
...
+    "shapely>=1.7",
+    "typing_extensions>=3.7.4.3",
+    "tqdm>=4.67.1", # version with the latest security fixes
+    "pycocotools>=2.0.4",
+    "PyYAML==6.0.2",
+    "polars>=1.31.0,<2.0.0",
+    "pandas>=1.4.0",
+    "pyarrow",
+    "json-stream",
+    "opencv-python-headless>=4.11.0.86",
```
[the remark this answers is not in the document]
AlbertvanHouten: DATUMARO_HEADLESS installation option was removed.

pyproject.toml
```diff
@@ -2,14 +2,136 @@
...
+    "defusedxml>=0.7.0",
+    "imagesize>=1.4.1",
+    "lxml<7,>=5.2.0",
+    "numpy>=1.23.4,<2.2.0",
+    "orjson>=3.10.0",
+    "Pillow>=10.3.0",
+    "ruamel.yaml>=0.17.0",
+    "shapely>=1.7",
+    "typing_extensions>=3.7.4.3",
+    "tqdm>=4.67.1", # version with the latest security fixes
```
gdlg: Can you remove this comment please? It doesn’t make sense. If we start put it for tqdm, then to be consistent, we would have to add it to all dependencies.

tox.ini
```diff
@@ -77,11 +104,10 @@ commands =
 deps =
     {[testenv:fuzzing]deps}
 extras =
+    default: default
     tf: tf
     tfds: tfds
     torch: torch
-    cli
-    experimental
```
gdlg, line 84: Are you sure? Did you tested it? When I tried, it didn’t work without those.

AlbertvanHouten: Let me add this back, this was unintended from merge conflict.

pyproject.toml
AlbertvanHouten: I agree, these comments are not helpful

gdlg (dismissed):
LGTM, thanks Albert!

gdlg (approved):
LGTM, thanks Albert!

Ewoodss:
Could a version of datumaro be released that includes these changes? right know i have to-do a weird work around in a docker container with adding rust and building the package myself. the last version is from a day before this change was made. even if only a pre-release version that would be really helpfull

AlbertvanHouten:
> Could a version of datumaro be released that includes these changes? right know i have to-do a weird work around in a docker container with adding rust and building the package myself. the last version is from a day before this change was made. even if only a pre-release version that would be really helpfull

We are currently working on a 2.0 version of datumaro which will include uv for dependency management. We chose not to include this anymore in the 1.x release of datumaro. We hope to release the next version within a month or so.

leoll2:
@Ewoodss could you share more details about your reasons to build the package from sources?

As Albert said, our development efforts are currently focused on a completely reworked version of Datumaro (v2.0, check the [`experimental`](https://github.com/open-edge-platform/datumaro/tree/develop/src/datumaro/experimental) module for reference) and we are not planning further 1.x releases.
However, if you're interested in a few specific bugfixes, I can create and publish patch release for 1.12 (v1.12.1) that includes them.

PR #1882, opened 2025-09-17
Lazy-loading during attribute access

gdlg:
This PR changes the lazy-loading behaviour such that lazy-loading happens when a sample attribute is accessed rather than when fetching the sample. This allows for a more fine-grained approach where attributes such as images are fetched only if they are actually used.

For instance, if a piece of code only need to access the sample labels, the sample image will not be loaded, thus improving the performance.

E.g.:
```
sample = dataset[0] # Old behaviour would load image here.
sample.label # At this point, the old behaviour would already have loaded sample.image even if it’s not used. The new behaviour sets the label without having to load the image, thus improving performance.
sample.image # The new behaviour loads the image here.
```

To do so, we track which converters are needed for which attribute, then when an attribute is accessed, we run the required converters. We distinguish between direct attributes (like labels) which we don’t require any converters and lazy attributes. Direct attributes are available as plain object attribute, no magic. Lazy attributes behaves like a class `@property` calling the a function to evaluate the attribute value. Once a lazy attribute has been evaluated, the computed value is stored in the sample class and it behaves like a direct attribute.

The existing implementation of attribute renaming was problematic because the renaming converter was inserted as the last converter and all attributes had a dependency on it which prevents per-attribute lazy loading. To fix this problem, this PR reworks attribute renaming to insert the converter as the first converter. Since this converter is the first, it can also be executed as a batch converter.

**Refactoring of converter pathfinding and lazy converter handling:**

* The logic for attribute renaming and deletion during schema conversion has been reworked. Now, an initial attribute remapping step is introduced at the start of the conversion process, and post-processing steps for renaming/deletion have been removed for clarity and correctness. (_create_initial_renaming_converter replaces _create_post_processing_for_semantic, and related changes in `_find_conversion_path_for_semantic`) [[1]](diffhunk://#diff-fb2a908f10fa67b50f19323df8324a854db0618c5c61595394aa828af411ca41L638-R677) [[2]](diffhunk://#diff-fb2a908f10fa67b50f19323df8324a854db0618c5c61595394aa828af411ca41L714-R715) [[3]](diffhunk://#diff-fb2a908f10fa67b50f19323df8324a854db0618c5c61595394aa828af411ca41L741-R731) [[4]](diffhunk://#diff-fb2a908f10fa67b50f19323df8324a854db0618c5c61595394aa828af411ca41L769-R755)
* The handling of lazy converters has been improved: instead of a flat list, lazy converters are now tracked as a dictionary mapping output attribute names to lists of converters, allowing for more precise and efficient application of lazy conversions. This affects both the return type of `ConversionPaths` and the internal logic for separating batch and lazy converters. [[1]](diffhunk://#diff-fb2a908f10fa67b50f19323df8324a854db0618c5c61595394aa828af411ca41L54-R54) [[2]](diffhunk://#diff-fb2a908f10fa67b50f19323df8324a854db0618c5c61595394aa828af411ca41R824) [[3]](diffhunk://#diff-fb2a908f10fa67b50f19323df8324a854db0618c5c61595394aa828af411ca41L856-R833) [[4]](diffhunk://#diff-fb2a908f10fa67b50f19323df8324a854db0618c5c61595394aa828af411ca41L887-R900)

**Dataset and Sample class improvements:**

* The `Sample` and `Dataset` classes have been updated to support the new lazy converter structure, storing and exposing lazy converters as dictionaries keyed by attribute name. The initialization and from_dataframe logic have been updated accordingly. [[1]](diffhunk://#diff-4ac196ddc4dc8e6d33daf684ded18886ff8774fadb8b6cbd4bfa88ca424bb34fR48-R55) [[2]](diffhunk://#diff-4ac196ddc4dc8e6d33daf684ded18886ff8774fadb8b6cbd4bfa88ca424bb34fL153-R172) [[3]](diffhunk://#diff-4ac196ddc4dc8e6d33daf684ded18886ff8774fadb8b6cbd4bfa88ca424bb34fL169-R190) [[4]](diffhunk://#diff-4ac196ddc4dc8e6d33daf684ded18886ff8774fadb8b6cbd4bfa88ca424bb34fL186-R200)
* The `Sample` class now tracks applied converters and stores schema and dataframe references for dynamic property loading, enabling more flexible and efficient lazy evaluation.

**Other improvements and minor changes:**

* The `AttributeRemapperConverter` now uses `rename()` to rename columns and keeps all other columns, instead of selecting only mapped columns.
* The image converter now creates output columns with explicit dtype using schema information, improving compatibility and correctness.
* Minor code cleanup and improvements, such as more precise `__repr__` output for `Sample` and updated imports. [[1]](diffhunk://#diff-4ac196ddc4dc8e6d33daf684ded18886ff8774fadb8b6cbd4bfa88ca424bb34fL56-R68) [[2]](diffhunk://#diff-4ac196ddc4dc8e6d33daf684ded18886ff8774fadb8b6cbd4bfa88ca424bb34fR9) [[3]](diffhunk://#diff-4ac196ddc4dc8e6d33daf684ded18886ff8774fadb8b6cbd4bfa88ca424bb34fL17-R19)

### Checklist

- [ ] I have added the description of my changes into [CHANGELOG](https://github.com/open-edge-platform/datumaro/blob/develop/CHANGELOG.md).


src/datumaro/experimental/converters.py
```diff
@@ -224,19 +224,23 @@ def convert(self, df: pl.DataFrame) -> pl.DataFrame:
 
                 # Convert to numpy array
                 img_array = np.array(img, dtype=np.uint8)
-                image_data.append(img_array.flatten().tolist())
+                image_data.append(img_array.flatten())
```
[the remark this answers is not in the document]
gdlg: This change is intended. Converting to a Python list leads to inefficiencies and wrong dtype because Python does not support uint8. We need to pass the Numpy array to Polars in order to keep the correct dtype.

src/datumaro/experimental/converter_registry.py
```diff
@@ -635,63 +629,52 @@ def _group_fields_by_semantic(schema: Schema) -> dict[Semantic, _SchemaState]:
...
-            if final_attr_spec.name != target_attr_spec.name:
+    for field_type, start_attr_spec in start_state.field_to_attr_spec.items():
+        if field_type in target_state.field_to_attr_spec:
+            target_attr_spec = target_state.field_to_attr_spec[field_type]
+            if start_attr_spec.name != target_attr_spec.name:
                 converter_needed = True
+                new_name = target_attr_spec.name
+                # If the new name would conflict with another attribute in the target_state, use a temporary name
+                if new_name in used_names:
+                    new_name = f"{new_name}_temp_{id(start_attr_spec)}"
```
AlbertvanHouten, line 683: In what case can we get a duplicate attribute name? Should this not raise an error?

gdlg: We can get a conflict between the naming in the source and target state. I think that it’s a bit unlikely in real use cases but the corner case remain. For instance, consider the following:
```
class SourceSample:
    image = image_path_field(...)

class TargetSample:
    image = image_field(...)
    image_info = image_info_field(...)
```

The `image` is used for the image_path_field in the source but image_field in the target. If the source is not renamed to a temporary value, the `image` column of the source will be overwritten by the converter to image_field. If for some reason, those columns are also needed by another converter (e.g. image_info_field), it will no longer be available since the column is now used by image_field. This example doesn’t quite work in practice because we use one converter to generate both image_field and image_info_field but if it were separate ones, it could happen. Once all the converters are processed, the temp column is no longer needed (hence the temporary nature).

src/datumaro/experimental/converter_registry.py
```diff
@@ -635,63 +629,52 @@ def _group_fields_by_semantic(schema: Schema) -> dict[Semantic, _SchemaState]:
     }
 
 
-def _create_post_processing_for_semantic(
-    final_state: _SchemaState, target_state: _SchemaState
-) -> Tuple[List[Converter], _SchemaState]:
+def _create_initial_renaming_converter(
+    start_state: _SchemaState, target_state: _SchemaState
+) -> Tuple[Optional[AttributeRemapperConverter], _SchemaState]:
```
samet-akcay, line 654: it's a bit nitpicky, and not related to this pr. but when renaming `AttributeRemapperConverter`, does `Converter` suffix needed?

samet-akcay: ah never mind, looks like a convention choice. I see that all the converter implementations follow this logic.

src/datumaro/experimental/dataset.py
```diff
@@ -227,18 +240,102 @@ def __getitem__(self, row_idx: int) -> DType:
...
+                )
+
+        # Create the sample with direct attributes, dataframe, and lazy converter info
+        lazy_converters = self._lazy_converters
+        dtype = self._dtype
+
+        # If there are lazy converters, create a dynamic class with descriptors
+        if lazy_converters:
+
+            @dataclass
```
samet-akcay: my only concern here is that the following lazy converter related class implementations somewhat break the `__getitem__` logic. It becomes hard to follow what's happening in the actual method. 

Would it be possible, and make sense, if these are moved somewhere else. eg., top of the module, in a separate module or a private class within Dataset?

samet-akcay (commented):
Thanks @gdlg. Overall it looks good to me. I only have a single concern, which I highlighted in the code: with `LazyState` and `LazyDescriptor` classes, `__getitem__` becomes a bit hard to follow/maintain. I was wondering if it would be possible to move it somewhere else?

src/datumaro/experimental/dataset.py
gdlg: I have moved it outside of the class. I am thinking of refactoring it in a subsequent PR to move this logic out of the dataset class entirely. Tiling and filtering bring their  own set of requirements for supporting dataset transforms and I would like to avoid adding more complexity to the dataset class.

samet-akcay (approved):
lgtm, thanks a lot for the great work!

PR #1905, opened 2025-09-30
Tiling support

gdlg:
This PR adds a new tiling transform. It also introduces a unified mechanism to represent transforms like tiling, converters, and filtering. Transforms can be chained and evaluated lazy.

The tiling transform can be used like so:
```
    # Create tiling config
    config = TilingConfig(tile_width=100, tile_height=100, overlap_x=20, overlap_y=20)

    # Create and apply tiling transform
    tiling_transform = create_tiling_transform(config)
    tiled_dataset = sample_dataset.transform(tiling_transform)
```

### Checklist


src/datumaro/experimental/tiling/tiler_registry.py
```diff
@@ -0,0 +1,654 @@
...
+    # Create tilers for each field type
+    tilers = create_tilers(schema, threshold_drop_ann)
+
+    tilers_filter_group_id = {}
+    tilers_by_group_id = defaultdict(list)
+    tilers_instance_by_name = {}
+
+    # Group filterable tilers
+    for semantic_tilers in tilers.values():
+        filterable_tilers = [tiler for tiler in semantic_tilers if tiler[1].is_filterable()]
```
AlbertvanHouten: The tiler[0]/[1] elements being accessed in this function is a bit hard to read. Could you extract the fields/tilers into separate variables when you retrieve them from create_tiler() on line 376? Maybe creating a class would be better to hold this data.

src/datumaro/experimental/tiling/tilers.py
```diff
@@ -0,0 +1,422 @@
...
+        return True
+
+    def tile(self, df: pl.DataFrame, tiles_df: pl.DataFrame) -> pl.DataFrame:
+        """Process bounding boxes for each tile."""
+        column_name = self.field_spec.name
+        results = []
+
+        for tile_row in tiles_df["tile"]:
+            image_id = tile_row["source_sample_idx"]
+            boxes = df[column_name][image_id]  # List of [x1, y1, x2, y2]
```
AlbertvanHouten: Should we add a check here if the format of the BBoxField is actually x1y1x2y2? I know we currently do not support other representations, but if we do add that it might break tiling unexpectedly.

src/datumaro/experimental/tiling/tilers.py
```diff
@@ -0,0 +1,422 @@
...
+                source_labels = df[source_sample_idx, column_name]
+                # Create list of True values matching label list length
+                keeps.append([True] * len(source_labels))
+                labels.append(source_labels)
+
+            # Return both the original labels and keep flags
+            return pl.DataFrame({column_name: labels, "keep": keeps})
+
+        # For non-list fields, just pass through the data
+        return pl.DataFrame({column_name: df[column_name].take(tiles_df["source_sample_idx"])})
```
AlbertvanHouten: Why is this implementation for is_list LabelField different? Wouldn't the labels apply to the whole image either way?

tests/integration/experimental/test_tiling_integration.py
```diff
@@ -0,0 +1,87 @@
+from typing import Any
+
+import numpy as np
+import polars as pl
+import pytest
+
+import datumaro.experimental.tiling.tilers  # Ensure tilers are registered
```
AlbertvanHouten: Should we include all things that need to be registered in experimental.__init__.py  to ensure they are always registered when anything from the experimental module is imported?

tests/unit/experimental/test_tiling.py
```diff
@@ -0,0 +1,362 @@
...
+def test_tiling_config():
+    config = TilingConfig(tile_width=50, tile_height=50)
+    assert config.tile_width == 50
+    assert config.tile_height == 50
+    assert config.overlap_x == 0
+    assert config.overlap_y == 0
+
+    config = TilingConfig(tile_width=50, tile_height=50, overlap_x=10, overlap_y=10)
+    assert config.overlap_x == 10
+    assert config.overlap_y == 10
```
AlbertvanHouten: I don't think this kind of test adds any value.

src/datumaro/experimental/tiling/tiler_registry.py
```diff
@@ -0,0 +1,654 @@
...
+                    source_sample_idx=idx,
+                    x=x,
+                    y=y,
+                    width=tile_width,
+                    height=tile_height,
+                )
+
+                # Convert tile info to polars using the field
+                tile_data = field.to_polars(tile_info_spec.name, tile_info)
+                tiles.append(tile_data)
```
AlbertvanHouten: I haven't looked too much into different tiling strategies, but would it not make more sense to ensure the tile sizes are consistent by adjusting the overlap for border tiles? That way, a tile size could be set to exactly the model input and no further image modifications would be necessary when feeding them into the model.

src/datumaro/experimental/tiling/tilers.py
gdlg: When is_list is False, the label correspond to a global classification label. When is_list is True, it is a list of labels typically attached associated to another annotation like a bounding box or a polygon.

If the tiling algorithm decides to remove one of the bounding box or polygon (e.g. if it’s outside the tile), then the corresponding label must be removed. In contrast, a global classification label always stay the same regardless of whether other annotations like a polygon are filtered out. So that’s why we have two different behaviours.

tests/integration/experimental/test_tiling_integration.py
gdlg: Yes, I think that this might be easier. Initially, I was thinking that we could allow the user to decide whether to register or not but it’s quite confusing.

tests/unit/experimental/test_tiling.py
gdlg: I agree. I’ll remove it.

src/datumaro/experimental/tiling/tiler_registry.py
gdlg: Fixed, I have created a named tuple.

src/datumaro/experimental/tiling/tilers.py
gdlg: I have added a check. I am wondering if we should remove the format attribute entirely until we properly support it.

AlbertvanHouten: I have checked with Geti tune, and they are representing a bbox as x/y/width/height so we should keep it. I will create a task to use an enum for these formats and add the converters. 

Would the tiling implementation know to first use the converter to x1/y1/x2/y2 before calling the logic here in case the bbox is in another format?

gdlg: > Would the tiling implementation know to first use the converter to x1/y1/x2/y2 before calling the logic here in case the bbox is in another format?

We need to think of a way to allow transforms/converters/etc to easily handle multiple formats without introducing too much complexity. It would also be useful to have a way to automatically test the transform against multiple formats.

AlbertvanHouten (approved):

PR #1908, opened 2025-10-01
Enable Renovate

AlexanderBarabanov:
This PR adds a workflows (`renovate.yml` and `renovate-config-validator.yml`) and configuration renovate.json5 to run Renovate as a GitHub Action.
Workflow `renovate.yml` will run daily and open PRs with suggested security upgrades and version upgrades (weekly and montly). 
Instead of using a Personal Access Token (PAT) that is tied to a particular user / faceless account this workflow uses private GitHub App with tuned permissions.

`renovate-config-validator.yml` workflow will validate changes in renovate.json5.

PR from Dependabot will be disabled.

Expected behavior:
PRs: https://github.com/AlexanderBarabanov/datumaro/pulls
Dashboard: https://github.com/AlexanderBarabanov/datumaro/issues/10

### Checklist

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -53,6 +53,22 @@
       matchPackagePatterns: ["*"],
       schedule: ["* * 1 * *"], // every month
     },
+
+    // uv version used in GitHub Actions is updated manually
+    {
+      enabled: false,
+      matchDatasources: ["github-releases"],
+      matchDepNames: ["astral-sh/uv"],
+      matchDepTypes: ["uses-with"],
+    },
+
+    // python version used in GitHub Actions is updated manually
+    {
+      enabled: false,
+      matchDatasources: ["github-releases"],
+      matchDepNames: ["python"],
+      matchDepTypes: ["uses-with"],
+    },
   ],
 
   // Enable security upgrades
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -50,10 +50,22 @@
       matchManagers: ["pip_requirements"],
     },
 
-    // Disable non-security upgrades for cargo
+    // Group Go version upgrades
+    {
+      enabled: true,
+      matchPackageNames: ["golang", "go"],
+      matchDatasource: ["crate"],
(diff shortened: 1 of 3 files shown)


gdlg (dismissed):
LGTM, thanks Alexander!

.github/workflows/renovate.yml
```diff
@@ -0,0 +1,84 @@
...
+#        default: false
+#        type: boolean
+#
+# Note: Renovate maintains and updates Dependency dashboard that is available in GitHub issues.
+
+name: Renovate
+on:
+  schedule:
+    # daily
+    - cron: "0 2 * * *"
```
AlbertvanHouten, line 42: Do we need to run and open PRs every day? Could it be on a weekly schedule, possibly?

.github/renovate.json5
```diff
@@ -0,0 +1,102 @@
...
+      matchManagers: ["pep621"],
+      schedule: ["* * * * 0"], // weekly
+    },
+
+    // numpy is updated manually
+    {
+      enabled: false,
+      matchDatasources: ["pypi"],
+      matchDepNames: ["numpy"],
+    },
```
AlbertvanHouten: Why is numpy updated manually?

AlbertvanHouten:
Side note, I have disabled the merge queue again. I thought I could enable it already even if there were no checks listed in it but turns out that does not work. Once merge queue checks are added, I will re-enable it.

.github/renovate.json5
AlexanderBarabanov: If enable numpy updates, Renovate will open PR like this https://github.com/AlexanderBarabanov/datumaro/pull/20/files
image

@AlbertvanHouten I thought that we restricted version in toml for some purpose, e.g. to avoid some breaking changes - let me know if it is safe to remove this rule.

AlbertvanHouten: This should be fine, if it breaks datumaro we will have to set a stricter range.

AlbertvanHouten (approved):

PR #1909, opened 2025-10-02
Multi label support

AlbertvanHouten:
This pull request introduces improvements to multi-label classification support in the experimental Datumaro codebase, specifically enhancing how multi-label fields are handled during conversion from legacy datasets and in the Polars serialization logic. The changes ensure that single labels are correctly treated as lists for multi-label fields and that multi-label configuration is properly propagated during legacy dataset conversion.

### Checklist

### DIFF
--- a/src/datumaro/experimental/fields.py
+++ b/src/datumaro/experimental/fields.py
@@ -397,7 +397,10 @@ class LabelField(Field):
             return {name: pl.Series(name, [None], dtype=pl_type)}
 
         if self.multi_label:
-            return {name: pl.Series(name, [to_numpy(value)], dtype=pl.List(self.dtype))}
+            # Handle single label as a list for multi-label fields
+            if not isinstance(value, list):
+                value = [value]
+            return {name: pl.Series(name, [value], dtype=pl_type)}
 
         return {name: pl.Series(name, [value], dtype=pl_type)}
 
--- a/src/datumaro/experimental/fields.py
+++ b/src/datumaro/experimental/fields.py
@@ -426,9 +426,6 @@ class LabelField(Field):
             return {name: pl.Series(name, [None], dtype=pl_type)}
 
         if self.multi_label:
-            # Handle single label as a list for multi-label fields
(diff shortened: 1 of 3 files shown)


src/datumaro/experimental/legacy.py
```diff
@@ -640,8 +640,10 @@ def convert_annotations(
         labels = [ann for ann in annotations if isinstance(ann, Label)]
         result = {}
         if len(labels) > 0:
-            # For single label classification, take the first label
-            result[self.name_prefix + "label"] = labels[0].label
+            if "multi_label_ids" in labels[0].attributes:
```
gdlg, line 643: Isn’t it possible to use this approach to detect whether the dataset is multi-label instead of convert_from_legacy’s multi-label parameter?

src/datumaro/experimental/fields.py
```diff
@@ -426,7 +426,10 @@ def to_polars(self, name: str, value: Any) -> dict[str, pl.Series]:
             return {name: pl.Series(name, [None], dtype=pl_type)}
 
         if self.multi_label:
-            return {name: pl.Series(name, [to_numpy(value)], dtype=pl.List(self.dtype))}
+            # Handle single label as a list for multi-label fields
+            if not isinstance(value, list):
+                value = [value]
```
gdlg: Here, this code assume that multi-label fields are represented as list, but we also want to support other arrays like Numpy and Torch. What’s the issue with the original code?

src/datumaro/experimental/legacy.py
AlbertvanHouten: This field is not present in every annotation, only if the annotation is actually multi label. I think this is the most reliable approach for that reason.

src/datumaro/experimental/fields.py
AlbertvanHouten: Good point. This is an issue coming from the legacy dataset, so I have addressed it in the legacy converter now.

PR #1925, opened 2025-10-14
Multi-label and hierarchical-label classification support 

AlbertvanHouten:
This pull request introduces comprehensive support for hierarchical label categories in the experimental categories module. It adds new data structures to represent hierarchical relationships between labels, updates legacy conversion logic to handle hierarchical and multi-label classification, and improves validation and compatibility. The changes also make several dataclasses mutable and update type imports for broader compatibility.

**Hierarchical label support and data structures:**

* Introduced new classes: `HierarchicalLabelCategory`, `LabelGroup`, and `HierarchicalLabelCategories` in `categories.py` to represent hierarchical label structures, label groups, and provide methods for hierarchy traversal and validation. These classes support parent-child relationships, groupings, and compatibility with existing interfaces.
* Added extensive validation in `HierarchicalLabelCategories.__post_init__` to ensure label uniqueness, group-label consistency, and valid parent references.
* Provided utility methods for hierarchy navigation (e.g., `find`, `get_children`, `get_parent`, `get_hierarchy_level`) and compatibility with legacy APIs.

**Legacy dataset conversion and analysis improvements:**

* Enhanced `analyze_legacy_dataset` and `convert_from_legacy` in `legacy.py` to detect hierarchical and multi-label projects, convert legacy label/group structures to the new classes, and ensure hierarchical labels are handled as lists. [[1]](diffhunk://#diff-aa35f06eaa7d35ff5ffa7077e181ed0b773549c22d42a92e78e076947f9b88f5L776-R817) [[2]](diffhunk://#diff-aa35f06eaa7d35ff5ffa7077e181ed0b773549c22d42a92e78e076947f9b88f5L803-R900) [[3]](diffhunk://#diff-aa35f06eaa7d35ff5ffa7077e181ed0b773549c22d42a92e78e076947f9b88f5L862-R949)
* Added helper functions `_attributes_to_dict` and `_has_derived_labels` for legacy attribute parsing and hierarchy detection.

**API and compatibility changes:**

* Added the `RESTRICTED` value to the `GroupType` enum to support empty label groups.
* Broadened type imports in `categories.py` for improved type hinting and compatibility.

**Other codebase updates:**

* Updated test imports in `test_categories.py` to include new hierarchical classes.
* Simplified Polars conversion logic in `fields.py` for label serialization.

These changes lay the groundwork for robust hierarchical and multi-label classification support in the experimental Datumaro API.

### Checklist


src/datumaro/experimental/categories.py
```diff
@@ -45,7 +46,7 @@ def from_str(cls, text: str) -> GroupType:
             raise ValueError(f"Invalid GroupType: {text}")
 
 
-@dataclass(frozen=True)
+@dataclass()
```
JortBergfeld: Are you sure this change is in line with the design? I think in general it is saver to keep the data immutable and make a new object explicitly whenever we need to change the data.

src/datumaro/experimental/categories.py
```diff
@@ -56,9 +57,11 @@ class Categories:
     pass
 
 
-@dataclass(frozen=True)
+@dataclass()
 class LabelCategories(Categories):
-    """Represents a group of labels with a specific group type and semantics."""
+    """Represents a group of labels with a specific group type and semantics.
```
JortBergfeld: Other docstrings seem to have the """ on a different line. Also a white line between the summary and additional info is common.

src/datumaro/experimental/categories.py
```diff
@@ -56,9 +57,11 @@ class Categories:
...
+    Use this for simple, non-hierarchical classification tasks.
```
JortBergfeld: I think I have seen this class being used in other type of projects, like detection and segmentation.

src/datumaro/experimental/categories.py
```diff
@@ -131,6 +134,231 @@ def __hash__(self):
...
+class HierarchicalLabelCategory:
+    """Represents a single label category with hierarchical support."""
+
+    name: str
+    parent: str = field(default="")
+    label_semantics: dict = field(default_factory=dict)
+
+    def __post_init__(self):
+        """Validate that name is not empty."""
+        if not self.name:
```
JortBergfeld: If we are going to validate, maybe it is best to also add `isinstance(name, str)` to avoid names like 1, etc.

src/datumaro/experimental/categories.py
```diff
@@ -131,6 +134,231 @@ def __hash__(self):
...
+
+    name: str
+    labels: Tuple[str, ...] = field(default_factory=tuple)
+    group_type: GroupType = GroupType.EXCLUSIVE
+
+    def __post_init__(self):
+        """Validate that name is not empty and labels is a tuple."""
+        if not self.name:
+            raise ValueError("Label group name cannot be empty")
+        if isinstance(self.labels, list):
```
JortBergfeld: I think it is cleaner to not do this, and just fix the calling code whenever an error is thrown. If you keep it, we might as well turn any iterable to a list, like sets, for example.

src/datumaro/experimental/categories.py
```diff
@@ -131,6 +134,231 @@ def __hash__(self):
...
+            object.__setattr__(self, "labels", tuple(self.labels))
+        elif not isinstance(self.labels, tuple):
+            raise TypeError("labels must be a tuple of strings")
+
+
+@dataclass()
+class HierarchicalLabelCategories(Categories):
+    """Represents hierarchical label categories with groups and parent-child relationships.
```
JortBergfeld: Same comment, it seems we use a different docstring format in other places.

src/datumaro/experimental/categories.py
```diff
@@ -131,6 +134,231 @@ def __hash__(self):
...
+        """Validate that there are no duplicate labels and build indices."""
+        # Convert lists to tuples if needed
+        if isinstance(self.items, list):
+            object.__setattr__(self, "items", tuple(self.items))
+        if isinstance(self.label_groups, list):
+            object.__setattr__(self, "label_groups", tuple(self.label_groups))
+
+        # Force each group's labels to tuple in case construction happened before class reload
+        for group in self.label_groups:
+            if isinstance(group.labels, list):  # resilience against stale instances
```
JortBergfeld: I think this is where immutable would help us to avoid bugs. This should already be validated in the LabelGroup creation, and after that it is immutable.

src/datumaro/experimental/categories.py
```diff
@@ -131,6 +134,231 @@ def __hash__(self):
...
+    """
+
+    items: Tuple[HierarchicalLabelCategory, ...] = field(default_factory=tuple)
+    label_groups: Tuple[LabelGroup, ...] = field(default_factory=tuple)
+    label_semantics: dict = field(default_factory=dict)
+
+    def __post_init__(self):
+        """Validate that there are no duplicate labels and build indices."""
+        # Convert lists to tuples if needed
+        if isinstance(self.items, list):
```
JortBergfeld: Again, I think throwing errors is better, fix the calling code.

src/datumaro/experimental/categories.py
```diff
@@ -131,6 +134,231 @@ def __hash__(self):
...
+        if isinstance(self.label_groups, list):
+            object.__setattr__(self, "label_groups", tuple(self.label_groups))
+
+        # Force each group's labels to tuple in case construction happened before class reload
+        for group in self.label_groups:
+            if isinstance(group.labels, list):  # resilience against stale instances
+                object.__setattr__(group, "labels", tuple(group.labels))
+
+        # Validate no duplicate names
+        seen_names = set()
```
JortBergfeld, line 188: Is it not better to use label comprehension followed by `set` and comparing the two `len`?

src/datumaro/experimental/categories.py
```diff
@@ -131,6 +134,231 @@ def __hash__(self):
...
+        # Validate that all labels in groups exist
+        for group in self.label_groups:
+            for label_name in group.labels:
+                if label_name not in seen_names:
+                    raise ValueError(
+                        f"Label '{label_name}' in group '{group.name}' not found in items"
+                    )
+
+    @property
+    @cache
```
JortBergfeld, line 215: If the class remains mutable, does this cache properly work?

src/datumaro/experimental/categories.py
```diff
@@ -131,6 +134,231 @@ def __hash__(self):
...
+    def _children_map(self) -> Dict[str, Tuple[str, ...]]:
+        """Cached mapping from parent names to child names."""
+        children_map: Dict[str, List[str]] = {}
+        for item in self.items:
+            if item.parent:
+                if item.parent not in children_map:
+                    children_map[item.parent] = []
```
JortBergfeld: You can use a defaultdict, that automatically creates an empty list, if the key doesn't exist yet.

src/datumaro/experimental/categories.py
```diff
@@ -131,6 +134,231 @@ def __hash__(self):
...
+            ]
+        ],
+    ) -> "HierarchicalLabelCategories":
+        """
+        Creates a HierarchicalLabelCategories from iterable.
+
+        Args:
+            iterable: This iterable object can be:
+                - a list of str - will be interpreted as list of Category names
+                - a list of positional arguments - will generate Categories with these arguments
```
JortBergfeld: I think an example will help to understand this method.

JortBergfeld (commented):
Looks good, but some comments.

src/datumaro/experimental/categories.py
AlbertvanHouten: Thanks, no it is not. I removed it during testing and forgot to add it back.

src/datumaro/experimental/categories.py
AlbertvanHouten: Currently, these are not always equal. The old datumaro is sometimes storing label ids as its name, and then the actual name is stored as an attribute. We can use a set once the old datumaro logic has been removed completely.

src/datumaro/experimental/categories.py
AlbertvanHouten: Yeah, this is a bit inconsistent throughout the code base at the moment. We need to introduce a linting rule to enforce uniform documentation.

src/datumaro/experimental/categories.py
AlbertvanHouten: I've removed this method, it shouldn't really be used and was copied over from the old datumaro logic.

src/datumaro/experimental/categories.py
```diff
@@ -131,6 +135,191 @@ def __hash__(self):
...
+    @cache
+    def _index_map(self) -> Dict[str, int]:
+        """Cached mapping from label names to indices."""
+        return {item.name: idx for idx, item in enumerate(self.items)}
+
+    @property
+    @cache
+    def _children_map(self) -> Dict[str, Tuple[str, ...]]:
+        """Cached mapping from parent names to child names."""
+        from collections import defaultdict
```
JortBergfeld: I don't think it will matter if you put this import at the top, I would say this is a bit of an overkill optimization.

JortBergfeld (dismissed):
Looks good, one more comment for your consideration.

JortBergfeld (approved):

PR #1933, opened 2025-10-28
Add filtering a dataset on subset

### LINKED ISSUE #1924: Add subset filtering
Add subset filtering for the Dataset class. This method should only work if the samples in the dataset have a [subset](https://github.com/open-edge-platform/datumaro/blob/develop/src/datumaro/experimental/fields.py#L1082) field present.

The interface is not strictly defined, but something along the lines of this would be probably be ideal:

```
class Dataset(Generic[DType]):
    def filter_by_subset(self, subset: Subset) -> Dataset[DType]:
        ....
```

JortBergfeld:
Filter a dataset on a subset, like training or validation.

Resolves #1924

### Checklist

### DIFF
--- a/src/datumaro/experimental/dataset.py
+++ b/src/datumaro/experimental/dataset.py
@@ -481,12 +481,10 @@ class Dataset(Generic[DType]):
         Returns:
             A new Dataset with items of the given subset.
         """
-        subset_column_name: str | None = None
-        for column_name, attribute_info in self.schema.attributes.items():
+        for subset_column_name, attribute_info in self.schema.attributes.items():
             if isinstance(attribute_info.annotation, SubsetField):
-                subset_column_name = column_name
                 break
-        if subset_column_name is None:
+        else:
             raise RuntimeError(
                 f"Dataset does not have an attribute for 'SubsetField': schema: {self.df.schema}"
             )
--- a/src/datumaro/experimental/dataset.py
+++ b/src/datumaro/experimental/dataset.py
@@ -24,6 +24,8 @@ from typing import (
 import polars as pl
 from typing_extensions import TypeGuard, TypeVar, dataclass_transform
 
+from datumaro.experimental.fields import Subset, SubsetField
+
 from .converter_registry import ConverterTransform, find_conversion_path
 from .schema import AttributeInfo, Field, Schema
 from .transform import IdentityTransform, Transform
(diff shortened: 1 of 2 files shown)


src/datumaro/experimental/dataset.py
```diff
@@ -469,6 +471,34 @@ def convert_to_schema(
...
+                f"Dataset does not have an attribute for 'SubsetField': schema: {self.df.schema}"
+            )
+
+        filtered_df = self.df.filter(self.df[subset_column_name] == subset.name)
+
+        return Dataset.from_dataframe(
+            df=filtered_df,
+            dtype_or_schema=self.dtype,
+            schema=self.schema,
+        )
```
leoll2, line 498: This implementation has the advantage of being fast and straightforward, but it's not optimal for memory because it allocates a new `Sample` object for every filtered item. Would it make sense to add a parameter `view: bool = False` which, if enabled, returns a `Dataset` constructed with the same `Sample` objects of the original dataset? I'm not familiar with the Datumaro internals, would that approach be even feasible (if samples are lazy-loaded, perhaps it's not)?

With an example:
```python
dataset = Dataset(SubsetSample)
dataset.append(SubsetSample(image=np.array([1]), subset=Subset.TRAINING))

dataset_view = dataset.filter_by_subset(Subset.TRAINING, view=True)

assert id(dataset_view[0]) == id(dataset[0])  # <- same Python object
```

src/datumaro/experimental/dataset.py
```diff
@@ -469,6 +471,34 @@ def convert_to_schema(
...
+        Returns:
+            A new Dataset with items of the given subset.
+        """
+        subset_column_name: str | None = None
+        for column_name, attribute_info in self.schema.attributes.items():
+            if isinstance(attribute_info.annotation, SubsetField):
+                subset_column_name = column_name
+                break
+        if subset_column_name is None:
+            raise RuntimeError(
```
```suggestion
        for column_name, attribute_info in self.schema.attributes.items():
            if isinstance(attribute_info.annotation, SubsetField):
                subset_column_name = column_name
                break
        else:
            raise RuntimeError(
```
I know that not everyone like `for...else`, feel free to ignore.

src/datumaro/experimental/dataset.py
AlbertvanHouten: This isn't a problem since the data frame will be queried to only returns rows with the correct subset. The sample objects are not pre-allocated in any dataset, but only created from the data frame when you make a call to the item.  Even on the same dataset  `id(dataset[0]) != id(dataset[0])`

AlbertvanHouten (approved):

PR #1946, opened 2025-11-04
Implement dataset export/import

AlbertvanHouten:
This pull request adds support to import and export datasets. An exported dataset will include all the images, where references to callables for images/masks are saved as PNG and image paths will be copied. The exported dataset also includes metadata, such as the categories and schema of the dataset and a version number in a JSON file. The polars dataframe is stored in parquet. Additionally, there is an option to export everything in a zip. This zip can also be imported as is without the need to manually extract it first. 

**Serialization and Deserialization Enhancements**
* Added `to_dict` and `from_dict` methods to `Categories`, `LabelCategories`, `HierarchicalLabelCategories`, and `MaskCategories`, enabling polymorphic serialization and reconstruction of category objects. [[1]](diffhunk://#diff-f0858b860d68536c04476766ef0512d873b7f7cc20dde377a42ec13a836d9622L58-R101) [[2]](diffhunk://#diff-f0858b860d68536c04476766ef0512d873b7f7cc20dde377a42ec13a836d9622R181-R223) [[3]](diffhunk://#diff-f0858b860d68536c04476766ef0512d873b7f7cc20dde377a42ec13a836d9622R407-R496) [[4]](diffhunk://#diff-f0858b860d68536c04476766ef0512d873b7f7cc20dde377a42ec13a836d9622R565-R601)
* Implemented `to_dict` and `from_dict` methods for the `Field` base class, using dataclass introspection to automatically serialize and reconstruct field attributes, including special handling for semantic and dtype fields.
* Added serialization and deserialization logic to the `Schema` class, including storing type information and reconstructing attribute types via module introspection.

**Dataset Import/Export API**
* Exposed `export_dataset` and `import_dataset` functions at the module level, and added `export` and `from_file` methods to the `Dataset` class for saving/loading datasets in a structured format. [[1]](diffhunk://#diff-6c66a059cb4075cd7afbc29b4f34236f8101a1a47cdcc429c7a2d25868cc74b5R8) [[2]](diffhunk://#diff-4ac196ddc4dc8e6d33daf684ded18886ff8774fadb8b6cbd4bfa88ca424bb34fR500-R554)

These changes lay the groundwork for robust dataset interchange and future compatibility across different category and schema types.

### Checklist

### DIFF
--- a/src/datumaro/experimental/categories.py
+++ b/src/datumaro/experimental/categories.py
@@ -55,7 +55,50 @@ class Categories:
     label attributes etc.
     """
 
-    pass
+    def to_dict(self) -> Dict[str, any]:
+        """
+        Serialize this Categories instance to a JSON-compatible dictionary.
+
+        This default implementation should be overridden by subclasses that have
+        specific serialization needs. Returns just the type by default.
+
+        Returns:
+            Dictionary representation of this Categories instance
+        """
+        return {"type": self.__class__.__name__}
+
+    @classmethod
+    def from_dict(cls, data: Dict[str, any]) -> "Categories":
+        """
+        Deserialize a Categories instance from a JSON dictionary.
+
+        This method uses polymorphic dispatch to create the correct subclass
+        based on the "type" field in the dictionary.
+
+        Args:
+            data: Dictionary containing serialized Categories data
+
+        Returns:
+            Reconstructed Categories instance of the appropriate subclass
+        """
+        cat_type = data.get("type")
+
+        if not cat_type:
+            raise ValueError("Categories dictionary must have a 'type' field")
+
+        # Import subclasses to make them available
+        subclass_map = {
+            "LabelCategories": LabelCategories,
+            "HierarchicalLabelCategories": HierarchicalLabelCategories,
+            "MaskCategories": MaskCategories,
+        }
+
+        if cat_type in subclass_map:
+            return subclass_map[cat_type].from_dict(data)
+        else:
+            # Unknown type - return base Categories with just the type info
+            # This allows forward compatibility with new category types
+            raise ValueError(f"Unknown categories type: {cat_type}")
 
 
 @dataclass(frozen=True)
@@ -134,6 +177,49 @@ class LabelCategories(Categories):
     def __hash__(self):
         # Include label_semantics in the hash
         return hash((self.labels, self.group_type, frozenset(self.label_semantics.items())))
+
+    def to_dict(self) -> Dict[str, any]:
+        """
+        Serialize to a JSON-compatible dictionary.
+
+        Returns:
+            Dictionary representation of this LabelCategories instance
+        """
+        return {
+            "type": "LabelCategories",
+            "labels": list(self.labels),
+            "group_type": self.group_type.name,
+            "label_semantics": {
+                k.name if isinstance(k, LabelSemantic) else str(k): v
+                for k, v in self.label_semantics.items()
+            },
+        }
+
+    @classmethod
+    def from_dict(cls, data: Dict[str, any]) -> "LabelCategories":
+        """
+        Deserialize from a JSON dictionary.
+
+        Args:
+            data: Dictionary containing serialized LabelCategories data
+
+        Returns:
+            Reconstructed LabelCategories instance
+        """
+        # Reconstruct label_semantics with proper LabelSemantic keys
+        label_semantics = {}
+        for k, v in data.get("label_semantics", {}).items():
+            try:
+                key = LabelSemantic[k]
+            except KeyError:
+                key = k
+            label_semantics[key] = v
+
+        return cls(
+            labels=tuple(data["labels"]),
+            group_type=GroupType[data["group_type"]],
+            label_semantics=label_semantics,
+        )
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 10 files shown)


src/datumaro/experimental/categories.py
```diff
@@ -55,7 +55,50 @@ class Categories:
     label attributes etc.
     """
 
-    pass
+    def to_dict(self) -> Dict[str, any]:
```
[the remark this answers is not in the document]
AlbertvanHouten: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/open-edge-platform/datumaro/pull/1946#discussion_r2493676079)

src/datumaro/experimental/schema.py
```diff
@@ -100,6 +104,97 @@ def __get__(self, instance, _):
...
+
+                # Handle semantic as special case (convert to string)
+                if field_name == "semantic" and isinstance(field_value, Semantic):
+                    field_dict[field_name] = field_value.name
+                # Handle Polars data types
+                elif field_name == "dtype" and isinstance(field_value, pl.DataType):
+                    field_dict[field_name] = str(field_value)
+                # Handle other types with __name__
+                elif field_name == "dtype" and hasattr(field_value, "__name__"):
+                    field_dict[field_name] = field_value.__name__
```
leoll2: What does `field_value.__name__` represent?

src/datumaro/experimental/export_import.py
```diff
@@ -0,0 +1,443 @@
...
+        raise ValueError(
+            "Exporting datasets with transforms is not yet supported. "
+            "Please materialize the dataset first by converting it."
+        )
+
+    # Create temporary directory if using zip, otherwise use output_path directly
+    temp_dir = None
+    if as_zip:
+        if output_path.suffix != ".zip":
+            output_path = output_path.with_suffix(".zip")
```
leoll2: Docstring says:

> output_path: Path to export to (directory or .zip file)

If output path is a directory, for example `foo/bar/`, then this line would output the zip at `foo/bar/.zip`. Basically it's missing the file basename, I think you should output to `foo/bar/dataset.zip` or similar.

src/datumaro/experimental/export_import.py
```diff
@@ -0,0 +1,443 @@
...
+    Raises:
+        ValueError: If the dataset has transforms (not yet supported)
+    """
+    output_path = Path(output_path)
+
+    # Check if dataset has transforms
+    if dataset._transforms is not None:
+        raise ValueError(
+            "Exporting datasets with transforms is not yet supported. "
+            "Please materialize the dataset first by converting it."
```
leoll2: Transforms are an implementation detail, correct if I'm wrong. I think a normal use wouldn't understand this error message. How do you materialize a dataset in practice?

src/datumaro/experimental/export_import.py
```diff
@@ -0,0 +1,443 @@
...
+        if export_images:
+            images_dir = work_dir / IMAGES_DIR
+            image_paths = _export_images_from_dataset(dataset, images_dir)
+
+        # Track which columns are Object types (excluded from parquet)
+        object_columns = [col for col, dtype in dataset.df.schema.items() if dtype == pl.Object]
+
+        # Create metadata
+        metadata = {
+            "version": VERSION,
```
leoll2, line 269: I suggest to use the version of the `datumaro` package here.

src/datumaro/experimental/export_import.py
```diff
@@ -0,0 +1,443 @@
...
+            "schema": dataset.schema.to_dict(),
+            "dtype": dataset.dtype.__name__ if dataset.dtype != Sample else None,
+            "image_paths": image_paths,
```
leoll2: The "image_paths" dictionary can be huge, I don't think it's a good idea to store it here. Assuming that all images are stored in the same folder, here you could just write the path to the base image folder. The name of the file (relative to the base image folder) can be stored in the dataframe.

src/datumaro/experimental/export_import.py
AlbertvanHouten: I have removed this check for now as it's not needed yet.

src/datumaro/experimental/export_import.py
AlbertvanHouten: Using `foo/bar/` produces zip `foo/bar.zip` which I think is fine.

src/datumaro/experimental/schema.py
AlbertvanHouten: Removed. This is necessary in case a field type is defined as e.g. `pl.Float32` instead of correct `pl.Float32()`. It is better to validate this during sample creation, though.

src/datumaro/experimental/dataset.py
```diff
@@ -507,6 +507,61 @@ def append_dataset(self, dataset: Dataset) -> None:
...
+
+        Args:
+            output_path: Path to export to (directory or .zip file)
+            export_images: Whether to export images from callable/path fields
+            as_zip: Whether to package everything as a ZIP file
+
+        Raises:
+            ValueError: If the dataset has transforms (not yet supported)
+        """
+        from .export_import import export_dataset
```
JortBergfeld: Does this import have to be in the function, not at the top? To be honest, I think we do not need to define class methods for export and import. These are huge operations and I believe it makes more sense to keep these functions separate from the Dataset class.

src/datumaro/experimental/export_import.py
JortBergfeld: I would disagree, changing a directory name to a filename is clearly unexpected behavior and we should see if we can either validate the output_path, or use a default name in case the output path has no file name. Also note that the doc string states directory or zip file, so possibly any name not ending in .zip should be treated as directory name.?Perhaps it is better to provide a separate parameter for output_directory (defaults to current working dir) and output_filename (defaults to datumaro_dataset.zip)?

src/datumaro/experimental/export_import.py
```diff
@@ -0,0 +1,432 @@
...
+        The imported Dataset instance
+
+    Raises:
+        FileNotFoundError: If required files are missing
+        ValueError: If the dataset format is invalid
+    """
+    input_path = Path(input_path)
+
+    # Handle ZIP files
+    if input_path.suffix == ".zip":
```
JortBergfeld: The simplest and most reliable way to check if a file is a valid ZIP file in Python is to use the is_zipfile() function from the built-in zipfile module.

This function checks the magic number (the identifying bytes) at the beginning of the file, not just the file extension, making it a robust check.

src/datumaro/experimental/export_import.py
```diff
@@ -0,0 +1,432 @@
...
+
+    finally:
+        # Clean up temporary directory if used
+        if as_zip and temp_dir is not None and temp_dir.exists():
+            shutil.rmtree(temp_dir)
+
+
+def import_dataset(
+    input_path: Union[str, Path],
+    dtype: Type["DType"] | None = None,
```
JortBergfeld, line 295: This seems weird to me, I would expect this to be part of the exported dataset. Can you provide an example in the docs explaining when and how this would be used?

src/datumaro/experimental/schema.py
```diff
@@ -188,5 +280,96 @@ def with_categories(self, categories: Dict[str, "Categories"]) -> "Schema":
...
+            # Attempt to reconstruct the actual type
+            if type_module and type_module != "builtins":
+                try:
+                    module = importlib.import_module(type_module)
+                    attr_type = getattr(module, type_name, object)
+                except (ImportError, AttributeError):
+                    attr_type = object
+            elif type_name in ["int", "float", "str", "bool", "list", "dict", "tuple"]:
+                # Handle built-in types
+                BUILTIN_TYPES = {
```
JortBergfeld: Not entirely sure if it matters, but normally I wouldn't declare a constant inside of a for loop, as it suggests the constant is recreated dynamically within the loop based on the input.

src/datumaro/experimental/schema.py
```diff
@@ -188,5 +280,96 @@ def with_categories(self, categories: Dict[str, "Categories"]) -> "Schema":
...
+            type_module = attr_dict.get("type_module")
+
+            # Attempt to reconstruct the actual type
+            if type_module and type_module != "builtins":
+                try:
+                    module = importlib.import_module(type_module)
+                    attr_type = getattr(module, type_name, object)
+                except (ImportError, AttributeError):
+                    attr_type = object
+            elif type_name in ["int", "float", "str", "bool", "list", "dict", "tuple"]:
```
JortBergfeld: You can use `type_name in BUILTIN_TYPES` if you put the constant declaration out of the loop

tests/integration/experimental/test_export_import.py
JortBergfeld, line 1: I believe for import export it would be nice to have a full integration tests that exports and reimports a dataset to make sure the full process works as expected, and most importantly that an exported dataset can be imported as well.

JortBergfeld (commented):
Looks good, but a few comments for improvements.

src/datumaro/experimental/export_import.py
AlbertvanHouten: This is included in the exported dataset and the user does not need to provide it upon import. However, for auto-completion/typing reasons it will be beneficial to provide this in certain cases.

tests/integration/experimental/test_export_import.py
AlbertvanHouten: There are several tests in this file that export and reimport a dataset. Especially the last test in this file is checking if every field can be imported/exported. What specifically is not yet covered?

src/datumaro/experimental/dataset.py
AlbertvanHouten: There are import loops if put at the top. I agree with you that this doesn't necessarily need to be part of the class. I will move it out of here.

src/datumaro/experimental/export_import.py
AlbertvanHouten: Fair enough, I will append a standard name if zipping and the user does not provide a .zip extension output_path

tests/integration/experimental/test_export_import.py
JortBergfeld: I would probably call tests that do both an export and import an integration test rather than a unit test, so perhaps these should be moved. I don't think anything drastic is missing, looking at the codecoverage it seems mostly exceptions (but do have a look, you are missing some to_dict and from_dicts). I was thinking more about a test that does a more generic assertion (`dataset_to_export == imported_dataset`) that would potentially also catch future changes to the export and import mechanism.

src/datumaro/experimental/export_import.py
JortBergfeld: I would probably update the docstring to emphasize that this is an optional parameter and some example(s) why one would use it.

tests/integration/experimental/test_export_import.py
AlbertvanHouten: Sure.. Moved the tests to integration folder.

JortBergfeld (approved):

Issue #1937, opened 2025-10-28
Reorganize repo file structure


AlbertvanHouten:
Reorganize the layout of the experimental folder by introducing smaller files for each specific type. For example, fields.py has grown >1000 lines, split each field into a separate file. Ensure the original import structure remains valid. Also consider test files.

PR #1956, opened 2025-11-13
Reorganize files in experimental folder

JortBergfeld:
This pull request refactors and modularizes the experimental dataset system in Datumaro, improving code organization and clarity. The changes split field and converter implementations into dedicated submodules, update import paths throughout the codebase, and enhance documentation and naming consistency to reflect the transition to the v2 dataset system.

**Refactoring and modularization of fields and converters:**

* Introduced new submodules for fields (`fields/annotations.py`, `fields/images.py`, `fields/masks.py`, `fields/datasets.py`) and converters (`converters/annotation_converters.py`, `converters/image_converters.py`, `converters/mask_converters.py`, etc.), and updated `src/datumaro/experimental/fields/__init__.py` and `src/datumaro/experimental/converters/__init__.py` to re-export their contents for better organization and discoverability. [[1]](diffhunk://#diff-29b9ccd281e40c6b1d4f8b918075d0a3f7ba4c4e7d464d6d6f6fb7f917c80e27R1-R99) [[2]](diffhunk://#diff-000e98833d5fba9d66f82db2422b240bd02e0fde9f86af6f7ffd9a431abd5b01R1-R85)

* Added a comprehensive and extensible `Converter` base class in `src/datumaro/experimental/converters/base.py`, including methods for type extraction, conversion logic, and attribute remapping, plus the `ConversionError` exception and `AttributeRemapperConverter` implementation.

**Updates to import paths and usage across the codebase:**

* Updated imports in `src/datumaro/experimental/__init__.py`, `src/datumaro/experimental/dataset.py`, and `src/datumaro/experimental/export_import.py` to use the new submodules and registry structure, replacing direct imports from previous monolithic modules. [[1]](diffhunk://#diff-6c66a059cb4075cd7afbc29b4f34236f8101a1a47cdcc429c7a2d25868cc74b5L5-R9) [[2]](diffhunk://#diff-4ac196ddc4dc8e6d33daf684ded18886ff8774fadb8b6cbd4bfa88ca424bb34fL14-R15) [[3]](diffhunk://#diff-12fc83a072e8732e4db0eea5c510e8299e4d581901074c71cd5628f434d9ba05L31-R34)

**Documentation and naming improvements:**

* Updated module-level docstrings in `categories.py` and `export_import.py` to clarify that they pertain to the v2 dataset system, improving clarity for future maintainers. [[1]](diffhunk://#diff-f0858b860d68536c04476766ef0512d873b7f7cc20dde377a42ec13a836d9622L6-R6) [[2]](diffhunk://#diff-12fc83a072e8732e4db0eea5c510e8299e4d581901074c71cd5628f434d9ba05L6-R6)

**(References: [[1]](diffhunk://#diff-29b9ccd281e40c6b1d4f8b918075d0a3f7ba4c4e7d464d6d6f6fb7f917c80e27R1-R99) [[2]](diffhunk://#diff-000e98833d5fba9d66f82db2422b240bd02e0fde9f86af6f7ffd9a431abd5b01R1-R85) [[3]](diffhunk://#diff-f857cdf6810a5e27dbaa36323bc181430a0ec4e8e5214e03c549163c374c0ac2R1-R297) [[4]](diffhunk://#diff-6c66a059cb4075cd7afbc29b4f34236f8101a1a47cdcc429c7a2d25868cc74b5L5-R9) [[5]](diffhunk://#diff-4ac196ddc4dc8e6d33daf684ded18886ff8774fadb8b6cbd4bfa88ca424bb34fL14-R15) [[6]](diffhunk://#diff-12fc83a072e8732e4db0eea5c510e8299e4d581901074c71cd5628f434d9ba05L31-R34) [[7]](diffhunk://#diff-f0858b860d68536c04476766ef0512d873b7f7cc20dde377a42ec13a836d9622L6-R6) [[8]](diffhunk://#diff-12fc83a072e8732e4db0eea5c510e8299e4d581901074c71cd5628f434d9ba05L6-R6)**

### Checklist


src/datumaro/v2/converters/__init__.py
```diff
@@ -0,0 +1,85 @@
+# Copyright (C) 2025 Intel Corporation
+#
+# SPDX-License-Identifier: MIT
+
+from datumaro.v2.converters.annotation_converters import (
```
leoll2: In the `__init__.py` files it is better to use relative imports (`from .annotation_converters import ...`), so that you don't have to update them if the module gets repositioned somewhere else in the repo.

src/datumaro/experimental/__init__.py
leoll2, line 1: Exporting **all** the symbols at the package root level may be a bit of hassle to maintain. Consider these alternatives:

- Use `*` imports, like `from .fields.annotations import *`. Using `*` is usually discouraged, but it can be handy in circumstances like this.
- Keep the symbols organized and accessible through high-level submodules, instead of exporting everything at the package root. For example, an import like `from datumaro.fields import BBoxField` would be perfectly acceptable from user perspective (provided that the `fields` module remains stable and doesn't get renamed), you don't have to also expose `from datumaro import BBoxField`. This is very common in large packages, not everything is made available at the package root.

src/datumaro/v2/converters/__init__.py
AlbertvanHouten: I disagree here, moving a module and updating relevant imports would be handled automatically in any modern IDE, so I don't think we need to consider that. Absolute imports would be better for readability and for consistency across the repo. It also means we don't need to configure a separate ruleset in linters for `__init__.py` files to skip checking these imports.

src/datumaro/components/dataset.py
```diff
@@ -65,7 +65,7 @@ def __iter__(self):
         yield from self.parent._data.get_subset(self.name)
 
     def __len__(self):
-        subset: DatasetItemStorageDatasetView.Subset = self.parent._data.get_subset(self.name)
+        subset: datumaro.v2.fields.datasets.Subset = self.parent._data.get_subset(self.name)
```
AlbertvanHouten: Changes in this file should be reverted

src/datumaro/v2/converters/__init__.py
leoll2: > Absolute imports would be better for readability

Absolute imports are longer, I don't see how that improves readability

> It also means we don't need to configure a separate ruleset in linters for __init__.py files to skip checking these imports.

Can you elaborate on this? What problems do linters have with relative imports?

leoll2:
> This pull request refactors and reorganizes the experimental dataset API into a new datumaro.v2 package

As discussed offline, I recommend to do the other way around: move the legacy classes to `datumaro.legacy` and the new ones to `datumaro`. Next target release is 2.0, where the old API can be broken in favor of the new one. From 2.0 onwards, the new API should remain relatively stable, therefore it's better to have the classes already in `datumaro` rather than a `datumaro.v2` folder that must be eventually removed. This approach also makes it very clear what classes are legacy; these can be gradually removed in later releases, as the equivalent features get reimplemented in the new API.

src/datumaro/v2/converters/__init__.py
AlbertvanHouten: Maybe readable is the wrong word, but absolute imports are clearer for sure. Importing from just `base` vs importing from `datumaro.converters.base` gives useful context. In general, it is recommended to not mix different import strategies in a repository. 

Regarding the linters, we are enforcing all imports to be absolute. We would have to add an exception to this rule for the `__init__.py` files (or add ignores everwyhere).

src/datumaro/experimental/__init__.py
JortBergfeld: I have chosen to reduce the number of imports, as importing using * might import/expose too much to the end user. I feel it is better to keep some control over what we expose.

src/datumaro/v2/converters/__init__.py
JortBergfeld: I will keep the absolute imports, as our goal is to have one set of rules, if possible all default rules of ruff, which includes absolute imports over relative ones.

JortBergfeld:
> > This pull request refactors and reorganizes the experimental dataset API into a new datumaro.v2 package
> 
> As discussed offline, I recommend to do the other way around: move the legacy classes to `datumaro.legacy` and the new ones to `datumaro`. Next target release is 2.0, where the old API can be broken in favor of the new one. From 2.0 onwards, the new API should remain relatively stable, therefore it's better to have the classes already in `datumaro` rather than a `datumaro.v2` folder that must be eventually removed. This approach also makes it very clear what classes are legacy; these can be gradually removed in later releases, as the equivalent features get reimplemented in the new API.

I have renamed the v2 directory back to experimental and we will follow your suggestion in a later PR.

src/datumaro/v2/converters/__init__.py
leoll2: Sounds good. I checked the code of other large, open-source Python projects to figure out what approach they follow, absolute vs relative imports. Turns out that it's more or less 50-50, there is no consensus - both approaches are fine.

src/datumaro/experimental/categories.py
```diff
@@ -3,7 +3,7 @@
 # SPDX-License-Identifier: MIT
 
 """
-Categories definitions for the experimental dataset system.
+Categories definitions for the v2 dataset system.
```
leoll2: to be reverted

tests/unit/experimental/test_type_registry.py
```diff
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: MIT
 
-"""Tests for the experimental type registry system."""
+"""Tests for the v2 type registry system."""
```
leoll2:
```suggestion
"""Tests for the type registry system."""
```

tests/unit/experimental/test_schema.py
```diff
@@ -8,6 +8,7 @@
 import polars as pl
 import pytest
 
+import datumaro.experimental.fields.base
```
leoll2: Unnecessary import?

tests/unit/experimental/test_converters.py
leoll2, line 1: In this or another PR, please also restructure the tests to match the layout of the source files

tests/unit/experimental/test_schema.py
AlbertvanHouten: @JortBergfeld could you also check why this isn't caught/fixed by ruff?

src/datumaro/experimental/categories.py
JortBergfeld: I removed v2 in most places, since it will be the default dataset format soon.

tests/unit/experimental/test_schema.py
JortBergfeld: I ignored all "F4" problems in the tests folder, since there are some tests relying on * imports or unused imports to get all formats/converters/etc and then loop over them, or to get fixtures or to test availability. I have now enabled F401 (unused imports) and fixed or ignored inline the errors, but I still disabled F403 and F405 (star imports and undefined class/object) to not have to rewrite some tests.

tests/unit/experimental/test_converters.py
JortBergfeld: The tests folder restructure will be done separately.

AlbertvanHouten (approved):

tests/unit/experimental/test_legacy.py
```diff
@@ -348,7 +347,7 @@ def test_convert_from_legacy_with_image_bytes():
     # Create legacy dataset
     legacy_dataset = LegacyDataset.from_iterable(items)
 
-    # Convert to experimental format
+    # Convert to v2 format
```
leoll2, line 350: This file still contains several "v2" references

tests/unit/experimental/test_legacy_integration.py
```diff
@@ -64,7 +64,7 @@ def test_object_detection(
         # Step 2: Import it back as a legacy dataset
         legacy_dataset = LegacyDataset.import_from(src_dir, format=input_format)  # pyright: ignore[reportUnknownMemberType]
 
-        # Step 3: Convert to experimental format
+        # Step 3: Convert to v2 format
```
leoll2, line 67: "v2" leftover also here

tests/unit/experimental/test_legacy.py
JortBergfeld: I believe in legacy conversion it makes sense to speak about legacy format and v2 format, since experimental format suggests the format should not be used. When we release 2.0 I assume the format is stable and usable in production.

leoll2:
Contributes to #1937

PR #1963, opened 2025-11-21
Coco Dataset support

### LINKED ISSUE #1964: Implement conversion to/from COCO format
COCO is a very popular format for annotating computer vision datasets, primarily for tasks like object detection and segmentation. It provides a structured way to store image and annotation information, including categories, bounding boxes, and segmentation masks, within a single JSON file.

While revamping the Datumaro core classes, we want to keep the possibility for users to convert their datasets to/from COCO format, as well as loading/storing COCO datasets from/to disk.

AlbertvanHouten:
This pull request introduces comprehensive support for the COCO dataset format in the `datumaro.experimental` module, including new data structures, IO utilities, and field types. The changes add the ability to load and save datasets in COCO format, define a schema for COCO samples, and extend the field system to handle captions, numerics and booleans. The numeric and boolean fields are generalized fields which can be used for multiple purposes. For example, numeric can be used to present a score for a prediction or as an identifier for images. Additionally, the enum for semantics has been replaced with a string to allow more flexibility. ScoreField has been removed, as NumericField can now be used to represent these values. 

**COCO Datasets** 
Logic to load and save coco datasets was added in a new folder: "experimental/data_formats/coco". There are 2 main files:

**io.py** Contains logic to both load and save COCO datasets, including those from different version years.
**sample.py** Defines a COCO sample that is able to hold all annotation data, including polygons, bboxes, keypoints and captions. 
 

**Performance**
Tiny coco dataset with 100 train and 100 val images loads in ~0.17 seconds. (https://github.com/lizhogn/tiny_coco_dataset)

### Checklist


src/datumaro/experimental/fields/base.py
```diff
@@ -26,13 +26,14 @@ class Semantic(Flag):
     """
     Used for disambiguation when multiple fields of the same type exist.
     Default is used for fields that don't need disambiguation.
-    Left/Right are used for stereo vision scenarios.
     """
 
     Default = auto()
-    Left = auto()
-    Right = auto()
```
leoll2, line 32: Why remove these two? They can be helpful to model samples acquired from stereo cameras.

tests/unit/experimental/test_dataset.py
```diff
@@ -252,11 +252,11 @@ def test_stereo_sample_with_semantics():
     """Test dataset with stereo samples using semantic tags."""
 
     class StereoSample(Sample):
-        left_image: np.ndarray[Any, Any] = image_field(dtype=pl.UInt8, format="RGB", semantic=Semantic.Left)
-        right_image: np.ndarray[Any, Any] = image_field(dtype=pl.UInt8, format="BGR", semantic=Semantic.Right)
+        left_image: np.ndarray[Any, Any] = image_field(dtype=pl.UInt8, format="RGB", semantic=Semantic.Bbox)
+        right_image: np.ndarray[Any, Any] = image_field(dtype=pl.UInt8, format="BGR", semantic=Semantic.Polygon)
```
leoll2: This change doesn't make sense - semantic.Bbox/Polygon does not relate to stereo data, while Left/Right does.

src/datumaro/experimental/fields/base.py
```diff
@@ -26,13 +26,14 @@ class Semantic(Flag):
...
     Anomaly = auto()
+    Bbox = auto()
+    Polygon = auto()
+    Keypoint = auto()
+    Caption = auto()
```
leoll2: Why do you need these new semantic attribute values? According to the description, `Semantic` attributes exist for disambiguation when multiple fields of the same type exist. Bounding boxes, polygons, ... are already represented by specific fields, what needs to be disambiguated?

src/datumaro/experimental/fields/annotations.py
```diff
@@ -424,3 +429,53 @@ def from_polars(self, name: str, row_index: int, df: pl.DataFrame, target_type:
...
+    def to_polars(self, name: str, value: Any) -> dict[str, pl.Series]:
+        """Convert caption value(s) to Polars series."""
+        dtype = pl.List(pl.Utf8()) if self.is_list else pl.Utf8()
+        return {name: pl.Series(name, [value], dtype=dtype)}
+
+    def from_polars(self, name: str, row_index: int, df: pl.DataFrame, target_type: type[T]) -> T:
+        """Reconstruct caption value(s) from Polars data."""
+        data = df[name][row_index]
+        if self.is_list and target_type is list:
+            return list(data) if data is not None else None  # type: ignore[return-value]
```
leoll2: Is `from_polars` expected to ever return `None`? If yes, I guess the function return signature should be `-> T | None`, else you can raise an exception if data is `None`.

src/datumaro/experimental/data_formats/coco/io.py
```diff
@@ -0,0 +1,179 @@
...
+                instances_by_image=instances_by_image,
+                keypoints_by_image=keypoints_by_image,
+                captions_by_image=captions_by_image,
+                subset=subset,
+            )
+            dataset.append(sample)
+
+        print(f"[COCO] Finished subset '{subset}' with {num_images} samples")
+
+    print(f"[COCO] Finished loading dataset from '{root_dir}' with {len(dataset)} samples in total")
```
leoll2: Please avoid `print` statements

src/datumaro/experimental/fields/base.py
AlbertvanHouten: I added these to disambiguate labels for shapes, not the shapes themselves. However, they ended up not being necessary since all shapes refer to the same object and have the same label.

src/datumaro/experimental/fields/images.py
```diff
@@ -303,3 +303,55 @@ def image_callable_field(format: str = "RGB", semantic: Semantic = Semantic.Defa
         ImageCallableField instance configured with the given parameters
     """
     return ImageCallableField(semantic=semantic, format=format)
+
+
+@dataclass(frozen=True)
+class ImageIdField(Field):
```
leoll2: I suggest to make this field more generic - `IdentifierField` - so that users can represent any kind of identifier with it, not only image ids.

src/datumaro/experimental/fields/base.py
leoll2: If they're not necessary, can you then remove them?

src/datumaro/experimental/data_formats/coco/sample.py
```diff
@@ -0,0 +1,72 @@
...
+    image_info_field,
+    image_path_field,
+    keypoints_field,
+    label_field,
+    polygon_field,
+    subset_field,
+)
+from datumaro.experimental.fields.annotations import caption_field
+from datumaro.experimental.fields.images import image_id_field
+from datumaro.experimental.fields.types import bool_field, numeric_field
```
leoll2: The new fields should be added to `__init__.py`, then you can import them from `datumaro.experimental.fields`

src/datumaro/experimental/data_formats/coco/io.py
```diff
@@ -0,0 +1,182 @@
...
+)
+from datumaro.experimental.data_formats.coco.sample import CocoCategories, CocoSample
+from datumaro.experimental.fields import Subset
+
+logger = logging.getLogger(__name__)
+
+
+def load_coco_dataset(root_dir: str, version: str = "2017") -> Dataset:
+    """
+    Load a COCO dataset (2017 layout) from its top-level directory.
```
leoll2, line 30: Is this method intended to load a generic dataset in COCO format, or only _the_ COCO dataset? I see that it exploits the class `CocoCategories`, which uses the default categories of the COCO dataset. For Datumaro, what we need is the ability to load/store any dataset in COCO format, with custom labels.

src/datumaro/experimental/fields/base.py
AlbertvanHouten: They are not used right now, but might still be useful in the future. Left/Right are also not used at the moment, but we still have them.

leoll2: True, but `Left` and `Right` have an actual meaning and can be used by a client working with stereo data. `Bbox`, `Polygon` and `Keypoint` are already represented by specific fields, it's unclear why there are `Semantic` values for them.

AlbertvanHouten: Like I mentioned above, they can be used to match labels to a shape when there are multiple shapes in a sample: 
```
class CocoSample(Sample):
    # Basic image information
    image: str = image_path_field()
    image_info: ImageInfo = image_info_field()

    # Object detection annotations (from instances_train/val)
    bboxes: np.ndarray | None = bbox_field(dtype=pl.Float32(), format="xywh", semantic=Semantic.Bbox)
    bbox_labels: np.ndarray | None = label_field(dtype=pl.Int32(), is_list=True, semantic=Semantic.Bbox)

    # Polygon segmentation annotations (from instances_train/val)
    polygons: np.ndarray | None = polygon_field(dtype=pl.Float32(), semantic=Semantic.Polygon)
    polygon_labels: np.ndarray | None = label_field(dtype=pl.Int32(), is_list=True, semantic=Semantic.Polygon)
```

AlbertvanHouten: After consideration, I will replace this enum with a string value for the fields with default value "Default".  The enum is too limiting to support every use case and would either need to be regularly extended or 3rd party users need to define their own Semantic enum.

src/datumaro/experimental/fields/images.py
AlbertvanHouten: Good point, I have instead removed identifierfield completely and used a NumericField for it. I don't think its necessary to have a specific implementation just for IDs. I will also remove ScoreField since it too can be replaced with NumericField. Semantics are now strings so they can be easily adjusted to make the intention of the field clear if needed.

src/datumaro/experimental/fields/annotations.py
```diff
@@ -420,7 +373,57 @@ def to_polars(self, name: str, value: Any) -> dict[str, pl.Series]:
...
+    is_list=True. Useful for tasks like image captioning, multi-caption datasets,
+    or storing alternative textual descriptions.
+
+    Attributes:
+        semantic: String tag describing the caption's purpose
+        is_list: Whether this field stores multiple captions (list[str])
+    """
+
+    semantic: str = "default"
+    is_list: bool = False
```
JortBergfeld, line 397: dtype will soon be compulsory, can you please add the dtype here?

src/datumaro/experimental/fields/types.py
```diff
@@ -0,0 +1,112 @@
...
+    """
+    Represents a boolean value or a list of boolean values.
+
+    Stored as Polars Boolean type and
+    converts back to Python ``bool``/``list[bool]`` or numpy arrays as needed
+    by the higher-level Dataset APIs.
+    """
+
+    semantic: str = "default"
+    is_list: bool = False
```
JortBergfeld, line 77: Similar here, can you add a dtype? `dtype: pl.DataType = field(default_factory=pl.Boolean, init=False)`

JortBergfeld (commented):
Looks good to me, but I was wondering if you imported the Coco dataset, what was the performance?

AlbertvanHouten:
> Looks good to me, but I was wondering if you imported the Coco dataset, what was the performance?

Updated description with stats

src/datumaro/experimental/data_formats/coco/io.py
AlbertvanHouten: I forgot to update this for loading coco datasets. Fixed now.

leoll2 (approved):

PR #1962, opened 2025-11-21
Add support for Python 3.14

leoll2:
### Summary

This PR enables compatibility with the latest Python 3.14, while dropping support for the old Python 3.10. Removing 3.10 was necessary because there's no numpy version which supports 3.10 and 3.14 at the same time. This PR also upgrades some dependency packages to a version compatible with 3.14; for breaking changes (e.g. in `shapely`) I have updated the source code.
CI was also updated to test both the minimum (3.11) and maximum (3.14) supported versions.

Resolves #1960 

### Checklist
- [x] I have added tests to cover my changes or documented any manual tests.
- [x] I have updated the [documentation](https://github.com/open-edge-platform/datumaro/tree/develop/docs) accordingly

PR #1968, opened 2025-11-24
Pin dependencies and restore py310 support

leoll2:
In a previous PR #1962 I had removed support for Python 3.10 because incompatible with the latest numpy. However, I realized that by loosening the numpy version constraint a little, we can still keep support for Python 3.10: it is restored by this PR. At the same time, I reordered the dependencies alphabetically and pinned their version with `~=` instead of `>=` to prevent breaking changes (only for dependencies that follow semantic versioning).


pyproject.toml
```diff
@@ -31,33 +11,34 @@ maintainers = [
...
-    "shapely>=2.1.2",
-    "typing_extensions>=3.7.4.3",
-    "tqdm>=4.67.1",
-    "pycocotools>=2.0.4",
-    "PyYAML==6.0.3",
-    "pandas>=2.3.3",
-    "pyarrow>=22.0",
-    "json-stream",
-    "opencv-python-headless>=4.11.0.86",
+    "attrs>=25.3",  # attrs doesn't follow semantic versioning
```
AlbertvanHouten, line 25: I think it's better to pin this package if it doesn't use semantic versioning to prevent future breakage.

leoll2: I thought about this, however pinning a package with `==` in a library would enforce a very strict dependency for client applications that depend on `datumaro`. There is no perfect solution to deal with libraries that don't respect semantic versioning, either you pin the package strictly (client unfriendly) or accept some risk of future breaking changes. In this specific case, the package `attrs` looks quite stable, it doesn't introduce breaking changes very often and when it does, it's for niche features we don't depend on.

AlbertvanHouten (approved):

PR #1972, opened 2025-12-02
Fix RGB/BGR color converter and legacy dataset detection

AlbertvanHouten:
Fix the RGB color converter. It did not properly handle the images returned from the polars data frame which are stored as a flattened list. Also added support for BGR to RGB conversion. 

Make the legacy converter more robust by adding explicit arguments to determine which dataset is being dealt with. It is not always possible to determine a hierarchical dataset just by looking at the dataset.

### Checklist


tests/unit/experimental/test_converters.py
```diff
@@ -82,10 +82,10 @@ def convert(self, df: pl.DataFrame) -> pl.DataFrame:
...
 def test_rgb_to_bgr_converter():
     """Test RGB to BGR format conversion."""
-    converter_instance = RGBToBGRConverter()  # type: ignore[call-arg]
+    converter_instance = RedBlueColorConverter()  # type: ignore[call-arg]
 
     # Create test data
-    rgb_data = np.array([[[255, 0, 0], [0, 255, 0]], [[0, 0, 255], [128, 128, 128]]])
+    rgb_data = np.array([255, 0, 0, 0, 255, 0, 0, 0, 255, 128, 128, 128])
     df = pl.DataFrame(
         {"image": [rgb_data.reshape(-1)], "image_shape": [[2, 2, 3]]},
```
leoll2: `reshape(-1)` seems unnecessary now that your `rgb_data` array is flat

leoll2 (dismissed):

tests/unit/experimental/test_converters.py
AlbertvanHouten: Indeed, removed.

PR #1973, opened 2025-12-03
Fix sample validation for complex types

### LINKED ISSUE #1971: Sample dtype validation fails with optional numpy arrays
Minimal reproducible example:
```python
import polars as pl
import numpy as np
import numpy.typing as npt
from datumaro.experimental import *
from datumaro.experimental.fields import *

class MySample(Sample):
    confidence: npt.NDArray[np.float32] | None = score_field(dtype=pl.Float32(), is_list=True)

s = MySample(confidence=np.array([0.8]))
```
Error:
```
Traceback (most recent call last):
  File "<python-input-0>", line 10, in <module>
    s = MySample(confidence=np.array([0.8]))
  File "/home/leoll2/training_extensions/application/backend/.venv/lib/python3.13/site-packages/datumaro/experimental/dataset.py", line 41, in __init__
    self.validate()
    ~~~~~~~~~~~~~^^
  File "/home/leoll2/training_extensions/application/backend/.venv/lib/python3.13/site-packages/datumaro/experimental/dataset.py", line 67, in validate
    if not self._validate_attribute_type(expected_type, value):
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^
  File "/home/leoll2/training_extensions/application/backend/.venv/lib/python3.13/site-packages/datumaro/experimental/dataset.py", line 89, in _validate_attribute_type
    return isinstance(value, expected_type)
```

JortBergfeld:
This pull request fixes a problem with type validation in the experimental dataset and sample modules. When we use the is_list boolean of score_field, a complex ndarray type is generated that includes an Any type, which is not accepted by isinstance. To circumvent problems with complex types, whenever isintance fails with a type error, we only validate against the origin type (ndarray instead of ndarray[float32], for example)

### Type validation improvements

* Updated the `_validate_attribute_type` method in `src/datumaro/experimental/dataset.py` to correctly handle type validation for generic types by using `origin` when available, improving support for complex type annotations.

### Test enhancements

* Added a new test, `test_sample_with_is_list`, in `tests/unit/experimental/test_sample.py` to verify that samples with list-type fields (using `is_list=True` in `score_field`) are created without validation errors.

Resolves #1971

### DIFF
--- a/src/datumaro/experimental/dataset.py
+++ b/src/datumaro/experimental/dataset.py
@@ -78,15 +78,20 @@ class Sample:
         # Union and Callable types have to be handled separately,
         # because isinstance() does not work with Callable types.
         origin = get_origin(expected_type)
-        if origin is Union:
+        if origin in {Union, types.UnionType}:
             # Check each type in the Union
-            return any(self._validate_attribute_type(typ, value) for typ in get_args(expected_type))
-        if origin in {typing.Callable, collections.abc.Callable} or expected_type in {
+            result = any(self._validate_attribute_type(typ, value) for typ in get_args(expected_type))
+        elif origin in {typing.Callable, collections.abc.Callable} or expected_type in {
             typing.Callable,
             collections.abc.Callable,
         }:
-            return callable(value)
-        return isinstance(value, origin or expected_type)
+            result = callable(value)
+        else:
+            try:
+                result = isinstance(value, expected_type)
+            except TypeError:
+                result = isinstance(value, origin)
+        return result
 
     @classmethod
     @cache
--- a/src/datumaro/experimental/dataset.py
+++ b/src/datumaro/experimental/dataset.py
@@ -86,7 +86,7 @@ class Sample:
             collections.abc.Callable,
         }:
             return callable(value)
-        return isinstance(value, expected_type)
+        return isinstance(value, origin or expected_type)
 
     @classmethod
     @cache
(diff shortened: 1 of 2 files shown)


src/datumaro/experimental/dataset.py
```diff
@@ -78,15 +78,20 @@ def _validate_attribute_type(self, expected_type: Any, value: Any) -> bool:
...
             collections.abc.Callable,
         }:
-            return callable(value)
-        return isinstance(value, expected_type)
+            result = callable(value)
+        else:
+            try:
+                result = isinstance(value, expected_type)
+            except TypeError:
+                result = isinstance(value, origin)
```
leoll2, line 95: For the sake of maintainability, can you add a comment to explain this part? It's otherwise difficult to understand for a developer who doesn't have context about this issue/PR.

leoll2 (commented):
Fix looks good to me, only one comment

tests/unit/experimental/test_sample.py
```diff
@@ -17,6 +18,7 @@
     bbox_field,
     image_field,
     image_info_field,
+    score_field,
```
AlbertvanHouten: score_field has been renamed to numeric_field in https://github.com/open-edge-platform/datumaro/pull/1963

AlbertvanHouten (approved):

PR #1978, opened 2025-12-10
Improve dtype conversions

### LINKED ISSUE #1977: Bbox array becomes 'object' after dtype conversion
After converting a `bbox_field` from one dtype to another, the outer array becomes of type `object` instead of the target data type.

## Example

```python
import numpy as np
import numpy.typing as npt
import polars as pl
from src.datumaro.experimental import *
from src.datumaro.experimental.fields import *
from src.datumaro.experimental.categories import *

class DetectionSampleFloat(Sample):
    bboxes: npt.NDArray[np.float32] = bbox_field(dtype=pl.Float32())

class DetectionSampleInt(Sample):
    bboxes: npt.NDArray[np.int_] = bbox_field(dtype=pl.Int32())

dataset_int = Dataset(DetectionSampleInt)

di_int = DetectionSampleInt(bboxes=np.array([[5, 5, 20, 20], [25, 30, 50, 60]]))
dataset_int.append(di_int)

dataset_float = dataset_int.convert_to_schema(DetectionSampleFloat)


print(dataset_int[0].bboxes, dataset_int[0].bboxes.dtype)
print(dataset_float[0].bboxes, dataset_float[0].bboxes.dtype)
```

Output:
```
[[ 5  5 20 20]
 [25 30 50 60]] int32
[array([ 5.,  5., 20., 20.], dtype=float32)
 array([25., 30., 50., 60.], dtype=float32)] object
```

The `dataset_float[0].bboxes.dtype` should be `float32` but it is instead `object`.

AlbertvanHouten:
This pull request introduces new converters for handling data type conversions for polygons and rotated bounding boxes, and refactors the existing bounding box and label converters to use a shared utility for fixed-size array casting. 

Also fixes the issue described in https://github.com/open-edge-platform/datumaro/issues/1977

### Checklist

### DIFF
--- a/src/datumaro/experimental/dataset.py
+++ b/src/datumaro/experimental/dataset.py
@@ -65,7 +65,7 @@ class Sample:
             field = attr_info.field
 
             if not self._validate_attribute_type(expected_type, value):
-                raise TypeError(f"Attribute `{name}` must be of type `{expected_type}` but is type {type(value)}.")
+                raise TypeError(f"Attribute `{name}` must be of type `{expected_type}` but is type `{type(value)}`.")
 
             # Custom field validation (if any)
             if hasattr(field, "validate"):
--- a/src/datumaro/experimental/dataset.py
+++ b/src/datumaro/experimental/dataset.py
@@ -65,7 +65,7 @@ class Sample:
             field = attr_info.field
 
             if not self._validate_attribute_type(expected_type, value):
-                raise TypeError(f"Attribute `{name}` must be of type `{expected_type}`.")
+                raise TypeError(f"Attribute `{name}` must be of type `{expected_type}` but is type {type(value)}.")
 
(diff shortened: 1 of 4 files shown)


src/datumaro/experimental/dataset.py
```diff
@@ -65,7 +65,7 @@ def validate(self) -> None:
             field = attr_info.field
 
             if not self._validate_attribute_type(expected_type, value):
-                raise TypeError(f"Attribute `{name}` must be of type `{expected_type}`.")
+                raise TypeError(f"Attribute `{name}` must be of type `{expected_type}` but is type {type(value)}.")
```
```suggestion
                raise TypeError(f"Attribute `{name}` must be of type `{expected_type}` but is type `{type(value)}`.")
```

leoll2 (dismissed):
Thanks for the fix!

PR #1981, opened 2025-12-11
Allow LazyImage loading from ImagePathField and add README

AlbertvanHouten:
This pull request introduces support for lazy image loading from an ImagePathField. The main changes add a new `LazyImage` class with global LRU caching, update image field handling to support automatic coercion to `LazyImage`, and provide utilities for cache management. This enables efficient memory usage and flexible image access patterns in dataset samples that have an ImagePathField.  This PR also generated an initial README file describing the experimental module. 

**Lazy image loading and caching:**

* Added new `src/datumaro/experimental/media.py` module, which defines the `LazyImage` class for lazy loading image data, a thread-safe global LRU cache, and utility functions for cache management (`set_image_cache_size`, `clear_image_cache`, `get_image_cache_size`). Also introduced the `ImagePathLike` type alias for flexible type annotations.

**Image field enhancements:**
* Updated the `image_path_field` factory to accept new parameters (`format`, `channels_first`) for configuring lazy image loading behavior.

**Sample initialization improvements:**

* Modified the `Sample` class in `src/datumaro/experimental/dataset.py` to use field-level coercion during initialization, enabling conversion to `LazyImage` when appropriate. This allows users to define ImagePathFields that are typed with `LazyImage` which can then be initialized with a string or Path to the image without throwing validation errors. 

### Checklist


src/datumaro/experimental/media.py
```diff
@@ -0,0 +1,285 @@
...
+    Thread-safe LRU cache for image data.
+
+    This cache limits the total number of images kept in memory.
+    When the cache is full, the least recently used image is evicted.
+    """
+
+    def __init__(self, maxsize: int = 100):
+        self._cache: OrderedDict[tuple[str, str, bool], np.ndarray] = OrderedDict()
+        self._maxsize = maxsize
+        self._lock = Lock()
```
leoll2: Why not use `cachetools` instead of implementing a cache from scratch? Specifically, I think you could apply the `@cachedmethod` decorator to `LazyImage.data`. Another benefit of cachetools is that it allows to control the cache size in bytes rather than number of items, so you can set a default of a few megabytes instead of 100 items (which can be very low or very high, depending on the image size).

src/datumaro/experimental/media.py
```diff
@@ -0,0 +1,285 @@
...
+        Get the shape of the image data.
+
+        Returns:
+            tuple: Shape of the image array. If channels_first is False: (H, W, C).
+                   If channels_first is True: (C, H, W). For grayscale: (H, W).
+        """
+        return self.data.shape
+
+    def __str__(self) -> str:
+        return str(self.path)
```
leoll2:
```suggestion
    def __str__(self) -> str:
        return f"LazyImage({self.path})"
```
An object representation should contain the class name, else the user can be confused about the type.

src/datumaro/experimental/media.py
```diff
@@ -0,0 +1,285 @@
...
+    def width(self) -> int:
+        """Get the image width without fully loading the image data."""
+        with Image.open(self.path) as img:
+            return img.width
+
+    @property
+    def height(self) -> int:
+        """Get the image height without fully loading the image data."""
+        with Image.open(self.path) as img:
+            return img.height
```
leoll2: Since these properties are small and static, you could cache the value in a lazily initialized attribute

```python
@dataclass
class LazyImage:
    ...
    _width: int | None
    _height: int | None

    @property
    def width(self) -> int:
        if self._width is None:
            with Image.open(self.path) as img:
                self._width = img.width
                self._height = img.height
        return self._width
```

src/datumaro/experimental/media.py
```diff
@@ -0,0 +1,285 @@
...
+
+    @property
+    def size(self) -> tuple[int, int]:
+        """Get the image size (width, height) without fully loading the image data."""
+        with Image.open(self.path) as img:
+            return img.size
```
leoll2: I find this property name `size` a bit misleading, I'd expect it to return the image size in bytes instead of the (w,h) tuple. A property that returns (w, h) is also redundant since `shape` already covers the same functionality in a more generic way.

src/datumaro/experimental/media.py
```diff
@@ -0,0 +1,285 @@
...
+        return str(self.path)
+
+    def __fspath__(self) -> str:
+        """Allow LazyImage to be used in os.path operations."""
+        return str(self.path)
+
+
+# Type alias for image path fields that accept strings, Paths, or LazyImage objects.
+# Use this type annotation to avoid type checker warnings when passing strings for LazyImage.
+ImagePathLike: TypeAlias = str | Path | LazyImage
```
leoll2: I don't understand how this type alias helps. `str`, `Path` and `LazyImage` are very different classes with no common methods.

If you define
```python
class TestSample(Sample):
    image: ImagePathLike = image_path_field()
```
and try to access `sample.image.path` or `sample.image.data`, any type checker would raise warnings like "class 'str' doesn't have attribute image".

src/datumaro/experimental/media.py
AlbertvanHouten: Good point, I will rewrite this part and use cachetools.

src/datumaro/experimental/media.py
AlbertvanHouten: The difference between `size` and `shape` is that `size` can be loaded from the image metadata through Pillow, while `shape` requires loading the full image into memory. I agree that the name can be slightly confusing, but I still think it's useful to have this option.

src/datumaro/experimental/media.py
AlbertvanHouten: Thanks, I'll use functools `cached_property` decorator to cache these.

src/datumaro/experimental/media.py
AlbertvanHouten: Yeah, you are right regarding the type checking warnings for when accessing data from the sample. It does help with the input arguments for the sample, but that only fixes half the problem. I will look into a better solution.

AlbertvanHouten: I've opted to just allow passing LazyImage directly to ImagePathField. I think this is the best approach to ensure typing works as expected.

uv.lock
```diff
@@ -45,8 +45,8 @@ name = "astunparse"
 version = "1.6.3"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "six" },
-    { name = "wheel" },
+    { name = "six", marker = "python_full_version < '3.13' and sys_platform != 'darwin'" },
+    { name = "wheel", marker = "python_full_version < '3.13' and sys_platform != 'darwin'" },
```
leoll2: Do you know why `uv lock` is adding these markers? Is there any package incompatible with 3.13 and darwin?

src/datumaro/experimental/media.py
```diff
@@ -0,0 +1,260 @@
+# Copyright (C) 2025 Intel Corporation
+
+#
+# SPDX-License-Identifier: MIT
```
leoll2, line 3:
```suggestion
# Copyright (C) 2025 Intel Corporation
#
# SPDX-License-Identifier: MIT
```

src/datumaro/experimental/media.py
```diff
@@ -0,0 +1,260 @@
...
+        new_cache: LRUCache[tuple[str, str, bool], np.ndarray] = LRUCache(
+            maxsize=maxsize_bytes, getsizeof=_get_image_size
+        )
+        if maxsize_bytes > 0:
+            for key, value in _image_cache.items():
+                try:
+                    new_cache[key] = value
+                except ValueError:
+                    # Cache is full, stop adding items
+                    break
```
leoll2: `new_cache` (`LRUCache`) does not raise a `ValueError` when full; it simply discards the least recently used item. To check if the cache has enough space to store another item without replacement, you need to compare `new_cache.currsize + _get_image_size(value)` and `maxsize_bytes`.

src/datumaro/experimental/media.py
```diff
@@ -0,0 +1,260 @@
...
+        return {
+            "count": len(_image_cache),
+            "current_size": _image_cache.currsize,
+            "max_size": _image_cache.maxsize,
+        }
+
+
+def get_image_cache_size() -> int:
+    """
+    Get the current number of images in the cache.
```
leoll2: It's a bit confusing that `set_image_cache_size` takes bytes as input, while `get_image_cache_size` returns number of items. I suggest to use a consistent definition of "size" as number of bytes, while the number of items can be called "length". Also, to get the length of the cache you don't need to acquire the lock, it's a read-only operation.

src/datumaro/experimental/media.py
```diff
@@ -0,0 +1,260 @@
...
+def _get_image_size(img: np.ndarray) -> int:
+    """Get the size of an image in bytes."""
+    return img.nbytes
+
+
+# Global image cache with byte-size limiting (default 256 MB)
+_image_cache: LRUCache[tuple[str, str, bool], np.ndarray] = LRUCache(
+    maxsize=_DEFAULT_CACHE_SIZE_BYTES, getsizeof=_get_image_size
+)
+_cache_lock = RLock()
```
leoll2: To avoid global variables, and also to expose our cache in more user-friendly way, I suggest to refactor with a static class.

```python
class ImageCache:
    """
    Global LRU cache for lazy-loaded images.

    The size of the cache can be updated dynamically, allowing fine-grained
    control over the memory usage.
    """
    _image_cache: : LRUCache[tuple[str, str, bool], np.ndarray] = LRUCache(
        maxsize=_DEFAULT_CACHE_SIZE_BYTES, getsizeof=_get_image_size
    )
    _cache_lock = RLock()

    @classmethod
    def set_size(cls, maxsize_bytes: int) -> None:
        """Set the size (in bytes)"""
        with _cache_lock:
            ...

    @classmethod
    def get_size(cls) -> int:
        """Get the size (in bytes)"""
        ...

     @classmethod
     def clear(cls) -> None:
         ...

     @classmethod
     def length(cls) -> int:
         return len(cls._cache)
```

src/datumaro/experimental/media.py
leoll2: Nice, clean solution! 👍🏼

src/datumaro/experimental/media.py
```diff
@@ -0,0 +1,260 @@
...
+            tuple: Shape of the image array. If channels_first is False: (H, W, C).
+                   If channels_first is True: (C, H, W). For grayscale: (H, W).
+        """
+        return self.data.shape
+
+    def __str__(self) -> str:
+        return f"{self.path}"
+
+    def __repr__(self) -> str:
+        return f"{self.__class__.__name__}(path={self.path})"
```
leoll2: I think the `__repr__` implementation is enough, for both `str` and `repr` purposes; I don't see why `str(lazy_image)` should return a path without the class name, it's confusing when debugging.

AlbertvanHouten: removed

src/datumaro/experimental/media.py
AlbertvanHouten: Thanks, fixed.

uv.lock
AlbertvanHouten: No idea. They are gone now after updating the lock file.

src/datumaro/experimental/README.md
```diff
@@ -0,0 +1,924 @@
...
+```python
+from datumaro.experimental import (
+    set_image_cache_size,
+    get_image_cache_size,
+    get_image_cache_info,
+    clear_image_cache,
+)
+
+# Set cache to 512 MB
+set_image_cache_size(512 * 1024 * 1024)
```
leoll2: This example needs to be updated to use `ImageCache`

src/datumaro/experimental/README.md
```diff
@@ -0,0 +1,924 @@
...
+
+```python
+# For memory-constrained environments
+set_image_cache_size(10)
+
+# For fast iteration over the same images
+set_image_cache_size(1000)
+
+# Disable caching entirely
+set_image_cache_size(0)
```
leoll2: This example needs to be updated to use `ImageCache`

src/datumaro/experimental/README.md
```diff
@@ -0,0 +1,924 @@
...
+    image_info: ImageInfo = image_info_field()
+```
+
+### 2. Use Lazy Loading for Large Datasets
+
+For datasets with many large images, use path or callable fields:
+
+```python
+class EfficientSample(Sample):
+    image: str = image_path_field()  # Store path, load on demand
```
leoll2:
```suggestion
For datasets with many large images, use path or callable fields:

```python
class EfficientSample(Sample):
    image: LazyImage = image_path_field()  # Store path, load on demand
```
If I'm not wrong, the field must be annotated as `LazyImage` to lazy-load and return an image, else it returns a string path.

src/datumaro/experimental/README.md
AlbertvanHouten: Thanks, fixed.

src/datumaro/experimental/README.md
```diff
@@ -0,0 +1,925 @@
+from datumaro.experimental import LazyImage
+
```
leoll2: This import seems out of place

AlbertvanHouten: Indeed, I didn't see PyCharm added this when I fixed the lazy image sample. Fixed once again now.

leoll2 (approved):

PR #1993, opened 2025-12-23
Fix RotatedBbox.wrap() by assigning correct attributes

williamcorsel:
The `RotatedBbox.wrap()` function tries to access wrong attributes. This causes errors in dataset transforms that include rotated bounding boxes. See the example:

```python
"""Minimal reproducible example demonstrating the bug in RotatedBbox.wrap method.

The wrap method tries to access item.x and item.y, but RotatedBbox only has
cx and cy properties (center coordinates), not x and y.
"""

from datumaro.components.annotation import RotatedBbox

bbox = RotatedBbox(cx=100.0, cy=200.0, w=50.0, h=30.0, r=45.0, label=0)
wrapped_bbox = bbox.wrap(label=1)

```

### Checklist

### DIFF
--- a/src/datumaro/components/annotation.py
+++ b/src/datumaro/components/annotation.py
@@ -1280,7 +1280,7 @@ class RotatedBbox(Shape):
         Returns:
             RotatedBbox: A new RotatedBbox instance with updated attributes.
         """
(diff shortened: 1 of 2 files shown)


leoll2 (approved):
Thank you for the fix!

PR #1994, opened 2025-12-24
Add categories validation

### LINKED ISSUE #1874: Categories validation
Add validation checks to the categories classes ([category.py](https://github.com/open-edge-platform/datumaro/blob/develop/src/datumaro/experimental/categories.py)) to ensure they are well-formed.

For example:
- In `LabelCategories`, all labels in `label_semantics` must be also defined in `labels`
- In `HierarchicalLabelCategories`, all labels in `label_groups` must also exist in `items`
- In `HierarchicalLabelCategories`, every label must belong to exactly one group

JortBergfeld:
This pull request introduces significant improvements to category handling and validation in the experimental Datumaro codebase. The main focus is on enforcing that fields referencing categories (such as labels and masks) use appropriate unsigned integer types, validating that data values correspond to defined categories, and improving the structure and typing of category classes. It also introduces stricter checks and error messages to help catch schema and data mismatches early.

**Category and Field Type Enforcement:**
- Added `BaseLabelCategories` as a base class for label categories, and updated `LabelCategories` and `HierarchicalLabelCategories` to inherit from it, improving type safety and organization. (`src/datumaro/experimental/categories.py`) [[1]](diffhunk://#diff-f0858b860d68536c04476766ef0512d873b7f7cc20dde377a42ec13a836d9622R105-R128) [[2]](diffhunk://#diff-f0858b860d68536c04476766ef0512d873b7f7cc20dde377a42ec13a836d9622L257-R278)
- Enforced that `LabelField` and mask-related fields must use unsigned integer types (e.g., `pl.UInt8`, `pl.UInt32`), and added runtime checks to raise errors if incorrect types are used. (`src/datumaro/experimental/fields/annotations.py`, `src/datumaro/experimental/fields/masks.py`) [[1]](diffhunk://#diff-bc3a3fc9d428403e8e117f703ffc69efaae506a2db15eea759c548e175698e75R167-R174) [[2]](diffhunk://#diff-2a524f5523ceb4ac2f8edce7fcbff9268da416a550e2166ef455df966a2eb9b0R36-R43) [[3]](diffhunk://#diff-2a524f5523ceb4ac2f8edce7fcbff9268da416a550e2166ef455df966a2eb9b0R127-R134) [[4]](diffhunk://#diff-2a524f5523ceb4ac2f8edce7fcbff9268da416a550e2166ef455df966a2eb9b0R196-R203) [[5]](diffhunk://#diff-2a524f5523ceb4ac2f8edce7fcbff9268da416a550e2166ef455df966a2eb9b0R279-R286)

**Category Validation Logic:**
- Implemented validation in `Dataset` to ensure that all values in fields with associated categories do not exceed the number of defined categories, raising informative errors otherwise. (`src/datumaro/experimental/dataset.py`) [[1]](diffhunk://#diff-4ac196ddc4dc8e6d33daf684ded18886ff8774fadb8b6cbd4bfa88ca424bb34fL290-R297) [[2]](diffhunk://#diff-4ac196ddc4dc8e6d33daf684ded18886ff8774fadb8b6cbd4bfa88ca424bb34fR328-R348)
- Added `get_expected_categories_type` methods to fields, and updated `Schema` to check that fields requiring categories have the correct type of categories attached, raising errors if not. (`src/datumaro/experimental/fields/base.py`, `src/datumaro/experimental/fields/annotations.py`, `src/datumaro/experimental/fields/masks.py`, `src/datumaro/experimental/schema.py`) [[1]](diffhunk://#diff-c62fadb5e9fc3ee7144e6fe7cc1e514221b7ccee6c892a64dd734ae5fe5645abR93-R99) [[2]](diffhunk://#diff-bc3a3fc9d428403e8e117f703ffc69efaae506a2db15eea759c548e175698e75R197-R202) [[3]](diffhunk://#diff-2a524f5523ceb4ac2f8edce7fcbff9268da416a550e2166ef455df966a2eb9b0R82-R84) [[4]](diffhunk://#diff-2a524f5523ceb4ac2f8edce7fcbff9268da416a550e2166ef455df966a2eb9b0R161-R163) [[5]](diffhunk://#diff-2a524f5523ceb4ac2f8edce7fcbff9268da416a550e2166ef455df966a2eb9b0R234-R236) [[6]](diffhunk://#diff-2a524f5523ceb4ac2f8edce7fcbff9268da416a550e2166ef455df966a2eb9b0R316-R318) [[7]](diffhunk://#diff-6b005a0f3a04209e7910d1ada68590978a59b2dac628f8953ee07711a2e06d8eR118-R138)

**Field and Data Type Updates:**
- Updated sample field definitions (e.g., in `CocoSample` and tests) to use unsigned integer types for label-related fields. (`src/datumaro/experimental/data_formats/coco/sample.py`, `tests/integration/experimental/test_export_import.py`) [[1]](diffhunk://#diff-335e43b7304266286c3e871d8bde69b5bac61c111c9b3f768a64c52924700347L34-R34) [[2]](diffhunk://#diff-335e43b7304266286c3e871d8bde69b5bac61c111c9b3f768a64c52924700347L43-R43) [[3]](diffhunk://#diff-ef7ef70a007f9bb0c6cf01021e927c6bbd39fbe5164b830620b9abb4d11ee354L752-R752)
- Changed default `label_field` dtype to `pl.UInt8` for consistency and safety. (`src/datumaro/experimental/fields/annotations.py`)

**Mask Category Improvements:**
- Improved `MaskCategories` with proper `__getitem__`, `__iter__`, `__len__`, and `__hash__` implementations, and ensured correct construction and usage in mask converters. (`src/datumaro/experimental/categories.py`, `src/datumaro/experimental/converters/mask_converters.py`) [[1]](diffhunk://#diff-f0858b860d68536c04476766ef0512d873b7f7cc20dde377a42ec13a836d9622L566-L568) [[2]](diffhunk://#diff-f0858b860d68536c04476766ef0512d873b7f7cc20dde377a42ec13a836d9622R641-R652) [[3]](diffhunk://#diff-be0e9424e00f51a63ddbe9dc5c7f01046dc343d7c335eee9ab198a031a4eaa7aL10-R10) [[4]](diffhunk://#diff-be0e9424e00f51a63ddbe9dc5c7f01046dc343d7c335eee9ab198a031a4eaa7aL63-R63)

These changes together provide better type safety, clearer error messages, and stricter validation for category-based fields, reducing the risk of subtle bugs and making the codebase more robust.

### Checklist

### DIFF
--- a/src/datumaro/experimental/categories.py
+++ b/src/datumaro/experimental/categories.py
@@ -534,7 +534,7 @@ class Colormap:
     data: dict[int, RgbColor] = field(default_factory=dict)
 
     def __post_init__(self):
-        """Validate that there are no duplicate colors."""
+        """Validate that there are no duplicate colors"""
         object.__setattr__(self, "_inverse_colormap", {v: k for k, v in self.data.items()})
 
     @property
@@ -634,7 +634,7 @@ class MaskCategories(Categories):
         colormap_dict = generate_colormap(size, include_background=include_background)
         colormap_data = {}
         for index, color in colormap_dict.items():
-            colormap_data[index] = RgbColor(*color) if isinstance(color, tuple) else color
+            colormap_data[index] = RgbColor(*(int(x) for x in color)) if isinstance(color, tuple) else color
 
         colormap = Colormap(data=colormap_data)
         return cls(colormap=colormap)
--- a/src/datumaro/experimental/categories.py
+++ b/src/datumaro/experimental/categories.py
@@ -104,7 +104,16 @@ class Categories:
 
 
 @dataclass(frozen=True)
-class LabelCategories(Categories):
+class BaseLabelCategories(Categories):
+    """
+    Base label categories class.
+
+    This class ensures fields related to labels will have a label categories attached to the attributes spec.
+    """
+
+
+@dataclass(frozen=True)
+class LabelCategories(BaseLabelCategories):
     """
     Represents a group of labels with a specific group type and semantics.
     Use this for simple, non-hierarchical tasks.
@@ -254,7 +263,7 @@ class LabelGroup:
 
 
 @dataclass(frozen=True)
-class HierarchicalLabelCategories(Categories):
+class HierarchicalLabelCategories(BaseLabelCategories):
     """
     Represents hierarchical label categories with groups and parent-child relationships.
     Use this for complex hierarchical classification tasks.
--- a/src/datumaro/experimental/categories.py
+++ b/src/datumaro/experimental/categories.py
@@ -111,6 +111,18 @@ class BaseLabelCategories(Categories):
     This class ensures fields related to labels will have a label categories attached to the attributes spec.
     """
 
+    def __getitem__(self, idx: int) -> Any:
+        """Get label category by index"""
+        raise NotImplementedError
+
+    def __len__(self) -> int:
+        """Get the number of label categories"""
+        raise NotImplementedError
+
+    def __iter__(self):
+        """Iterate over the label categories"""
+        raise NotImplementedError
+
 
 @dataclass(frozen=True)
 class LabelCategories(BaseLabelCategories):
@@ -572,9 +584,6 @@ class MaskCategories(Categories):
     labels: list[str] = field(default_factory=list)
     colormap: Colormap = field(default_factory=Colormap)
 
-    def __hash__(self):
-        return hash((tuple(self.labels), frozenset(self.colormap.items())))
-
     def to_dict(self) -> dict[str, Any]:
         """
         Serialize to a JSON-compatible dictionary.
@@ -629,3 +638,15 @@ class MaskCategories(Categories):
 
         colormap = Colormap(data=colormap_data)
         return cls(colormap=colormap)
+
+    def __getitem__(self, idx: int) -> RgbColor:
+        return self.colormap[idx]
+
+    def __iter__(self):
+        return iter(self.colormap.data.values())
+
+    def __len__(self) -> int:
+        return len(self.colormap)
+
+    def __hash__(self):
+        return hash((tuple(self.labels), frozenset(self.colormap.data.items())))
--- a/src/datumaro/experimental/categories.py
+++ b/src/datumaro/experimental/categories.py
@@ -102,15 +102,6 @@ class Categories:
         # This allows forward compatibility with new category types
         raise ValueError(f"Unknown categories type: {cat_type}")
 
-
-@dataclass(frozen=True)
-class BaseLabelCategories(Categories):
-    """
-    Base label categories class.
-
-    This class ensures fields related to labels will have a label categories attached to the attributes spec.
-    """
-
     def __getitem__(self, idx: int) -> Any:
         """Get label category by index"""
         raise NotImplementedError
@@ -122,6 +113,15 @@ class BaseLabelCategories(Categories):
     def __iter__(self):
         """Iterate over the label categories"""
         raise NotImplementedError
+
+
+@dataclass(frozen=True)
+class BaseLabelCategories(Categories):
+    """
+    Base label categories class.
+
+    This class ensures fields related to labels will have a label categories attached to the attributes spec.
+    """
 
 
 @dataclass(frozen=True)
--- a/src/datumaro/experimental/categories.py
+++ b/src/datumaro/experimental/categories.py
@@ -103,15 +103,15 @@ class Categories:
         raise ValueError(f"Unknown categories type: {cat_type}")
 
     def __getitem__(self, idx: int) -> Any:
-        """Get label category by index"""
+        """Get category by index"""
         raise NotImplementedError
 
     def __len__(self) -> int:
-        """Get the number of label categories"""
+        """Get the number of categories"""
         raise NotImplementedError
 
     def __iter__(self):
-        """Iterate over the label categories"""
+        """Iterate over the categories"""
         raise NotImplementedError
 
 
--- a/src/datumaro/experimental/dataset.py
+++ b/src/datumaro/experimental/dataset.py
@@ -292,7 +292,7 @@ class Dataset(Generic[DType]):
 
         new_row = pl.DataFrame(series_data).cast(dict(self.df.schema))  # type: ignore
 
-        # Validate fields with categories have meaning
+        # Validate fields with categories have integers that refer to existing categories
         self._validate_fields_with_categories(df=new_row)
 
         # Use vstack instead of extend for object columns since extend doesn't support them
@@ -327,7 +327,10 @@ class Dataset(Generic[DType]):
 
     def validate_fields_with_categories(self) -> None:
         """
-        Validates that each value in field columns with categories has meaning.
+        Validates that each integer value in field columns with categories refers to existing categories
+
+        Each field that requires categories should have an unsigned integer dtype, therefore checking the maximum value
+        of each column is lower or equal than the number of associated categories is enough.
         """
         self._validate_fields_with_categories(df=self.df)
 
@@ -337,6 +340,8 @@ class Dataset(Generic[DType]):
         for field_name, categories in fields_with_categories.items():
             if df[field_name].dtype.is_object():
                 continue
+
+            # Label fields can be lists or lists of lists (in the case of multi-labels), so explode column twice.
             field_max = df.select(pl.col(field_name).explode().explode().max()).item()
             # Optional fields can be None, so we can skip the check
             if field_max is not None and len(categories) <= field_max:
--- a/src/datumaro/experimental/dataset.py
+++ b/src/datumaro/experimental/dataset.py
@@ -287,9 +287,13 @@ class Dataset(Generic[DType]):
 
         series_data: dict[str, pl.Series] = {}
         for key, attr_info in self._schema.attributes.items():
-            series_data.update(attr_info.field.to_polars(key, getattr(sample, key)))
+            value = getattr(sample, key)
+            series_data.update(attr_info.field.to_polars(key, value))
 
         new_row = pl.DataFrame(series_data).cast(dict(self.df.schema))  # type: ignore
+
+        # Validate fields with categories have meaning
+        self._validate_fields_with_categories(df=new_row)
 
         # Use vstack instead of extend for object columns since extend doesn't support them
         if any(dtype == pl.Object for dtype in self.df.schema.values()):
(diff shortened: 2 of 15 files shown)


src/datumaro/experimental/schema.py
```diff
@@ -113,6 +115,27 @@ def with_categories(self, categories: dict[str, "Categories"]) -> "Schema":
...
+        """
+        if self._fields_with_categories is None:
+            self._fields_with_categories = {}
+            for attribute, attribute_info in self.attributes.items():
+                if expected_categories_type := attribute_info.field.get_expected_categories_type():
+                    categories = attribute_info.categories
+                    if categories is None or not isinstance(attribute_info.categories, expected_categories_type):
+                        raise ValueError(
+                            f"Expected schema attribute '{attribute}' to have categories defined of type "
+                            f"'{expected_categories_type}', found '{categories}' instead."
```
leoll2:
```suggestion
                            f"'{expected_categories_type}', found '{categories}' instead. "
```

src/datumaro/experimental/fields/masks.py
```diff
@@ -32,6 +33,14 @@ class MaskField(Field):
     channels_first: bool = False
     has_channels_dim: bool = False
 
+    def __post_init__(self) -> None:
+        super().__post_init__()
+        if not (self.dtype.is_(pl.Boolean) or self.dtype.is_unsigned_integer()):
+            raise ValueError(
+                "A mask field's dtype has to be a Boolean or an unsigned integer dtype (e.g. UInt8). These integers "
+                "refer to mask color info provided in mask categories that is defined in a dataset's schema."
+            )
```
leoll2, line 43:
```suggestion
            raise ValueError("MaskField's dtype must be a polars Boolean or unsigned integer type (e.g. UInt8).")
```
Since the definition of a Field/Sample does not depend on Dataset, I would omit the second sentence because it could generate confusion - the user can build a wrong-typed sample and trigger this exception even without creating a Dataset.

src/datumaro/experimental/fields/masks.py
```diff
@@ -242,6 +276,14 @@ class MaskCallableField(Field):
     semantic: str = "default"
     dtype: pl.DataType = field(default_factory=pl.UInt8)
 
+    def __post_init__(self) -> None:
+        super().__post_init__()
+        if not (self.dtype.is_(pl.Boolean) or self.dtype.is_unsigned_integer()):
+            raise ValueError(
+                "A mask field's dtype has to be a Boolean or an unsigned integer dtype (e.g. UInt8). These integers "
+                "refer to mask color info provided in mask categories that is defined in a dataset's schema."
+            )
```
leoll2:
```suggestion
            raise ValueError("MaskCallableField dtype must be a polars Boolean or unsigned integer type (e.g. UInt8).")
```

src/datumaro/experimental/schema.py
```diff
@@ -113,6 +115,27 @@ def with_categories(self, categories: dict[str, "Categories"]) -> "Schema":
                 raise ValueError(f"Attribute '{attr_name}' not found in schema")
         return new_schema
 
+    def get_fields_with_categories(self) -> dict[str, Categories]:
+        """
+        Return a dict containing only attributes of fields that require categories for meaning
+        """
```
leoll2, line 125:
```suggestion
        """
        Get the attributes that semantically require a category.
        
        Returns:
            dict mapping attribute names to the respective required categories.
            Attributes that do not strictly require a categories are not returned.
        """
```

src/datumaro/experimental/schema.py
```diff
@@ -113,6 +115,27 @@ def with_categories(self, categories: dict[str, "Categories"]) -> "Schema":
                 raise ValueError(f"Attribute '{attr_name}' not found in schema")
         return new_schema
 
+    def get_fields_with_categories(self) -> dict[str, Categories]:
```
leoll2: Rename suggestion: `_get_fields_with_required_categories` (the `_` because it's for internal use, it shouldn't be in the public API imho)

src/datumaro/experimental/categories.py
```diff
@@ -102,9 +102,30 @@ def from_dict(cls, data: dict[str, Any]) -> Categories:
...
+        """Get label category by index"""
+        raise NotImplementedError
+
+    def __len__(self) -> int:
+        """Get the number of label categories"""
+        raise NotImplementedError
+
+    def __iter__(self):
+        """Iterate over the label categories"""
+        raise NotImplementedError
```
leoll2, line 115: Base class `Categories` should be generic, without referencing subclasses like label categories. I suggest to move these methods to `BaseLabelCategories`.

src/datumaro/experimental/schema.py
JortBergfeld: Removing the closing " would cause a syntactical error, was this what you intended to suggest?

src/datumaro/experimental/fields/masks.py
JortBergfeld: Given the purpose of datumaro, I think it is safe to assume users will eventually create a dataset, and given that we both fix the dtype and require categories, I think we should be clear about the choices that we made and the intended use of a field/category/dataset. 

As a user, I would first assume a label field would be a string dtype, so knowing where to put what information would be useful to me. This also comes back to our previous discussion, where I suggested to add the categories to the Sample class and use that as a schema. Now to some extend the sample does not have meaning until it is added to a dataset, and depending on the schema, the same sample could have different meaning in different datasets. 

Your suggested error message only helps users who already understand the meaning, and put for example Int32, but won't help users who try to create a field with string or object dtype.

src/datumaro/experimental/schema.py
JortBergfeld: This method is used within the Dataset class and is therefore considered public, I am pretty sure ruff or some other linters would mark it as bad coding if I would turn it into a private method. I think we do not allow schemas to be changed in a dataset, so we could move this method to the dataset class and make it private there?

src/datumaro/experimental/categories.py
JortBergfeld: I forgot to update the description, but I am pretty sure we want these methods for all categories, since we enforce the corresponding fields to be `UInt` or `Boolean`, so that we can do `categories[index]` to get the corresponding category.

JortBergfeld:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/open-edge-platform/datumaro/pull/1994#pullrequestreview-3611084187)

src/datumaro/experimental/schema.py
leoll2: No, my suggestion is to add a trailing space at the end of the sentence. Otherwise the next sentence (after the line) is incorrectly spaced:
image

src/datumaro/experimental/fields/masks.py
leoll2: I see where you're coming from, the string scenario is a good example of potential common mistake which I didn't consider. I suggest this description, what do you think?
`"A mask field's dtype must be a polars Boolean or unsigned integer type (e.g. UInt8). This integer normally represents the index of a category, with reference to the mask categories of the the dataset to which the sample is (or will be) appended."`

src/datumaro/experimental/schema.py
leoll2: I think this method fits better here in `Schema` than `Dataset`. Let's keep it public, but maybe rename to `get_fields_with_required_categories` to highlight that not all the fields are returned in the dict.

src/datumaro/experimental/categories.py
leoll2: I see, then please update the docstring or remove it (built-in methods are self explanatory).

src/datumaro/experimental/fields/annotations.py
```diff
@@ -163,6 +164,14 @@ class LabelField(Field):
     multi_label: bool = False  # Flag to indicate if this field should handle multi-labels
     is_list: bool = False
 
+    def __post_init__(self) -> None:
+        super().__post_init__()
+        if not self.dtype.is_unsigned_integer():
+            raise ValueError(
+                "LabelField's dtype has to be an unsigned integer dtype (e.g. UInt8). These integers refer to label "
+                "info provided in label categories that is defined in a dataset's schema."
```
leoll2:
```
"A label field's dtype must be a polars unsigned integer type (e.g. UInt8). This integer normally represents the index of a category, with reference to the label categories of the dataset to which the sample is (or will be) appended."
```
For consistency with the masks message

src/datumaro/experimental/dataset.py
```diff
@@ -321,6 +325,27 @@ def slice(self, offset: int, length: int | None = None) -> Dataset[DType]:
...
+        """
+        self._validate_fields_with_categories(df=self.df)
+
+    def _validate_fields_with_categories(self, df: pl.DataFrame) -> None:
+        fields_with_categories = self._schema.get_fields_with_required_categories()
+
+        for field_name, categories in fields_with_categories.items():
+            if df[field_name].dtype.is_object():
+                continue
+            field_max = df.select(pl.col(field_name).explode().explode().max()).item()
```
leoll2, line 345: Why `explode()` twice?

leoll2 (commented):
LGTM

src/datumaro/experimental/dataset.py
JortBergfeld: Because the label field can be both `is_multi_label` and `is_list`, causing a list of list of integers. If not a list, the explode command doesn't do anything, so this command can be applied to all types of label field.

leoll2: Makes sense, could you add a comment in the code as well?

src/datumaro/experimental/dataset.py
```diff
@@ -287,10 +287,14 @@ def append(self, sample: DType) -> None:
 
         series_data: dict[str, pl.Series] = {}
         for key, attr_info in self._schema.attributes.items():
-            series_data.update(attr_info.field.to_polars(key, getattr(sample, key)))
+            value = getattr(sample, key)
+            series_data.update(attr_info.field.to_polars(key, value))
 
         new_row = pl.DataFrame(series_data).cast(dict(self.df.schema))  # type: ignore
 
+        # Validate fields with categories have meaning
```
AlbertvanHouten: The "have meaning" part fo this comment is unclear to me. If its not needed, I suggest removing this comment since the name of the function is already clear.

src/datumaro/experimental/dataset.py
```diff
@@ -321,6 +325,27 @@ def slice(self, offset: int, length: int | None = None) -> Dataset[DType]:
 
         return dataset
 
+    def validate_fields_with_categories(self) -> None:
+        """
+        Validates that each value in field columns with categories has meaning.
```
AlbertvanHouten: Could you add a comment here explaining that this function is only checking the maximum value of the categories?

tests/unit/experimental/data_formats/coco/test_coco_io_unit.py
```diff
@@ -81,7 +81,7 @@ def test_load_coco_dataset_errors(tmp_path: Path):
 
 
 def test_save_coco_dataset_writes_expected_structure(tmp_path: Path):
-    dataset = Dataset(CocoSample, categories={"labels": CocoCategories()})
+    dataset = Dataset(CocoSample, categories={"labels": CocoCategories(), "caption_group_ids": CocoCategories()})
```
AlbertvanHouten: This change can be reverted again

AlbertvanHouten (approved):

PR #2007, opened 2026-01-14
YOLO dataset import/export

### LINKED ISSUE #2002: Import/export support for YOLO datasets
Add support for importing and exporting YOLO format datasets under the new Dataset interface

AlbertvanHouten:
This pull request introduces support for the YOLO dataset format (both Ultralytics and traditional variants) in the experimental data formats module. It adds the necessary classes, constants, and I/O functions for reading and writing YOLO datasets, updates the base data format interface to handle YOLO, and ensures licensing headers are consistent across the codebase. The changes also include new unit tests for the YOLO sample class.

Resolves #2002 

### Checklist


src/datumaro/experimental/data_formats/yolo/constants.py
```diff
@@ -0,0 +1,47 @@
+# Copyright (C) 2022-2026 Intel Corporation
+# LIMITED EDGE SOFTWARE DISTRIBUTION LICENSE
```
AlexanderBarabanov: Please check if we need to update license here.

AlbertvanHouten: What would be wrong with it?

AlexanderBarabanov: [The license](https://github.com/open-edge-platform/geti/blob/main/LICENSE) you are going to apply is used in Geti repo and has some restrictions and limitations in contrast with `MIT` (that is OSS-approved license) that is used by datumaro repo. 
Applying this license to datumaro will make this repo not applicable for some use cases - please check p 2.3 from LIMITED EDGE SOFTWARE DISTRIBUTION LICENSE.

So, if your contribution is not a copy-paste from the repo with LIMITED EDGE SOFTWARE DISTRIBUTION LICENSE - i would propose to use `MIT` here.
`# SPDX-License-Identifier: MIT`

AlbertvanHouten: Thanks, I've removed all the geti related licenses from datumaro files.

src/datumaro/experimental/data_formats/base.py
```diff
@@ -58,6 +75,7 @@ def save_dataset(
         data_format: The format to save the dataset in.
         images_dir_path: Path to image directory (str) or dict mapping subset names to paths.
         annotations_path: Path to annotations file (str) or dict mapping subset names to paths.
+        root_dir: Root directory for YOLO format datasets.
```
leoll2, line 78: For COCO format, can you validate that `root_dir` is None? If the user provides it, it's most likely a mistake.

src/datumaro/experimental/data_formats/base.py
```diff
@@ -58,6 +75,7 @@ def save_dataset(
         data_format: The format to save the dataset in.
         images_dir_path: Path to image directory (str) or dict mapping subset names to paths.
         annotations_path: Path to annotations file (str) or dict mapping subset names to paths.
```
leoll2, line 77: Perhaps you can add a note to these two parameters: "Required exclusively for the COCO format".

src/datumaro/experimental/data_formats/yolo/io.py
```diff
@@ -0,0 +1,104 @@
...
+logger = logging.getLogger(__name__)
+
+
+def load_yolo_dataset(root_dir: str, format: str = "auto") -> Dataset:
+    """
+    Load a YOLO dataset from its top-level directory.
+
+    Supports both Ultralytics format and traditional YOLO format:
+
+    Ultralytics format::
```
leoll2, line 31: YOLO Ultralytics is the format described here, right? 
https://docs.ultralytics.com/datasets/#contribute-new-datasets

tests/unit/experimental/data_formats/yolo/test_yolo_helpers.py
```diff
@@ -0,0 +1,507 @@
...
+    assert len(bboxes) == 2
+    assert len(labels) == 2
+    assert labels == [0, 1]
+    # Check first bbox: center_x=0.5*640=320, center_y=0.5*480=240, w=0.2*640=128, h=0.3*480=144
+    assert bboxes[0] == pytest.approx([320.0, 240.0, 128.0, 144.0])
+    # Check second bbox: center_x=0.25*640=160, center_y=0.75*480=360, w=0.1*640=64, h=0.2*480=96
+    assert bboxes[1] == pytest.approx([160.0, 360.0, 64.0, 96.0])
+
+
+def test_parse_yolo_annotation_handles_empty_file(tmp_path: Path):
```
leoll2, line 102: Do you also have a test for _writing_ an empty annotation? Basically, the test should verify that a sample without annotations still produces an (empty) annotation file.

leoll2 (commented):
A few small suggestions, LGTM, thanks

src/datumaro/experimental/data_formats/yolo/io.py
AlbertvanHouten: Yes, the other traditional format is the one that is currently being exported by geti classic.

src/datumaro/experimental/data_formats/base.py
AlbertvanHouten: I'm leaning towards removing this function because it requires passing the format of the dataset anyways which means you might as well call the correct load/save function directly. Especially if we keep adding more datasets which might require more custom arguments. I will keep it for now though and add your suggestions.

src/datumaro/experimental/data_formats/base.py
AlbertvanHouten: I think we can leave it as is, if the user does provide the root_dir but annotations/images dir are still correct it will still work. The same case for importing yolo datasets and passing unnecessary annotation/images dir, as long as root_dir is correct it will work.

src/datumaro/experimental/data_formats/base.py
leoll2: I don't mind removing the `save_dataset` method, since the caller knows the target format thus he can call the specialized save function. `load_dataset` could be also removed, but we need a function to auto-detect the format and possibly dispatch to the specialized loader: such a function would be very helpful for Geti backend, which by design should be agnostic to the internal format details (i.e. Geti is unable to auto-detect the dataset format on its own, it needs help from Datumaro).

leoll2 (approved):

PR #2008, opened 2026-01-14
Ensure new datumaro dataset is picklable

AlbertvanHouten:
OTX was unable to run certain models with multiple workers due to some data within the experimental datumaro dataset not being picklable, mainly polars dataframes with `pl.Object` columns. 

This pull request introduces improvements to serialization and pickling support for datasets and transforms that use Polars DataFrames with Object columns, and refactors mask loader logic to ensure picklability. It also streamlines and enhances the legacy annotation converter and its tests, improving clarity and reliability.

### Checklist

### DIFF
--- a/src/datumaro/experimental/converters/registry.py
+++ b/src/datumaro/experimental/converters/registry.py
@@ -734,6 +734,45 @@ class ConverterTransform(Transform):
 
         self.apply(batch_outputs)
 
+    def __getstate__(self) -> dict:
+        """Prepare the transform for pickling.
+
+        Polars DataFrames with Object columns cannot be serialized using Polars' default
+        serialization. This method extracts Object columns as Python lists before pickling.
+        """
+        state = self.__dict__.copy()
+
+        # Check if DataFrame has Object columns
+        object_columns = [col for col, dtype in self._df.schema.items() if dtype == pl.Object]
+
+        if object_columns:
+            # Extract Object column data as Python lists
+            state["_object_column_data"] = {col: self._df[col].to_list() for col in object_columns}
+            # Create a DataFrame without Object columns for serialization
+            non_object_df = self._df.drop(object_columns)
+            state["_df"] = non_object_df
+            state["_object_columns_schema"] = dict.fromkeys(object_columns, pl.Object)
+        else:
+            state["_object_column_data"] = None
+            state["_object_columns_schema"] = None
+
+        return state
+
+    def __setstate__(self, state: dict) -> None:
+        """Restore the transform after unpickling.
+
+        Reconstructs Object columns from the Python lists stored during pickling.
+        """
+        object_column_data = state.pop("_object_column_data", None)
+        object_columns_schema = state.pop("_object_columns_schema", None)
+
+        self.__dict__.update(state)
+
+        # Restore Object columns if they were extracted
+        if object_column_data is not None and object_columns_schema is not None:
+            for col, values in object_column_data.items():
+                self._df = self._df.with_columns(pl.Series(col, values, dtype=pl.Object()))
+
     def apply(self, fields: Sequence[str]) -> pl.DataFrame:
         required_inputs = set()
         for field in fields:
--- a/src/datumaro/experimental/converters/registry.py
+++ b/src/datumaro/experimental/converters/registry.py
@@ -15,6 +15,7 @@ import polars as pl
 from datumaro.experimental.categories import Categories
 from datumaro.experimental.converters.base import AttributeRemapperConverter, ConversionError, Converter
 from datumaro.experimental.fields.base import Field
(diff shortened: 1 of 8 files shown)


src/datumaro/experimental/dataset.py
```diff
@@ -232,6 +232,47 @@ def __init__(
         self.df = pl.DataFrame(schema=self._generate_polars_schema())
         self._transforms: Transform | None = None
 
+    def __getstate__(self) -> dict[str, Any]:
+        """
+        Prepare the dataset for pickling.
+
+        Polars DataFrames with Object columns cannot be serialized using Polars' default
```
leoll2, line 240: Which columns of the OTX datasets are Object-like?

src/datumaro/experimental/dataset.py
```diff
@@ -232,6 +232,47 @@ def __init__(
...
+        serialization. This method extracts Object columns as Python lists before pickling.
+        """
+        state = self.__dict__.copy()
+
+        # Check if DataFrame has Object columns
+        object_columns = [col for col, dtype in self.df.schema.items() if dtype == pl.Object]
+
+        if object_columns:
+            # Extract Object column data as Python lists
+            state["_object_column_data"] = {col: self.df[col].to_list() for col in object_columns}
```
leoll2: This operation is memory unbounded, it could potentially throw a OOM for large datasets. I can't think of a way to avoid it, just something to keep in mind.

src/datumaro/experimental/dataset.py
```diff
@@ -232,6 +232,47 @@ def __init__(
...
+            # Create a DataFrame without Object columns for serialization
+            non_object_df = self.df.drop(object_columns)
+            state["df"] = non_object_df
+            state["_object_columns_schema"] = dict.fromkeys(object_columns, pl.Object)
+        else:
+            state["_object_column_data"] = None
+            state["_object_columns_schema"] = None
```
leoll2: Are these lines necessary? Your `__setstate__` implementation assumes value `None` when these keys are missing (L266-267).

tests/unit/experimental/test_legacy.py
```diff
@@ -76,38 +77,26 @@ def test_image_media_converter_get_schema_attributes():
     assert attributes["image_path"].type is str
 
 
-def test_image_media_converter_convert_item_media_with_path_and_size():
-    """Test media conversion with path and size."""
+def test_image_media_converter_convert_item_media_with_path():
```
leoll2: The name of these tests can be made more readable by grouping them in a test class:
```python
class TestImageMediaConverter:
    def test_convert_item_media_with_path():
        ...
```

tests/unit/experimental/test_legacy.py
```diff
@@ -76,38 +77,26 @@ def test_image_media_converter_get_schema_attributes():
...
     # Create a dataset to get the converter
     items = [DatasetItem(id="test", media=Image.from_file("/path/to/image.jpg"))]
     dataset = LegacyDataset.from_iterable(items)
 
     converter = ForwardImageMediaConverter.create(dataset)
     assert converter is not None
 
-    image_media = Image.from_file("/path/to/image.jpg", size=(480, 640))  # pyright: ignore[reportUnknownMemberType]
+    # Test with size
+    image_media_with_size = Image.from_file("/path/to/image.jpg", size=(480, 640))  # pyright: ignore[reportUnknownMemberType]
```
leoll2: Datumaro doesn't use Pyright afaik, the error suppression markers can be removed.

src/datumaro/experimental/converters/registry.py
```diff
@@ -734,6 +734,45 @@ def __init__(self, parent: Transform, schema: Schema, conversion_paths: Conversi
 
         self.apply(batch_outputs)
 
+    def __getstate__(self) -> dict:
```
leoll2, line 738: `ConverterTransform` and `Dataset` have an identical implementation of `__getstate__` and `__setstate__`. To avoid code duplicates, you could extract the common logic to a function and call it from both classes.

src/datumaro/experimental/dataset.py
AlbertvanHouten: Lazy referenced images/masks can store a callable as a pl.Object.  

If we convert from legacy dataset to the new one, all images/masks will be stored as a callable.

src/datumaro/experimental/dataset.py
AlbertvanHouten: Removed

tests/unit/experimental/test_legacy.py
AlbertvanHouten: I've grouped all tests under a class now in this file

src/datumaro/experimental/converters/registry.py
AlbertvanHouten: Good point, moved the logic to an utility file.

leoll2 (approved):

PR #2010, opened 2026-01-15
Propagate categories when converting datasets

### LINKED ISSUE #2009: Missing dataset categories after schema conversion
This minimal reproducible example shows that the schema conversion (`convert_to_schema`) does not preserve the label categories.

```python
import numpy as np
import numpy.typing as npt
import polars as pl
from datumaro.experimental import *
from datumaro.experimental.fields import *
from datumaro.experimental.categories import *

class Sample1(Sample):
    label: npt.NDArray[np.int_] = label_field(dtype=pl.UInt8())

class Sample2(Sample):
    label: npt.NDArray[np.int_] = label_field(dtype=pl.UInt8())

dataset_1 = Dataset(
    Sample1,
    categories={"label": LabelCategories(labels=tuple(["car", "truck", "motorbike"]))}
)
dataset_2 = dataset_1.convert_to_schema(Sample2)

print(dataset_1.schema.attributes["label"].categories)
# -> LabelCategories(labels=('car', 'truck', 'motorbike'), group_type=<GroupType.EXCLUSIVE: 0>, label_semantics={})

print(dataset_2.schema.attributes["label"].categories)
# -> None
```
If the categories were correctly handled, both `print` statements should return identical results.

AlbertvanHouten:
This pull request improves category propagation when converting datasets to new schemas in the registry module. The main change is that categories are now automatically merged from the source schema to the target schema when the target does not have explicit categories defined. This ensures that category information is preserved during schema conversion. 

Resolves #2009

### Checklist

### DIFF
--- a/src/datumaro/experimental/converters/registry.py
+++ b/src/datumaro/experimental/converters/registry.py
@@ -117,6 +117,35 @@ class _SchemaState:
         """Get AttributeSpec for a specific field type."""
         return self.field_to_attr_spec.get(field_type)
 
+    def merge_categories_from(self, other: _SchemaState) -> _SchemaState:
+        """
+        Create a new state where None categories are filled from another state.
+
+        This is used to propagate categories from the source schema to the target
+        schema when the target doesn't have explicit categories defined.
+
+        Args:
+            other: The state to take categories from when this state has None
+
+        Returns:
+            A new _SchemaState with merged categories
+        """
+        merged_field_to_attr_spec = {}
+        for field_type, attr_spec in self.field_to_attr_spec.items():
+            if attr_spec.categories is None and field_type in other.field_to_attr_spec:
+                other_attr_spec = other.field_to_attr_spec[field_type]
+                if other_attr_spec.categories is not None:
+                    # Create a new AttributeSpec with the categories from the source
+                    merged_attr_spec = AttributeSpec(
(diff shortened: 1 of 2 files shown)


src/datumaro/experimental/converters/registry.py
```diff
@@ -117,6 +117,35 @@ def get_attr_spec_for_field_type(self, field_type: type[Field]) -> AttributeSpec
...
+        schema when the target doesn't have explicit categories defined.
+
+        Args:
+            other: The state to take categories from when this state has None
+
+        Returns:
+            A new _SchemaState with merged categories
+        """
+        merged_field_to_attr_spec = {}
+        for field_type, attr_spec in self.field_to_attr_spec.items():
```
leoll2, line 134: Unrelated question: how does `field_to_attr_spec` work for samples that have multiple fields of the same kind (e.g, a stereo Sample with two images left and right)?

leoll2 (approved):
Thanks for the fix, it works!

src/datumaro/experimental/converters/registry.py
AlbertvanHouten: If there are multiple fields of the same type, they will have a different semantic. This function is called from within the context of `_find_conversion_path_for_semantic` so it will only propagate them if the semantic is the same.

PR #2015, opened 2026-01-23
ci: Add Collect Library Licenses workflow

mramotowski:
Add Collect Library Licenses workflow
https://github.com/intel-innersource/applications.ai.geti.geti-release/issues/154
### Checklist

https://github.com/mramotowski/datumaro/actions/runs/21285428949/job/61265285603?pr=1

### DIFF
--- a/.github/workflows/collect-sbom-library.yaml
+++ b/.github/workflows/collect-sbom-library.yaml
@@ -1,0 +1,23 @@
+name: Collect Library Licenses
+
+on:
+  schedule:
+    - cron: "0 2 * * 1" # Weekly on Monday
+  workflow_dispatch:
+
+permissions: {}
+
+jobs:
+  collect-licenses:
+    permissions:
+      contents: read
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        with:
+          persist-credentials: false
+
+      - name: Collect library licenses
+        uses: open-edge-platform/geti-ci/actions/collect-sbom-library@eee8dda846963533ec8b4498086f4432e3adea59
+        with:
+          path: .


.github/workflows/collect-sbom-library.yaml
```diff
@@ -0,0 +1,25 @@
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+
+      - name: Collect library licenses
+        uses: open-edge-platform/geti-ci/actions/collect-sbom-library@63eb5023d2b0c0a1bd9f15341add63b167fc9458
+        with:
+          path: .
+          extras: "dev,test,docs,tf,torch,kaggle,scipy,nlp,tabulate,cli,visualizer,h5py,nibabel,protobuf,experimental"
```
AlbertvanHouten: This extras option seems a little brittle. If we add another at some point we need to ensure it is updated. Would it be possible to update the action to automatically discover all options and install them?

AlbertvanHouten (dismissed):

.github/workflows/collect-sbom-library.yaml
mramotowski: We probably can enable `--all-extras` flag if extras flag is empty but then someone would need make it working cause currently on my machine I got error when trying to run it.

.github/workflows/collect-sbom-library.yaml
```diff
@@ -0,0 +1,25 @@
+name: Collect Library Licenses
+
+on:
+  push:
+    branches:
```
AlexanderBarabanov: Maybe we do not need it on push, please add `workflow_dispatch` trigger as well.

.github/workflows/collect-sbom-library.yaml
```diff
@@ -0,0 +1,25 @@
...
+    - cron: "0 2 * * 1" # Weekly on Monday
+
+permissions:
+  contents: read
+
+jobs:
+  collect-licenses:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
```
AlexanderBarabanov: Please use the latest `v6.0.1` version and pin it with hash.
```suggestion
      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
```

AlexanderBarabanov (approved):

PR #2021, opened 2026-01-29
Fix annotation mismatch for multi-polygon annotations

### LINKED ISSUE #5344: Annotations Mismatch in Datumaro dataset
Annotations mismatch in Datumaro dataset. Instance Segmentation training fails due to number of boxes and masks mismatch. Moreover # annotations in dm_subset != # annotations in json file

**Steps to Reproduce**

1. Take real dataset to train on (wgisd in my case, I will provide env and a dataset)
2. Run normal training: otx train --config src/otx/recipe/instance_segmentation/maskrcnn_r50_tv.yaml --data_root ../../detection/wgisd_merged_coco_small
3. Investigate dataset in OTXInstanceSegmentation Dataset
4. Ensure that all annotations are correctly loaded, compare with original json file

**Environment:**

- OS: 
- Framework version: 
- Python version:
- OpenVINO version:
- CUDA/cuDNN version:
- GPU model and memory:

AlbertvanHouten:
This [issue](https://github.com/open-edge-platform/training_extensions/issues/5344) from the OTX repo was facing problems with mismatching annotation counts in legacy datumaro format compared against the new datumaro format. After debugging, the COCO dataset referenced in the issue has images with multi-polygon annotations in a single bounding box. The new datumaro version was only keeping one polygon per bounding box, causing annotation data to be dropped. Below is a sample of the issue where the polygons in the image all belong to one bounding box. The annotations don't really make sense but datumaro was dropping everything but the tiny annotation. 
image

This PR addresses the issue in 2 ways; 

1.  For legacy datumaro datasets, the converter from old to new will now always generate a bounding box for every polygon and save every polygon as a separate annotation. This is not ideal since it drops some information for  instance segmentation but will be safe for further conversion to fit OTX's sample schemas. 
2. ~~For imported COCO datasets into the new format, the multi-polygon annotations will now be stored as is and no longer drop the additional polygons.~~ _second fix is not yet possible and [reverted](https://github.com/open-edge-platform/datumaro/pull/2021/commits/b2e83f648efc328138551049e9c3fb75589477f4), need to be able to store multiple polygons in a single PolygonField_

These multi-polygon COCO annotations need to be revisited at some point. Old datumaro is currently always assuming that each polygon is part of the object while generally the convention for multi-polygon COCO datasets is to have the first polygon be the outline of the object and any polygons thereafter should be interpreted as a hole. 

### Checklist

### DIFF
--- a/src/datumaro/experimental/data_formats/coco/helpers.py
+++ b/src/datumaro/experimental/data_formats/coco/helpers.py
@@ -388,9 +388,9 @@ def _collect_instances_for_image(
     image_id: int,
     instances_by_image: dict[int, list[dict]],
     cat_id_to_idx: dict[int, int],
-) -> tuple[list[list[float] | None], list[np.ndarray], list[int | None], list[float], list[bool]]:
+) -> tuple[list[list[float] | None], list[list[np.ndarray]], list[int | None], list[float], list[bool]]:
     bboxes_list: list[list[float] | None] = []
-    polygons_list: list[np.ndarray] = []
+    polygons_list: list[list[np.ndarray]] = []
     labels_list: list[int | None] = []
     areas_list: list[float] = []
     iscrowd_list: list[bool] = []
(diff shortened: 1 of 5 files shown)


tests/unit/experimental/test_legacy.py
leoll2, line 1: Can you add a unit test for the scenario where a legacy dataset has two polygons inside one bounding box?

src/datumaro/experimental/data_formats/coco/helpers.py
```diff
@@ -399,29 +397,72 @@ def _collect_instances_for_image(
...
+        segmentation = ann.get("segmentation")
 
-        if "area" in ann and isinstance(ann["area"], int | float):
-            areas_list.append(float(ann["area"]))
-        elif bbox is not None and len(bbox) == 4:
-            areas_list.append(float(bbox[2] * bbox[3]))
-        else:
-            areas_list.append(0.0)
+        # Handle multi-part polygons by splitting them into separate annotations
+        polygon_parts = _segmentation_to_poly_parts(segmentation)
```
leoll2, line 403: Please add a unit test for the scenario where a COCO dataset has multiple polygons for the same object.

leoll2 (commented):
About this statement:
> generally the convention for multi-polygon COCO datasets is to have the first polygon be the outline of the object and any polygons thereafter should be interpreted as a hole.

Where did you find this information? In the [COCO spec](https://cocodataset.org/#format-data), I couldn't find any mention of such convention, instead it says:
> Note that a single object (iscrowd=0) may require multiple polygons, for example if occluded.

Which implies that no hierarchy exists between such polygons; they represent visually disconnected regions of the same physical object - e.g. consider this example with _one_ car occluded by a tree: two polygons are needed to precisely outline the car visible region, whereas the whole object is captured by one bounding box.

Image

In this example, you could reduce it to a single polygon if you accept to include part of the occluding object (tree) in the area, but this approach may not be feasible for all use situations (e.g. if the occluding object is relatively large compared to the relevant object, and the visible part is very fragmented).

AlbertvanHouten:
> Where did you find this information? In the [COCO spec](https://cocodataset.org/#format-data), I couldn't find any mention of such convention, instead it says:

I got this from claude but it seems to not be entirely true. It depends on the dataset and different conventions are being used.

src/datumaro/experimental/data_formats/coco/helpers.py
AlbertvanHouten: added

tests/unit/experimental/test_legacy.py
AlbertvanHouten: added

leoll2 (approved):

PR #2024, opened 2026-02-02
Fix image paths of exports with images exported

### LINKED ISSUE #2019: Export dataset does not handle image paths properly
Consider the following script, which exports a Dataset that contains an image path field.
```python
import numpy as np
import polars as pl

from datumaro.experimental import Dataset, Sample
from datumaro.experimental.export_import import export_dataset
from datumaro.experimental.fields import image_path_field


class SampleWithPath(Sample):
    img: str = image_path_field()

sample = SampleWithPath(img="/home/leoll2/media/cat/cat1.jpg")
dataset = Dataset(SampleWithPath)
dataset.append(sample)

export_dataset(
    dataset=dataset,
    output_path="./exported_dataset",
    as_zip=False,
    export_images=True,
)
```
If you examine the Parquet file of the exported dataset, it looks like this:

Image

The path is incorrect because it points to the original media file instead of the exported one, namely `./images/img_000000.jpg` relative to `exported_dataset`.

The logic of the `export_dataset` method shall be adjusted as follows:
-  if `export_images=False`, the path written to the Parquet file should point to the original image, as in the current implementation (example: `/home/leoll2/media/cat/cat1.jpg`); the dataset can be then reloaded with `import_dataset`, provided that the image still exists at the original location.
- if `export_images=True`, the path should be relative to the export folder (example: `./images/img_000000.jpg`); the dataset should be portable, i.e. it should be always possible to relocate the dataset and reload it independently from the original media.

JortBergfeld:
This pull request refines the image export logic in the experimental export/import module, ensures consistent handling of image paths, and introduces a new integration test to verify correct behavior when exporting datasets with or without images. The changes improve both code correctness and test coverage.

**Image Export Logic Improvements:**

* Changed the image export path construction in `_export_image_path_field` to use string conversion for `Path` and updated the naming convention to use underscores instead of slashes, ensuring paths are consistent and safe for file systems.
* Updated the export logic in `export_dataset` to correctly update DataFrame image path fields with their new relative paths when `export_images=True`, using a left join with a mapping DataFrame, and to retain original absolute paths when `export_images=False`.

**Test Adjustments and Additions:**

* Modified integration tests to match the new image path convention, checking for the correct file existence and path format in exported datasets.
* Added a new integration test `test_export_dataset_export_images_true_and_false` to verify that image paths are stored as relative paths in Parquet when images are exported, and as absolute paths otherwise. This test also checks that the exported files exist in the correct location.

Resolves #2019 

### Checklist

### DIFF
--- a/src/datumaro/experimental/export_import.py
+++ b/src/datumaro/experimental/export_import.py
@@ -82,10 +82,9 @@ def _export_image_path_field(
             return None
 
         extension = source_path.suffix if source_path.suffix else ".png"
-        rel_path = f"{field_name}/{idx:06d}{extension}"
+        rel_path = f"{field_name}_{idx:06d}{extension}"
         abs_path = output_dir / rel_path
 
-        abs_path.parent.mkdir(parents=True, exist_ok=True)
         shutil.copy2(source_path, abs_path)
         return str(rel_path)
     except Exception as e:
@@ -251,7 +250,7 @@ def export_dataset(
                 if not path_map:
                     continue
 
-                # Create a mapping DataFrame from the exported paths (O(ExportedCount) vs O(DatasetCount))
+                # Create a mapping DataFrame from the exported paths
                 mapping_df = pl.DataFrame(
                     {"__idx": list(path_map.keys()), "__new_path": list(path_map.values())},
                     schema={"__idx": pl.UInt32, "__new_path": pl.String},
@@ -390,7 +389,7 @@ def _reconstruct_field_values(
 
     values: list[object | None] = []
     for idx in range(num_rows):
-        pattern = f"{field_name}/{idx:06d}.*" if is_path_field else f"{field_name}_{idx:06d}.*"
+        pattern = f"{field_name}_{idx:06d}.*"
         matches = sorted(images_base_dir.glob(pattern))
         file_path = matches[0] if matches else None
 
--- a/src/datumaro/experimental/export_import.py
+++ b/src/datumaro/experimental/export_import.py
@@ -76,7 +76,7 @@ def _export_image_path_field(
 ) -> str | None:
     """Export an ImagePathField by copying the file from the filesystem."""
     try:
-        source_path = Path(value)
+        source_path = Path(str(value))
         if not source_path.exists():
             print(f"Warning: Image file not found: {value}")
             return None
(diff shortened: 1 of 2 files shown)


src/datumaro/experimental/export_import.py
```diff
@@ -76,14 +76,14 @@ def _export_image_path_field(
...
     try:
-        source_path = Path(value)
+        source_path = Path(str(value))
         if not source_path.exists():
             print(f"Warning: Image file not found: {value}")
             return None
 
         extension = source_path.suffix if source_path.suffix else ".png"
-        rel_path = f"{field_name}/{idx:06d}{extension}"
-        abs_path = output_dir / rel_path.replace("/", "_")
```
JortBergfeld, line 86: @AlbertvanHouten I do not understand why for image paths the relative path is defined with a slash, but it is saved on disk with an underscore. However, if we export images, the relative path in the dataframe needs to be updated the location to which the image is actually stored.

src/datumaro/experimental/export_import.py
```diff
@@ -239,19 +239,39 @@ def export_dataset(
...
+        df_to_export = dataset.df
         # Export images if requested
         if export_images:
             images_dir = work_dir / IMAGES_DIR
-            _export_images_from_dataset(dataset, images_dir)
+            images_paths = _export_images_from_dataset(dataset, images_dir)
+
+            df_to_export = df_to_export.with_row_index("__idx")
+            for field_name, path_map in images_paths.items():
+                if not path_map:
```
leoll2, line 250: In what cases does `path_map` evaluate false?

src/datumaro/experimental/export_import.py
```diff
@@ -239,19 +239,39 @@ def export_dataset(
...
+                mapping_df = pl.DataFrame(
+                    {"__idx": list(path_map.keys()), "__new_path": list(path_map.values())},
+                    schema={"__idx": pl.UInt32, "__new_path": pl.String},
+                )
+
+                # Left join to apply updates; rows not in path_map will have null paths
+                df_to_export = (
+                    df_to_export.join(mapping_df, on="__idx", how="left")
+                    .with_columns(pl.col("__new_path").alias(field_name))
+                    .drop("__new_path")
```
leoll2, line 263: Isn't `.with_column(pl.col("__new_path").alias(field_name)).drop("__new_path")` equivalent to `.rename({"__new_path": field_name})`?

src/datumaro/experimental/export_import.py
```diff
@@ -239,19 +239,39 @@ def export_dataset(
...
+                # Create a mapping DataFrame from the exported paths
+                mapping_df = pl.DataFrame(
+                    {"__idx": list(path_map.keys()), "__new_path": list(path_map.values())},
+                    schema={"__idx": pl.UInt32, "__new_path": pl.String},
+                )
+
+                # Left join to apply updates; rows not in path_map will have null paths
+                df_to_export = (
+                    df_to_export.join(mapping_df, on="__idx", how="left")
+                    .with_columns(pl.col("__new_path").alias(field_name))
```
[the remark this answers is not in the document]
JortBergfeld: We do not want the old path if there is no new path, since then probably the export of the image failed.

src/datumaro/experimental/export_import.py
JortBergfeld: I will double check, but this was mainly for performance in case the path map is empty, we do not want to do a full join with all null values. But perhaps I should double check, perhaps we do want to update the column to all null values.

src/datumaro/experimental/export_import.py
JortBergfeld: No, rename only works if the column doesn't exist yet. Also, the join replaces the old pahts with null values if there is no new path. Update could work if we wanted to keep the old path. But now that I think about it, we probably want to do an inner join. If there is no new path, the export of the image failed, and we should probably drop the row. However, this also contains new paths for masks and image callables. I am not sure if we always want to drop a row if there is no new path for one of the other type of fields.

leoll2 (approved):

JortBergfeld:
This PR will be merged as is, because it fixes the bug described in the ticket. However, the join method used here will quietly skip missing images and masks. We will continue investigating the different use cases we have to support and implement a more robust solution in another PR.

PR #2027, opened 2026-02-05
Add StringField

leoll2:
This PR implements a new field type `StringField` which to store arbitrary strings. It can be used to attach metadata to a dataset sample, or even an id.

### Checklist

- [x] I have updated the [documentation](https://github.com/open-edge-platform/datumaro/tree/develop/docs) accordingly

### DIFF
--- a/src/datumaro/experimental/fields/types.py
+++ b/src/datumaro/experimental/fields/types.py
@@ -16,9 +16,12 @@ class NumericField(Field):
     """
     Represents a numeric value with arbitrary semantics.
 
-    Use this for generic per-sample or per-instance numeric attributes such as
-    area, depth, distance, score, etc. Supports scalar values or lists of values when
-    ``is_list=True``.
+    This field can be used, for instance, to represent attributes like area, depth, distance, score, etc.
+
+    Attributes:
+        semantic: A string tag describing the purpose of the value (e.g., "area")
+        dtype: Polars data type for the numeric value (e.g., pl.Float32, pl.Int64)
+        is_list: If True, the field can hold a list of numeric values instead of a single value.
     """
 
     semantic: str = "default"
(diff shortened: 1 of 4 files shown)


src/datumaro/experimental/fields/types.py
```diff
@@ -111,3 +111,51 @@ def bool_field(
...
+        return {name: pl.Series(name, [value], dtype=self._pl_type)}
+
+    def from_polars(self, name: str, row_index: int, df: pl.DataFrame, target_type: type[T]) -> T:
+        data = df[name][row_index]
+        if target_type is list:
+            return list(data) if data is not None else None  # type: ignore[return-value]
+        return from_polars_data(data, target_type)
+
+
+def string_field(
```
JortBergfeld, line 151: Do you know if pyrefly properly handles the types of these fields? To be honest, I don't even remember what type checking problem these methods fix, except the return type has to be `Any` instead of the `Field` type. I find it pretty ugly to introduce these methods that just return the field.

src/datumaro/experimental/fields/types.py
JortBergfeld: If you're fixing the docstring, perhaps also rewrite this as the others:
`return pl.List(self.dtype) if self.is_list else self.dtype`

JortBergfeld (dismissed):

src/datumaro/experimental/fields/types.py
leoll2: Updated!

src/datumaro/experimental/fields/types.py
leoll2: I don't know, it's something to check when we have time. There are many abstractions and magic code in the Datumaro internals, it can be quite difficult for the type checkers to infer the intended types for each field.

JortBergfeld (approved):

PR #2031, opened 2026-02-10
Add support for saving datasets as zip archives

### LINKED ISSUE #2026: Add optional zipping to save_dataset (parity with export_dataset(..., as_zip=True)
**Is your feature request related to a problem? Please describe.**

In Geti Tune, we export datasets using `datumaro`, and we rely on two different APIs:
- `save_dataset(...)` for formats like `COCO` and `YOLO` (`VOC` TBD)
- `export_dataset(..., as_zip=True)` for `DATUMARO_V2`

For `DATUMARO_V2` we can directly produce a zipped artifact. However, for `COCO` and `YOLO` we currently need to:
1. Call `save_dataset(...)` into a target directory.
2. Walk the directory and create a .zip manually.
3. Clean up the target directory non-zipped contents afterwards.

**Describe the solution you'd like to propose.**

Extend `save_dataset` with an optional parameter (e.g. `as_zip: bool = False`, and optionally `zip_name: str | None = None`) that:
1. Writes the dataset to a temporary directory.
2. Produces a zip archive at the target location.
3. Removes the temporary directory after zipping.

This would:
- Align the behavior of `save_dataset` with `export_dataset(..., as_zip=True)`.
- Simplify downstream code (no need for manual zipfile handling and cleanup).
- Provide a consistent way to obtain zipped exports across different formats (`COCO`, `YOLO`, etc.) using a single, high-level API.

**Describe alternatives you've considered**

**Additional context**

AlbertvanHouten:
Adds support for zipping exported datasets. Also refactors the save dataset api by removing images/annotation/root_dir paths and assuming default paths for the data formats. 

resolves #2026 

### Checklist

### DIFF
--- a/src/datumaro/experimental/data_formats/base.py
+++ b/src/datumaro/experimental/data_formats/base.py
@@ -82,6 +82,9 @@ def save_dataset(
     Raises:
         ValueError: If the data format is not supported.
     """
+    if data_format not in DataFormat:
+        raise ValueError(f"Unsupported data format: {data_format}")
+
     output_dir = os.path.dirname(output_path) if as_zip else output_path
     if output_dir:
         os.makedirs(output_dir, exist_ok=True)
@@ -107,11 +110,10 @@ def _save_dataset_as_zip(
 ) -> None:
     """Save dataset as a zip archive."""
     # Strip .zip extension if present since shutil.make_archive adds it automatically
-    if output_path.endswith(".zip"):
+    if output_path.lower().endswith(".zip"):
         output_path = output_path[:-4]
 
-    temp_dir = tempfile.mkdtemp()
-    try:
+    with tempfile.TemporaryDirectory() as temp_dir:
         _save_dataset_to_dir(
             dataset=dataset,
             data_format=data_format,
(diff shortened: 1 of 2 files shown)


src/datumaro/experimental/data_formats/base.py
```diff
@@ -63,42 +66,86 @@ def load_dataset(
...
+
+def _save_dataset_as_zip(
+    dataset: Dataset,
+    data_format: DataFormat,
+    output_path: str,
+) -> None:
+    """Save dataset as a zip archive."""
+    # Strip .zip extension if present since shutil.make_archive adds it automatically
+    if output_path.lower().endswith(".zip"):
+        output_path = output_path[:-4]
```
leoll2: FYI you can use [`os.path.splitext`](https://docs.python.org/3/library/os.path.html#os.path.splitext) to split the path root from the extension.

leoll2 (dismissed):

src/datumaro/experimental/data_formats/base.py
AlbertvanHouten: thx, fixed.

leoll2 (approved):

PR #2035, opened 2026-02-12
Enforce semantics for generic fields and add bool polars converter

AlbertvanHouten:
This PR makes the semantic required for BoolField, NumericField and Stringfield. The default value could cause problems when converting between different Schemas.

Also adds a bool from_polars converter and adds tests for the generic fields. 

Adds support to filter datasets by list/tuple of subsets 

### Checklist

### DIFF
--- a/src/datumaro/experimental/data_formats/coco/sample.py
+++ b/src/datumaro/experimental/data_formats/coco/sample.py
@@ -9,6 +9,7 @@ import polars as pl
 from datumaro.experimental import LazyImage, Sample
 from datumaro.experimental.categories import LabelCategories
 from datumaro.experimental.data_formats.coco.constants import COCO_LABEL_TO_SUPER
+from datumaro.experimental.data_formats.semantics import AREAS, CAPTION_GROUP_IDS, IMAGE_ID, ISCROWD
 from datumaro.experimental.fields import (
     ImageInfo,
     Subset,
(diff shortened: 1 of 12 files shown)


leoll2 (dismissed):

itallix (commented):
Could you please also include `bool_field` & `BoolField` to exports in `datumaro.experimental.fields`. This seems to be missing.

itallix (dismissed):

src/datumaro/experimental/data_formats/coco/sample.py
```diff
@@ -34,19 +35,19 @@ class CocoSample(Sample):
...
     # Caption annotations (from captions_train/val)
-    captions: np.ndarray | None = caption_field(is_list=True, semantic="caption")
-    caption_group_ids: np.ndarray | None = numeric_field(dtype=pl.UInt32(), is_list=True, semantic="caption")
+    captions: np.ndarray | None = caption_field(is_list=True)
+    caption_group_ids: np.ndarray | None = numeric_field(dtype=pl.UInt32(), is_list=True, semantic=CAPTION_GROUP_IDS)
 
     # Dataset organization
     subset: Subset = subset_field()
-    image_id: int | None = numeric_field(dtype=pl.Int32(), semantic="image_id")
+    image_id: int | None = numeric_field(dtype=pl.Int32(), semantic=IMAGE_ID)
```
leoll2, line 50: Does `image_id` have to be integer (`numeric_field`) or can it be generalized to string (`string_field`)?

src/datumaro/experimental/dataset.py
```diff
@@ -570,23 +570,27 @@ def convert_to_schema(
             categories=inferred_categories,
         )
 
-    def filter_by_subset(self, subset: Subset) -> Dataset[DType]:
+    def filter_by_subset(self, subset: Subset | list[Subset] | tuple[Subset, ...]) -> Dataset[DType]:
```
```suggestion
    def filter_by_subset(self, subset: Subset | Sequence[Subset]) -> Dataset[DType]:
```

leoll2 (commented):
Minor comments, LGTM

src/datumaro/experimental/data_formats/coco/sample.py
AlbertvanHouten: For compatibility with COCO it has to be numeric. Converters will check for same field type first, then semantic. If the field type is different in the source dataset, converting to COCO will result in missing image_ids.

leoll2 (approved):

PR #2041, opened 2026-02-17
Automatic sample type detection on data import

AlbertvanHouten:
This pull request introduces automatic detection of the correct `Sample` subclass when importing datasets, improving extensibility and usability for custom dataset schemas. It adds a registration mechanism for `Sample` subclasses, updates the import logic to match schemas to registered types, and provides comprehensive tests for these features. The changes also fix schema serialization/deserialization to properly support union types with None values.

NOTE: Datasets exported before this PR is merged with union types that can be None can still throw an error. Only newly exported datasets will work correctly. I don't think its necessary to add backwards compatible logic at this time. 

### Checklist

### DIFF
--- a/src/datumaro/experimental/export_import.py
+++ b/src/datumaro/experimental/export_import.py
@@ -61,6 +61,11 @@ def register_sample(cls: type[Sample]) -> type[Sample]:
             image: Annotated[np.ndarray, ImageField()]
             label: Annotated[int, ScalarField()]
     """
+    # Validate that only proper Sample subclasses (excluding Sample itself)
+    # can be registered. This prevents invalid entries that could later cause
+    # failures during schema inference.
+    if not isinstance(cls, type) or not issubclass(cls, Sample) or cls is Sample:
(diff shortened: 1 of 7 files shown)


src/datumaro/experimental/export_import.py
```diff
@@ -36,6 +37,39 @@
...
+
+        @register_sample
+        class MySample(Sample):
+            image: Annotated[np.ndarray, ImageField()]
+            label: Annotated[int, ScalarField()]
+    """
+    # Validate that only proper Sample subclasses (excluding Sample itself)
+    # can be registered. This prevents invalid entries that could later cause
+    # failures during schema inference.
+    if not isinstance(cls, type) or not issubclass(cls, Sample) or cls is Sample:
```
leoll2, line 67: The last condition (`cls is Sample`) seems redundant: the `Sample` class is defined in Datumaro, so a user can't mistakenly decorate it with `@register_sample`.

leoll2 (approved):

PR #2047, opened 2026-02-20
Add label filtering functionality to the dataset class

AlbertvanHouten:
This pull request introduces a new `filter_by_labels` method to the `Dataset` class, enabling users to filter datasets based on label values. The implementation is robust, supporting a variety of label field configurations (single label, multi-label, list fields, and hierarchical categories), and includes comprehensive validation and error handling. Extensive unit tests have been added to verify functionality, edge cases, and error conditions.

**New filtering functionality:**

- Added the `filter_by_labels` method to the `Dataset` class, allowing filtering of dataset items by label names or indices. The method supports auto-detection of the label field when the schema contains exactly one `LabelField`, and works with both `LabelCategories` and `HierarchicalLabelCategories`.

Required for https://github.com/open-edge-platform/training_extensions/issues/5564

### Checklist

### DIFF
--- a/src/datumaro/experimental/dataset.py
+++ b/src/datumaro/experimental/dataset.py
@@ -692,7 +692,7 @@ class Dataset(Generic[DType]):
                 label_indices = expand_indices_with_ancestors(categories, label_indices)
 
             # Sort indices to maintain a consistent order in the new categories
-            sorted_indices = sorted(label_indices)
+            sorted_indices = sorted(set(label_indices))
 
             # Create mapping from old indices to new indices (0, 1, 2, ...)
             old_to_new_index_map = {old_idx: new_idx for new_idx, old_idx in enumerate(sorted_indices)}
--- a/src/datumaro/experimental/dataset.py
+++ b/src/datumaro/experimental/dataset.py
@@ -692,37 +692,74 @@ class Dataset(Generic[DType]):
     ) -> pl.DataFrame:
         """Return a filtered DataFrame keeping rows that match any of *label_indices*.
 
+        This method both filters rows (keeping only those with at least one matching label)
+        AND removes non-matching labels from within each sample's label field.
+
         The filtering strategy is chosen based on the ``is_list`` and
         ``multi_label`` flags of *label_field_instance*.
         """
         col = pl.col(label_field_name)
 
         if label_field_instance.is_list and label_field_instance.multi_label:
-            # List(List(UInt)) - explode outer list, check inner lists, re-aggregate.
-            mask = (
-                self.df.with_row_index("__row_idx__")
-                .explode(label_field_name)
-                .with_columns(
-                    pl.col(label_field_name).list.eval(pl.element().is_in(label_indices)).list.any().alias("__match__")
+            # List(List(UInt)) - for each inner list, keep only matching elements,
+            # then filter out empty inner lists, then filter out rows with no remaining labels.
+            return self.df.with_columns(
+                col.list.eval(
+                    pl.element()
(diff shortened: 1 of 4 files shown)


src/datumaro/experimental/dataset.py
```diff
@@ -599,6 +601,300 @@ def filter_by_subset(self, subset: Subset | Sequence[Subset]) -> Dataset[DType]:
...
+
+        attr_info = self.schema.attributes[label_field_name]
+        if not isinstance(attr_info.field, LabelField):
+            raise TypeError(f"Attribute '{label_field_name}' is a {type(attr_info.field).__name__}, not a LabelField.")
+
+        return label_field_name
+
+    @staticmethod
+    def _resolve_label_indices(
+        labels: Sequence[str | int],
```
JortBergfeld: I suppose it should be `Sequence[int] | Sequence[str]`, to ensure we do not have to deal with the duplicate label problem described above by copilot

src/datumaro/experimental/dataset.py
```diff
@@ -599,6 +601,300 @@ def filter_by_subset(self, subset: Subset | Sequence[Subset]) -> Dataset[DType]:
...
+    ) -> pl.DataFrame:
+        """Return a filtered DataFrame keeping rows that match any of *label_indices*.
+
+        This method both filters rows (keeping only those with at least one matching label)
+        AND removes non-matching labels from within each sample's label field.
+
+        The filtering strategy is chosen based on the ``is_list`` and
+        ``multi_label`` flags of *label_field_instance*.
+        """
+        col = pl.col(label_field_name)
```
JortBergfeld: Prefer to call this `label_col` as it is easy to confuse `col` with the Polars method.

src/datumaro/experimental/dataset.py
```diff
@@ -599,6 +601,300 @@ def filter_by_subset(self, subset: Subset | Sequence[Subset]) -> Dataset[DType]:
...
+        Args:
+            df: The DataFrame to modify
+            label_field_name: Name of the label column
+            label_field_instance: The LabelField instance
+            old_to_new_index_map: Mapping from old indices to new indices
+
+        Returns:
+            DataFrame with remapped label indices
+        """
+        col = pl.col(label_field_name)
```
JortBergfeld: Similar here, I prefer `label_col`

src/datumaro/experimental/dataset.py
```diff
@@ -599,6 +601,300 @@ def filter_by_subset(self, subset: Subset | Sequence[Subset]) -> Dataset[DType]:
...
+        Example:
+            ```python
+            # Auto-detect (schema has exactly one LabelField)
+            filtered = dataset.filter_by_labels(["cat", "dog"])
+
+            # Using label indices
+            filtered = dataset.filter_by_labels([0, 1])
+
+            # Mixing label names and indices
+            filtered = dataset.filter_by_labels(["cat", 1])
```
JortBergfeld: I'm not sure why we would support this and certainly not why we would promote its usage.

JortBergfeld (commented):
Looks good, but in general I feel that we are adding too much code to the dataset class for this one feature. Perhaps we can split at least the private helper methods into a different module/class, or everything and just keep a helper method that calls the other class. After all, it returns a new dataframe, it doesn't update it, so it could be an independent function.

BTW, my comments are minor, but not approving as I am waiting for your response to copilot's comments

src/datumaro/experimental/dataset.py
AlbertvanHouten: I've removed the example.

JortBergfeld (approved):

PR #2063, opened 2026-03-03
Support Pascal VOC dataset format for import/export

### LINKED ISSUE #2003: Import/export support for Pascal VOC datasets
Add support for importing and exporting datasets in Pascal VOC format.

AlbertvanHouten:
This pull request adds support for the Pascal VOC dataset format to the experimental data formats module. It introduces new classes, constants, and I/O functions for VOC, updates the core dataset loading and saving logic to handle VOC datasets, and enhances field semantics to support shared categories (important for segmentation masks). 

**Pascal VOC format support:**

* Added a new `VOC` entry to the `DataFormat` enum and extended the dataset loading and saving logic in `base.py` to support VOC datasets, including both standard and simple directory layouts. 
* Implemented VOC dataset I/O functions in `voc/io.py`, supporting loading and saving VOC datasets, with full documentation and error handling.
* Defined VOC-specific constants (directory names, label lists, etc.).
* Added a new `VocSample` class and `VocCategories` for representing VOC data, including support for VOC-specific fields such as `difficult`, `truncated`, `occluded`, and `pose`, as well as lazy-loaded segmentation masks.
* Added VOC-specific semantic tags (`DIFFICULT`, `TRUNCATED`, `OCCLUDED`, `POSE`) for use in annotations.

**Field and schema enhancements:**

* Extended the `Field` base class to support a `categories_from` attribute, allowing fields (such as `instance_mask`) to share categories with another field (such as `class_mask`).
* Improved mask field handling to check for `None` values when validating callables.

**Import/export and format detection improvements:**

* Added `import_voc_dataset` and updated the dataset import logic to detect and import VOC datasets.

Resolves #2003

### Checklist


src/datumaro/experimental/data_formats/voc/sample.py
```diff
@@ -0,0 +1,98 @@
...
+    """
+    Sample class for Pascal VOC format datasets.
+
+    Pascal VOC format stores object detection and segmentation annotations
+    with bounding boxes in xyxy format (xmin, ymin, xmax, ymax) absolute coordinates,
+    along with label names and optional attributes like difficult, truncated, occluded, and pose.
+
+    Segmentation masks are loaded lazily from SegmentationClass/ and SegmentationObject/
+    directories when accessed.
+    """
```
leoll2, line 47: Does this sample also work for classification? In Geti, we need VOC primarily for the classification task.

src/datumaro/experimental/fields/base.py
```diff
@@ -32,10 +32,15 @@ class Field:
     Attributes:
         semantic: A string tag used for disambiguation when multiple fields
             of the same type exist (e.g., "default", "left", "right").
+        categories_from: Optional name of another field from which to share categories.
```
leoll2, line 35: I don't fully understand the purpose of `categories_from`. As far as I understand, the categories are defined at Dataset level, so what does it mean that a field shares the categories of another field?

AlbertvanHouten: Categories are attached to a field, not the entire dataset. Pascal VOC has a label field and a class mask_field and instance mask_field. Label_field defines the regular classes through LabelCategories while the mask fields hold MaskCategories that contain additional info like a colormap. The MaskCategories for instance_masks and class_masks are the same though so that is what categories_from is being used for.

src/datumaro/experimental/data_formats/voc/sample.py
AlbertvanHouten: Yes, all fields are optional so it can work with less info

src/datumaro/experimental/fields/base.py
leoll2: Ok, now I understand that `instance_mask` reuses the `MaskCategories` of `class_mask`, but what ensures that `MaskCategories` bind with `class_mask` in the first place, rather than with `instance_mask`? Is it because it's the first `mask_callable_field` to be declared in the sample? Or is it because it's the only `mask_callable_field` without a `categories_from`? What would happen if you didn't set `categories_from` in `instance_mask`?

AlbertvanHouten: The categories you pass to the dataset are defined in a dictionary where the key is the field name and value is the categories themselves. This ensures the binding to the correct field. If you didnt set categories_from and also don't manually pass categories through this dictionary to the dataset a [ValueError](https://github.com/open-edge-platform/datumaro/blob/develop/src/datumaro/experimental/schema.py#L194) will be thrown because it expects certain fields like label_field and mask_field to have categories attached to them for validation purposes. 

image

leoll2 (approved):

PR #2071, opened 2026-03-12
Add missing media/image converters 

AlbertvanHouten:
**New converters for image and media field interoperability:**

* Added `ImagePathToImageInfoConverter` to efficiently extract image dimensions from file paths by reading image headers only, avoiding full image loads.
* Added `ImagePathToImageCallableConverter` to wrap image paths as lazy-loading callables, enabling on-demand image loading.
* Added `MediaPathToImagePathConverter` to convert media paths (images or video frames) to image paths, including frame extraction from videos as needed.
* Added `ImagePathToMediaPathConverter` to promote image paths to media paths, supporting unified handling of images and videos.
* Added `MediaInfoToImageInfoConverter` and `ImageInfoToMediaInfoConverter` for converting between rich media metadata and simple image info structs.


src/datumaro/experimental/converters/image_converters.py
```diff
@@ -569,3 +569,129 @@ def convert(self, df: pl.DataFrame) -> pl.DataFrame:
...
+            DataFrame with ImageInfoField column containing width and height
+        """
+        input_col = self.input_path.name
+        output_col = self.output_info.name
+
+        widths: list[int | None] = []
+        heights: list[int | None] = []
+
+        for path in df[input_col]:
+            if path is None:
```
leoll2, line 618: In what scenario does an ImagePathField contain a null value?

src/datumaro/experimental/converters/image_converters.py
```diff
@@ -569,3 +569,129 @@ def convert(self, df: pl.DataFrame) -> pl.DataFrame:
...
+                continue
+
+            try:
+                with Image.open(str(path)) as img:
+                    w, h = img.size
+                    widths.append(w)
+                    heights.append(h)
+            except Exception:
+                widths.append(None)
+                heights.append(None)
```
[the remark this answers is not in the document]
leoll2: I agree with Copilot that setting width/height to null for images that can't be opened is an implicit behavior. I think it would be better to not catch the exception, what do you think?

src/datumaro/experimental/converters/video_converters.py
```diff
@@ -342,3 +347,318 @@ def load_media() -> np.ndarray:
...
+            DataFrame with MediaInfoField column
+        """
+        input_col = self.input_info.name
+        output_col = self.output_info.name
+
+        return df.with_columns(
+            pl.struct(
+                pl.col(input_col).struct.field("width").alias("width"),
+                pl.col(input_col).struct.field("height").alias("height"),
+                pl.lit(None, dtype=pl.String()).alias("source_path"),
```
leoll2: Is `source_path` only used for videos?

AlbertvanHouten: It is set in neither ImageInfo or VideoInfo actually. I will remove it, paths should be in Image/VideoPathFields

src/datumaro/experimental/converters/image_converters.py
AlbertvanHouten: It shouldn't. I've changed it to raise an error if it happens

leoll2 (approved):

PR #2076, opened 2026-03-18
Refactor dataset import/export functions and add direct_only option

AlbertvanHouten:
This pull request introduces a new `direct_only` option for schema conversion operations, allowing users to restrict converters to only those where all output field types are a subset of the input field types (i.e., preventing cross-field-type conversions such as BBoxField→PolygonField). 

It also merges `save_dataset`/`export_dataset` and `load_dataset`/`import_dataset` methods as their logic overlaps. `export_images` and `export_videos` arguments were also unified into a single `export_media` argument. 

### Checklist


src/datumaro/experimental/docs/import_export.md
```diff
@@ -205,37 +193,42 @@ dataset = load_yolo_dataset(root_dir="/path/to/yolo_dataset", format="ultralytic
...
+parameter for format-agnostic code:
 
 ```python
-from datumaro.experimental.data_formats.base import DataFormat, load_dataset, save_dataset
+from datumaro.experimental import import_dataset, export_dataset
+from datumaro.experimental.data_formats.base import DataFormat
 
 # Load COCO
-dataset = load_dataset(
+dataset = import_dataset(
```
leoll2, line 206: I think the names `load_dataset`/`save_dataset` are more intuitive than `import_dataset`/`export_dataset`. What do you think?

AlbertvanHouten: It might undersell it a bit. There is a lot of additional functionality when importing/exporting like conversion to different dataset types,  automatic schema detection and handling of media when exporting. The legacy logic and some other dataset libraries also seem to mainly use export/import terminology.

leoll2 (approved):

PR #2083, opened 2026-03-24
Add type converters for NumericField, BoolField, and StringField

AlbertvanHouten:
### Checklist


src/datumaro/experimental/converters/type_converters.py
```diff
@@ -0,0 +1,251 @@
+# Copyright (C) 2025 Intel Corporation
```
leoll2: Wrong year

src/datumaro/experimental/converters/type_converters.py
```diff
@@ -0,0 +1,251 @@
...
+        output_is_list = self.output_numeric.field.is_list
+
+        if input_is_list and not output_is_list:
+            log.warning(
+                "Converting list to scalar for numeric field '%s': "
+                "only the first element per sample is kept, all other elements are discarded.",
+                input_col,
+            )
+            # List(dtype) → dtype: take first element
+            df = df.with_columns(pl.col(input_col).list.first().alias(output_col))
```
leoll2: I don't think we should allow list to scalar conversions, they are likely to cause bugs

AlbertvanHouten: I can have it raise an error instead but then I will also apply this everywhere, including the LabelShapeConverter. This will block MultiLabelClassification -> MultiClassClassification conversions. What do you think?

leoll2: I'm fine with blocking the ML -> MC path, I can't imagine a scenario where it's okay to drop all labels except the first one. Labels themselves don't have a specific order, the first label isn't more special than the second or third label.

leoll2 (approved):

PR #2086, opened 2026-03-25
Reimplement label shape converter to support multi_label/is_list swaps

AlbertvanHouten:
Reimplement swaps from label_field(is_list=True, multi_label=False) to label_field(is_list=False, multi_label=True) and vice-versa

Also keep the original image names when exporting datasets. 

Fix regression from https://github.com/open-edge-platform/datumaro/pull/2083

### Checklist

### DIFF
--- a/src/datumaro/experimental/export_import.py
+++ b/src/datumaro/experimental/export_import.py
@@ -148,17 +148,32 @@ def _export_image_path_field(
     idx: int,
     output_dir: Path,
 ) -> str | None:
-    """Export an ImagePathField by copying the file from the filesystem."""
+    """Export an ImagePathField by copying the file from the filesystem.
+
+    The original filename is preserved.  When a name collision occurs
+    (e.g. two source images from different directories share the same
+    filename) a numeric suffix is appended to disambiguate.
+    """
     try:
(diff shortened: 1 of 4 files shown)


src/datumaro/experimental/export_import.py
```diff
@@ -148,17 +148,32 @@ def _export_image_path_field(
...
-        abs_path = output_dir / rel_path
+        # Preserve the original filename; fall back to index-based naming
+        # only when the source has no name at all.
+        dest_name = source_path.name if source_path.name else f"{field_name}_{idx:06d}.png"
+        abs_path = output_dir / dest_name
+
+        # Handle name collisions with a numeric suffix
+        counter = 1
+        while abs_path.exists():
+            stem = source_path.stem if source_path.stem else f"{field_name}_{idx:06d}"
```
leoll2, line 171: Can the `else` condition actually happen? Is it possible for a path to not have a stem?

leoll2 (approved):

src/datumaro/experimental/export_import.py
AlbertvanHouten: I don't think so.

PR #2087, opened 2026-03-27
Media converter improvements 

AlbertvanHouten:
Separate media converters from video converters in another file
Fix an issue with output spec for media path conversions
Add VideoInfo converters. 

### Checklist


src/datumaro/experimental/converters/video_converters.py
leoll2, line 1: Could you please highlight the parts that changed for:

> Fix an issue with output spec for video conversions

Because of the refactoring, it's a bit challenging to review the diffs

AlbertvanHouten: Its easiest to spot in the test files here (which were wrong previously) https://github.com/open-edge-platform/datumaro/pull/2087/changes#diff-11e8edff0211fb80b8104e388a0092b03ca8d81785b992717790625dd12d1267

The fix is in filter_output_spec in the media converters

Before: 
image

After:
image


The fix takes both format and channels_first from the output field now rather than the input.

leoll2 (approved):

PR #2112, opened 2026-04-21
feat(dataset): optimize batch dataset building with PyArrow integration

AlbertvanHouten:
This pull request introduces a major performance optimization for batch dataset construction in the experimental API, especially when handling deeply nested data types (such as polygons and bounding boxes). The core improvement is a new batch path that collects raw Python values per column and constructs Polars Series in bulk using PyArrow, dramatically reducing overhead compared to the previous per-sample DataFrame approach. Additionally, new helper methods and refactorings make the batch path accessible to both the main dataset and legacy converters.

Tested with a large dataset which went from 27 seconds to load down to 3.5 seconds. 

The most important changes are:

**Performance improvements for batch dataset construction:**

* Added `build_series_bulk` and `polars_dtype_to_pyarrow` helpers in the new `arrow_utils.py` module to efficiently build Polars Series from lists of Python values, leveraging PyArrow for nested types. 
* Refactored `Dataset.append_batch` to collect all raw values per column and construct Series in bulk, replacing the slow per-sample DataFrame concatenation. 
**API and field improvements:**

* Introduced the `to_python_scalars` method to the `Field` base class and implemented fast-path overrides for annotation fields, enabling efficient extraction of nested Python lists for batch operations.

### Checklist


tests/unit/experimental/test_arrow_utils.py
```diff
@@ -0,0 +1,581 @@
...
+        arr = np.array([1, 2, 3], dtype=np.int32)
+        assert numpy_to_nested_lists(arr) == [1, 2, 3]
+
+
+# ---------------------------------------------------------------------------
+# polars_dtype_to_pyarrow
+# ---------------------------------------------------------------------------
+
+
+class PolarsDtypeToPyarrowTest:
```
leoll2, line 86: If possible, test files should be named after the corresponding source file.

Method `polars_dtype_to_pyarrow` is in `experimental/arrow_utils.py`, so I expect these tests to be in `tests/unit/experimental/test_arrow_utils.py`.

leoll2 (commented):
> Tested with a large dataset which went from 27 seconds to load down to 3.5 seconds.

This optimization looks huge, wow! What I don't understand is this part:

> batch path that collects raw Python values per column and constructs Polars Series in bulk using PyArrow

How is it possible that working with Python lists is faster than pl.Series? Is it the overhead of initializing many small-sized pl.Series?

AlbertvanHouten:
> > Tested with a large dataset which went from 27 seconds to load down to 3.5 seconds.
> 
> This optimization looks huge, wow! What I don't understand is this part:
> 
> > batch path that collects raw Python values per column and constructs Polars Series in bulk using PyArrow
> 
> How is it possible that working with Python lists is faster than pl.Series? Is it the overhead of initializing many small-sized pl.Series?

It is indeed the overhead of initialising the many small size pl.Series where the bulk path pays that overhead only once.

leoll2 (approved):

PR #2131, opened 2026-05-18
Refactor ImageField for channels_first conversion and tensor format

### LINKED ISSUE #2130: Bug: TensorField.from_polars() uses wrong inverse transpose, swapping image H and W
## Description

When reading image tensors from parquet/arrow format, `TensorField.from_polars()` applies `np.transpose(data, (2, 0, 1))` to undo the CHW→HWC conversion done during export. However, the correct inverse of `transpose(2, 0, 1)` is `transpose(1, 2, 0)`, not `transpose(2, 0, 1)` again.

This results in the spatial dimensions (H and W) being exchanged in the returned tensor:
- Stored CHW: `(3, 720, 1280)`
- After wrong transpose `(2, 0, 1)`: `(1280, 3, 720)` — should be `(720, 1280, 3)` (HWC)

The `dm_image_info` metadata correctly reports `height=720, width=1280`, but the actual pixel tensor has H and W swapped after any CHW recovery heuristic.

## Impact

This bug causes **silent data corruption** downstream. Any consumer that trusts `dm_image_info` for spatial dimensions (e.g., bounding-box canvas_size) will have a mismatch with the actual pixel layout. In our case, this caused detection evaluation to score 0% because ground-truth bounding boxes referenced a coordinate space inconsistent with the decoded image.

## Steps to Reproduce

```python
from datumaro.experimental.dataset import Dataset

# Any parquet dataset with video frame images (W > H, e.g., 1280x720)
ds = Dataset("/path/to/parquet/dataset")
item = ds[0]

# dm_image_info says width=1280, height=720
print(item.dm_image_info)  # ImageInfo(width=1280, height=720)

# But the image tensor has H and W swapped
print(item.image.shape)    # torch.Size([1280, 3, 720]) — W=1280 is in H position
```

## Expected Behavior

`item.image.shape` should be `(3, 720, 1280)` (CHW) or `(720, 1280, 3)` (HWC), consistent with `dm_image_info`.

## Suggested Fix

In `TensorField.from_polars()`, change:
```python
data = np.transpose(data, (2, 0, 1))  # WRONG inverse
```
to:
```python
data = np.transpose(data, (1, 2, 0))  # correct inverse of (2, 0, 1)
```

The transpose `(2, 0, 1)` maps `(H, W, C) → (C, H, W)`. Its inverse must map `(C, H, W) → (H, W, C)`, which is `(1, 2, 0)`.

## Environment

- Datumaro version: 2.0.0 (develop branch)
- Installed from: `git+https://github.com/open-edge-platform/datumaro.git@develop`
- Python: 3.12
- OS: Linux

## Workaround

We currently work around this in the consuming library by detecting the H/W swap (comparing image shape to `bboxes.canvas_size` from `dm_image_info`) and applying `tensor.transpose(-2, -1)` to correct it.

AlbertvanHouten:
Transpose in to/from polars methods of the Tensorfield were incorrectly applied resulting in different outputs when doing a round-trip on a tensor with channels_first=True.

This PR removes the transpose from Tensorfield entirely, it is not needed as no converter needs to know about the stored channel order of a tensor. Instead, ImageField now implements its own to/from_polars methods and gets a fix so the round-trip of an image in polars with channels_first=True gives identical output. 

Resolves #2130 

### Checklist

### DIFF
--- a/src/datumaro/experimental/fields/images.py
+++ b/src/datumaro/experimental/fields/images.py
@@ -26,12 +26,10 @@ class TensorField(Field):
     Attributes:
         semantic: String tag describing the tensor's purpose
         dtype: Polars data type for tensor elements
-        channels_first: Whether the tensor uses channels-first format (C, H, W) vs channels-last (H, W, C)
     """
 
     semantic: str = "default"
     dtype: pl.DataType = field(default_factory=pl.UInt8)
-    channels_first: bool = False
 
     def to_polars_schema(self, name: str) -> dict[str, pl.DataType]:
         """Generate Polars schema with separate columns for data and shape."""
@@ -40,9 +38,6 @@ class TensorField(Field):
     def to_polars(self, name: str, value: Any) -> dict[str, pl.Series]:
         """Convert tensor to flattened data and shape information."""
         numpy_value = to_numpy(value, self.dtype)
-
-        if self.channels_first and numpy_value is not None:
-            numpy_value = np.transpose(numpy_value, (2, 0, 1))
 
         schema = self.to_polars_schema("tensor")
 
@@ -62,9 +57,6 @@ class TensorField(Field):
         shape = df[name + "_shape"][row_index]
         numpy_data = np.array(flat_data).reshape(shape) if flat_data is not None else None
 
-        if self.channels_first and numpy_data is not None:
-            numpy_data = np.transpose(numpy_data, (2, 0, 1))
-
         return from_polars_data(numpy_data, target_type)  # type: ignore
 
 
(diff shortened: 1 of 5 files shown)


src/datumaro/experimental/fields/images.py
```diff
@@ -83,18 +75,61 @@ def tensor_field(dtype: Any, semantic: str = "default") -> Any:
...
+        """Generate Polars schema with separate columns for data and shape."""
+        return {name: pl.List(self.dtype), name + "_shape": pl.List(pl.Int32())}
+
+    def to_polars(self, name: str, value: Any) -> dict[str, pl.Series]:
+        """Convert image tensor to flattened data."""
+        numpy_value = to_numpy(value, self.dtype)
+
+        if self.channels_first and numpy_value is not None:
+            # User provides CHW; store as HWC internally
+            numpy_value = np.transpose(numpy_value, (1, 2, 0))
```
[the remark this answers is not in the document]
AlbertvanHouten: Not an issue, grayscale images shouldn't use channels_first as grayscale doesn't have channels.

src/datumaro/experimental/fields/images.py
```diff
@@ -83,18 +75,64 @@ def tensor_field(dtype: Any, semantic: str = "default") -> Any:
...
+        semantic: String tag describing the image's purpose
+        dtype: Polars data type for pixel values
         format: Image color format (e.g., "RGB", "BGR", "RGBA")
+        channels_first: Whether the user provides/receives CHW format
     """
 
+    semantic: str = "default"
+    dtype: pl.DataType = field(default_factory=pl.UInt8)
     format: str = "RGB"
+    channels_first: bool = False
```
kprokofi, line 99: should we change code in getitune after that PR, right?

AlbertvanHouten: As far as I can see not, only TensorField loses channels_first now but its not used in getitune. ImageField previously already had channels_first through inheritance of Tensorfield.

kprokofi (approved):

kprokofi:
I will install this branch in local environment and check whether it resolves the problem with misaligned shapes in getitune

UPD: the branch fully resolves the dimension mismatch issue.

leoll2 (approved):

PR #2177, opened 2026-07-20
Tiling patch for instance segmentation

A-Artemis:
Moved tiling patch from Geti Tune to Datumaro
The tiling patch for instance segmentation ensures that the the amount of annotations and labels match for tiled datasets.

### Checklist

- [x] I have updated the [documentation](https://github.com/open-edge-platform/datumaro/tree/develop/docs) accordingly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Experimental module
+
+#### Bug fixes
+- Fixed `InstanceMaskTiler` not filtering instance masks during tiling. Instance
+  masks are now pruned to the instances whose bounding box intersects each tile,
+  mirroring `BboxTiler`/`LabelTiler`, so masks stay aligned with boxes and labels
+  (previously `len(masks) != len(bboxes) == len(labels)`, which corrupted
+  instance-segmentation training).
+
 ## Release 1.13.0
 
 This release is centered around the maturation of the experimental dataset
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,17 +5,6 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## Unreleased
-
-### Experimental module
-
-#### Bug fixes
-- Fixed `InstanceMaskTiler` not filtering instance masks during tiling. Instance
-  masks are now pruned to the instances whose bounding box intersects each tile,
-  mirroring `BboxTiler`/`LabelTiler`, so masks stay aligned with boxes and labels
-  (previously `len(masks) != len(bboxes) == len(labels)`, which corrupted
-  instance-segmentation training).
-
 ## Release 1.13.0
 
 This release is centered around the maturation of the experimental dataset
--- a/src/datumaro/experimental/tiling/tilers.py
+++ b/src/datumaro/experimental/tiling/tilers.py
@@ -86,18 +86,57 @@ class MaskTiler(Tiler):
         return pl.DataFrame({column_name: results_data, shape_column: results_shape})
 
 
+def _instance_intersects_tile(instance: np.ndarray, x: int, y: int, width: int, height: int) -> bool:
+    """Return whether an instance mask's bounding box intersects the tile.
+
+    The criterion mirrors :class:`BboxTiler` exactly: using the instance's
+    full-image bounding box ``(x1, y1, x2, y2)`` (derived from its non-zero
+    pixels, with ``x2``/``y2`` exclusive), the instance is kept iff::
+
+        (x2 > tile_x) & (x1 < tile_x2) & (y2 > tile_y) & (y1 < tile_y2)
+
+    An all-zero mask (no pixels) has no bounding box and is dropped.
+
+    Args:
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,17 @@ All notable changes to this project will be documented in this file.
...
+## Unreleased
+
+### Experimental module
+
+#### Bug fixes
+- Fixed `InstanceMaskTiler` not filtering instance masks during tiling. Instance
+  masks are now pruned to the instances whose bounding box intersects each tile,
+  mirroring `BboxTiler`/`LabelTiler`, so masks stay aligned with boxes and labels
+  (previously `len(masks) != len(bboxes) == len(labels)`, which corrupted
+  instance-segmentation training).
```
leoll2: Please revert changes to this file

src/datumaro/experimental/tiling/tilers.py
```diff
@@ -86,18 +86,57 @@ def tile(self, df: pl.DataFrame, tiles_df: pl.DataFrame, slice_offset: int = 0)
...
+    as :class:`BboxTiler` (the instance's full-image mask bounding box must
+    intersect the tile rectangle). Instances are kept in their original order so
+    alignment with boxes/labels is preserved.
     """
 
     field_spec: AttributeSpec[InstanceMaskField]
 
     def tile(self, df: pl.DataFrame, tiles_df: pl.DataFrame, slice_offset: int = 0) -> pl.DataFrame:
-        """Extract instance mask regions for each tile."""
+        """Extract and filter instance mask regions for each tile."""
```
leoll2, line 211: Have you tried to benchmark the performance of this function? Its computational complexity depends quadratically on the tile size, linearly on the dataset size and number of instances per image: in theory, it could easily become very expensive, in practice I don't know if it becomes a bottleneck.

A-Artemis: I believe it is the same as when it was in Geti, there was no major slow down. I check the timings in the logs when the detailed performance logs were enabled.

A-Artemis: I've optimized it.

src/datumaro/experimental/tiling/tilers.py
```diff
@@ -86,37 +86,126 @@ def tile(self, df: pl.DataFrame, tiles_df: pl.DataFrame, slice_offset: int = 0)
...
         shape_column = f"{column_name}_shape"
         results_data = []
         results_shape = []
 
+        # Cache per source image so the O(H*W) mask scan (reshape + bbox
+        # derivation) is paid at most once per image instead of once per tile.
+        # Many tiles typically share the same source image, so this avoids
+        # O(num_tiles * num_instances * H * W) work on large / dense images.
+        instances_cache: dict[int, np.ndarray] = {}
+        boxes_cache: dict[int, np.ndarray] = {}
```
leoll2: This cache grows unbounded, for large datasets it's very likely to cause OOM errors. Assuming that all the tiles of one image are processed sequentially rather than interleaved with other images' tiles (assumption to verify), then it would be better to only cache the last seen image.

kprokofi: Agree this cached should be done per image probably as stated in the comment, but as far as I can see. This tile method is applied on the whole data frame. 
Consider more smart caching 

2) We apply it for Instace Seg. Should the same be applied for bounding box and semantic seg?

leoll2 (approved):

src/datumaro/experimental/tiling/tilers.py
```diff
@@ -86,37 +131,127 @@ def tile(self, df: pl.DataFrame, tiles_df: pl.DataFrame, slice_offset: int = 0)
...
+    # Reduce over the full mask once per instance (the only O(H*W) pass).
+    rows_any = instances.any(axis=2)  # (N, H): rows containing any pixel
+    cols_any = instances.any(axis=1)  # (N, W): cols containing any pixel
+    for i in range(num_instances):
+        rows = np.nonzero(rows_any[i])[0]
+        cols = np.nonzero(cols_any[i])[0]
+        if rows.size == 0 or cols.size == 0:
+            continue
+        boxes[i] = (cols[0], rows[0], cols[-1] + 1, rows[-1] + 1)
+    return boxes
```
AlbertvanHouten, line 164: Certain mask shapes might still result in a very large bbox which then if tiled in a certain size could result in tiles without any mask part in it. If the goal is to have the same amount annotations/labels at all times then you should look for a solution that uses the mask directly.

AlbertvanHouten: For example: 

Image

A-Artemis: The BBoxTiler and LabelTiler also use the large bounding box of the mask, and the new custom tiler added is consistent with their implementation. This keeps mask, boxes, and labels amount equal across all tilers.

I agree that there are chances where the tiler will create tiles outside the mask, but if we want to keep the tiles to exclusively overlap the mask then we need to update all tiling functionality.

AlbertvanHouten: BBoxTiler using a bbox of the annotation is a pixel perfect wrap and for LabelTiler it will also always have the label so it is fine there. Wrapping a bbox around a mask adds pixels and is the only outlier that can produce incorrect results.

AlbertvanHouten: What is the purpose of this PR, just to filter out tiles with no annotation data? If so I guess a few possible empty tiles but having a quick solution is the better approahc then

A-Artemis: Yes, the purpose is to drop tiles with no annotations for instance segmentation, keeping the count of masks, tiles, and labels all equal.

There are cases that tiles are empty but intersect the bbox, but since it uses the same criteria as the BboxTiler I think it is safe to merge this.

AlbertvanHouten (approved):

PR #2185, opened 2026-08-05
Fix out-of-bounds annotation issue in Geti

itallix:
- [x]  Clip floating conversion noise for cxcywh -> x1y1x2y2

### Checklist


src/datumaro/experimental/converters/annotation_converters.py
```diff
@@ -1149,10 +1149,10 @@ def convert(self, df: pl.DataFrame) -> pl.DataFrame:
...
             intermediate_expr = pl.col(input_bbox_name).list.eval(
                 pl.concat_arr(
-                    pl.element().arr.get(0) - pl.element().arr.get(2) / 2,  # x1 = cx - w/2
-                    pl.element().arr.get(1) - pl.element().arr.get(3) / 2,  # y1 = cy - h/2
-                    pl.element().arr.get(0) + pl.element().arr.get(2) / 2,  # x2 = cx + w/2
-                    pl.element().arr.get(1) + pl.element().arr.get(3) / 2,  # y2 = cy + h/2
+                    (pl.element().arr.get(0) - pl.element().arr.get(2) / 2).clip(lower_bound=0),  # x1 = cx - w/2
+                    (pl.element().arr.get(1) - pl.element().arr.get(3) / 2).clip(lower_bound=0),  # y1 = cy - h/2
+                    (pl.element().arr.get(0) + pl.element().arr.get(2) / 2).clip(lower_bound=pl.element().arr.get(2)),  # x2 = cx + w/2
+                    (pl.element().arr.get(1) + pl.element().arr.get(3) / 2).clip(lower_bound=pl.element().arr.get(3)),  # y2 = cy + h/2
```
leoll2: I don't understand what's happening here, why are you lower-clipping `x2` based on `w` (width of the box)?

The boundaries of `x2` and `y2` should be the width and height *of the image*, respectively, NOT the width and height *of the box* (`arr.get(2)` and `arr.get(3)`).

itallix: Indeed. Added by mistake. We cannot clip these coords here properly, but small fractions shouldn't be an issue anyway, will be imported as is.

src/datumaro/experimental/converters/annotation_converters.py
```diff
@@ -968,26 +968,28 @@ def op(x: pl.Expr, y: pl.Expr) -> pl.Expr:
         return df.with_columns(
             pl.struct([pl.col(input_polygon_name), pl.col(image_shape_name)])
             .map_elements(
-                lambda row: [
+                lambda row: (
```
leoll2, line 971: Can you confirm the only change here is some parentheses `( ... )` around the list comprehension, only for the sake of formatting? Github diff shows many lines changed, but I think there's no functional change in this block. Just double-checking.

leoll2 (approved):

src/datumaro/experimental/converters/annotation_converters.py
leoll2: To clip `x2` and `y2` (upper bound) you need the image width and height, which I think is stored in another column called `f"{input_image_name}_shape"` (at least for polygons, L957).

src/datumaro/experimental/converters/annotation_converters.py
itallix: This is how ruff formats lamda bodies - updated by `uv run ruff format .`. I don't see any functional changes here.

src/datumaro/experimental/converters/annotation_converters.py
itallix: So far, the main problem is negative values as Geti enforces strict validation for that. This is possible for x1y1, but should be safe to ignore other two.

PR #2197, opened 2026-09-01
Harden field/type resolution against arbitrary input

leoll2:
This PR addresses potential vulnerabilities in the Datumaro field/type resolution logic that exploit `importlib.import_module` with untrusted, arbitrary data. Now, only modules under `datumaro`, `numpy` and `torch`/`torchvision` are allowed to go through the resolution path.

Tested in Geti (training and dataset I/E).

### Checklist


src/datumaro/experimental/schema.py
```diff
@@ -67,15 +67,32 @@ class AttributeSpec(Generic[TField]):
     "NoneType": type(None),
 }
 
+# Schemas can be loaded from untrusted dataset files (e.g. a Datumaro-format
+# dataset received from an external source). Resolving a type module/name pair
+# read from such a file must never be allowed to import and return an arbitrary
+# importable object since that value is later invoked as a callable elsewhere
+# (see Field.from_polars implementations).
+# Only modules that are part of Datumaro itself or numpy are ever imported here.
+_ALLOWED_TYPE_MODULES = frozenset({"numpy"})
```
AlbertvanHouten: I think you should also include torch.tensor here

AlexanderBarabanov (commented):
Thank you, Leonardo!
Can we add unit tests here as well (positive and negative)?

src/datumaro/experimental/schema.py
leoll2: Right, added

AlbertvanHouten (approved):

PR #2204, opened 2026-09-13
Fix invalid escape sequences in Datumaro format reader and converter tests

MBK-fr:
### Problem

Three modules use non-raw string literals with invalid escapes:
- `src/datumaro/plugins/data_formats/datumaro/base.py` - `\s`, `\d` (version-parsing regexes)
- `tests/unit/experimental/converters/test_image_converters.py` - `\.`
- `tests/unit/experimental/converters/test_mask_converters.py` - `\(`

Python 3.12+ emits `SyntaxWarning: invalid escape sequence` for these on import / test collection.

### Why it matters for Datumaro

CI runs Python 3.9-3.12; the warnings show on the 3.12 job. (The existing `\d+\.d+` pattern is left exactly as-is - only the backslashes are escaped, no regex-logic change.)

### Fix

Escape the backslashes; each string value is unchanged (ruff W605-style).

### Verification

`ast.dump` unchanged per file; `python3 -W error` py_compile passes on all three after, failed before. Commit carries the DCO `Signed-off-by` line required by the repo's contributing guide.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_014HBQNzAf5C2E3MiJC48HQw


leoll2:
Hi @MBK-fr , thanks for your contribution. Could you link an an example of CI run where you observed the warnings?

